PR #24, opened 2024-12-10
feat: add `phpunit` plugin

azjezz:


bendavies:
missing an addition to [mago-full.toml](https://github.com/carthage-software/mago/blob/main/examples/mago-full.toml)?

azjezz:
we should probably get rid of that file and write proper documentation. but i guess we can add it there for now.

PR #25, opened 2024-12-11
Add support for `protected(set)` and `public(set)`

ryangjchandler:
`protected(set)` and `public(set)` are semantically valid.  `public public(set)` is silly but valid PHP code. PHP doesn't warn or error because of the identical visibilities.

I've added a new linter rule to tell you that the additional write visibility is redundant too:

image

Side note - I've marked the fix as potentially unsafe since there could be a reason for having identical read/write visibilities, but also happy to change that to a safe fix since it doesn't really have an effect.

### DIFF
--- a/crates/linter/src/plugin/redundancy/rules/redundant_write_visibility.rs
+++ b/crates/linter/src/plugin/redundancy/rules/redundant_write_visibility.rs
@@ -1,0 +1,62 @@
+use mago_ast::Modifier;
+use mago_ast::Property;
+use mago_fixer::SafetyClassification;
+use mago_reporting::Annotation;
+use mago_reporting::Issue;
+use mago_reporting::Level;
+use mago_span::HasSpan;
+use mago_walker::Walker;
+
+use crate::context::LintContext;
+use crate::rule::Rule;
+
+#[derive(Clone, Debug)]
+pub struct RedundantWriteVisibilityRule;
+
+impl Rule for RedundantWriteVisibilityRule {
+    fn get_name(&self) -> &'static str {
+        "redundant-write-visibility"
+    }
+
+    fn get_default_level(&self) -> Option<Level> {
+        Some(Level::Help)
+    }
+}
+
+impl<'a> Walker<LintContext<'a>> for RedundantWriteVisibilityRule {
+    fn walk_in_property(&self, property: &Property, context: &mut LintContext<'a>) {
+        let modifiers = property.modifiers();
+
+        if modifiers.is_empty() {
+            return;
+        }
+
+        let Some(write_visibility) = modifiers.get_first_write_visibility() else {
+            return;
+        };
+
+        let Some(read_visibility) = modifiers.get_first_read_visibility() else {
+            return;
+        };
+
+        match (read_visibility, write_visibility) {
+            (Modifier::Public(_), Modifier::PublicSet(_))
+            | (Modifier::Protected(_), Modifier::ProtectedSet(_))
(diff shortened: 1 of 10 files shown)


azjezz:
Huh! i was not even aware that those modifiers were supported!

Thank you for your contribution! :D

crates/linter/src/plugin/redundancy/rules/redundant_write_visibility.rs
```diff
@@ -0,0 +1,55 @@
+use mago_ast::{Modifier, Property};
+use mago_reporting::{Annotation, Issue, Level};
+use mago_span::HasSpan;
+use mago_walker::Walker;
+
+use crate::{context::LintContext, rule::Rule};
+
+#[derive(Clone, Debug)]
+pub struct RedundantWriteVisibility;
```
azjezz:
```suggestion
pub struct RedundantWriteVisibilityRule;
```

crates/linter/src/plugin/redundancy/rules/redundant_write_visibility.rs
```diff
@@ -0,0 +1,55 @@
...
+            return;
+        };
+
+        match (read_visibility, write_visibility) {
+            (Modifier::Public(_), Modifier::PublicSet(_)) | (Modifier::Protected(_), Modifier::ProtectedSet(_)) | (Modifier::Private(_), Modifier::PrivateSet(_)) => {
+                let issue = Issue::new(context.level(), "identical write visibility has no effect")
+                    .with_help("remove the redundant write visibility modifier")
+                    .with_annotations(vec![
+                        Annotation::secondary(read_visibility.span()),
+                        Annotation::secondary(write_visibility.span()),
```
azjezz:
```suggestion
                        Annotation::primary(write_visibility.span())
                              .with_message("redundant write visibility."),
```

we should highlight this as primary since it is the span with the "problem".

ryangjchandler:
Commit is saying unverified, but I've definitely got a GPG key setup 🫠

crates/semantics/src/walker.rs
```diff
@@ -1038,9 +1038,9 @@ impl SemanticsWalker {
                         last_visibility = Some(modifier.span());
```
azjezz, line 1138:
```suggestion
                        last_write_visibility = Some(modifier.span());
```

this looks like a bug ( by me ) 😅

crates/linter/src/plugin/redundancy/rules/redundant_write_visibility.rs
```diff
@@ -0,0 +1,55 @@
...
+        };
+
+        let Some(read_visibility) = modifiers.get_first_read_visibility() else {
+            return;
+        };
+
+        match (read_visibility, write_visibility) {
+            (Modifier::Public(_), Modifier::PublicSet(_)) | (Modifier::Protected(_), Modifier::ProtectedSet(_)) | (Modifier::Private(_), Modifier::PrivateSet(_)) => {
+                let issue = Issue::new(context.level(), "identical write visibility has no effect")
+                    .with_help("remove the redundant write visibility modifier")
```
azjezz:
```suggestion
                    .with_help("remove the redundant write visibility modifier.")
```

crates/semantics/src/walker.rs
ryangjchandler: no this looks fine, there's no `last_write_visibility` var in that method 👍

azjezz: Ah, my bad :)

azjezz:
Thank you!

PR #26, opened 2024-12-11
feat(linter): introduce laravel plugin

azjezz:
this PR introduces a new laravel plugin, which contains a single rule.

in the future, we plan to add more rules, but this should be our starting point for now, similar to what we did with symfony and phpunit.

### DIFF
--- a/crates/linter/src/plugin/laravel/rules/safety/no_request_all.rs
+++ b/crates/linter/src/plugin/laravel/rules/safety/no_request_all.rs
@@ -1,0 +1,87 @@
+use mago_ast::*;
+use mago_ast_utils::reference::*;
+use mago_reporting::*;
+use mago_span::HasSpan;
+use mago_walker::Walker;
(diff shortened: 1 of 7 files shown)


crates/linter/src/plugin/laravel/rules/safety/no_request_all.rs
```diff
@@ -0,0 +1,87 @@
...
+                    };
+
+                    let fqcn = context.lookup_name(identifier);
+
+                    fqcn.eq_ignore_ascii_case(REQUEST_FACADE)
+                        || fqcn.eq_ignore_ascii_case(REQUEST_FQCN)
+                        || context.lookup(&identifier.value()).eq_ignore_ascii_case(REQUEST_CLASS)
+                }
+                _ => {
+                    // we do not care abotu closure creation..
```
bendavies, line 71: about

PR #30, opened 2024-12-19
chore: prefer strict comparison in help message

bendavies:


crates/linter/src/plugin/best_practices/rules/no_multi_assignments.rs
```diff
@@ -36,7 +36,7 @@ impl<'a> Walker<LintContext<'a>> for NoMultiAssignmentsRule {
                     .with_message("Consider splitting this statement into multiple assignments."),
             )
             .with_note("Multiple assignments in a single statement can be confusing and lead to unexpected behavior.")
-            .with_help(format!("Did you mean `{a} = ({b} == {c})` instead? Ensure the intended logic is clear."));
+            .with_help(format!("Did you mean `{a} = ({b} === {c})` instead? Ensure the intended logic is clear."));
```
azjezz, line 39: I think we might want to keep this as is.

The reason is this message just asks the user if they mistyped `==` as `=`, and it is less likely that someone would write `=` instead of `==` 🤔

bendavies: fair!

PR #33, opened 2024-12-24
Update directory walking to be multithreaded

giorgiopogliani:
Hello,

 I try to test most cases and it seems that the behaviour is the same or even improved. For example in the excludes now both `**/vendor/composer/*` and `vendor/composer/*` works.

Regarding the dir walking:
-  | am using new added jwalk crate which works in async and on mutiple cores.
-  I ignore before walking down directories that starts with a `.` and node_modules folders. This folder it's something new but I can't really think of a reason why we shouldn't ignore it, it's pretty common, has a huge amount of dirs and files, and it's not PHP.

I am fairly new rust so hopefully this looks good enough

### DIFF
--- a/src/source.rs
+++ b/src/source.rs
@@ -10,7 +10,7 @@ use crate::error::Error;
 /// Load the source manager by scanning and processing the sources
 /// as per the given configuration.
 ///
-/// #_Arguments
+/// # Arguments
 ///
 /// * `interner` - The interner to use for string interning.
 /// * `configuration` - The configuration to use for loading the sources.
--- a/src/source.rs
+++ b/src/source.rs
@@ -2,7 +2,6 @@ use ahash::HashSet;
 use mago_interner::ThreadedInterner;
 use mago_source::SourceManager;
 use std::path::Path;
-use tracing::debug;
 
 use crate::config::source::SourceConfiguration;
 use crate::consts::PHP_STUBS;
--- a/src/source.rs
+++ b/src/source.rs
@@ -1,13 +1,8 @@
-use std::path::Path;
-use std::path::PathBuf;
-
 use ahash::HashSet;
-use async_walkdir::Filtering;
-use async_walkdir::WalkDir;
-use futures::StreamExt;
-
 use mago_interner::ThreadedInterner;
 use mago_source::SourceManager;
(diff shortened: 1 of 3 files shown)


src/source.rs
```diff
@@ -16,7 +10,7 @@ use crate::error::Error;
 /// Load the source manager by scanning and processing the sources
 /// as per the given configuration.
 ///
-/// # Arguments
+/// #_Arguments
```
azjezz:
```suggestion
/// # Arguments
```

src/source.rs
```diff
@@ -33,85 +27,62 @@ pub async fn load(
...
+    let manager = SourceManager::new(interner.clone());
+    let extensions: HashSet<&String> = extensions.iter().collect();
+    let has_paths = !paths.is_empty();
+    let has_includes = !includes.is_empty();
+    let has_excludes = !excludes.is_empty();
+
+    let entries = jwalk::WalkDir::new(root.clone()).process_read_dir(|_, _, _, children| {
+        children.iter_mut().for_each(|dir_entry_result| {
+            if let Ok(dir_entry) = dir_entry_result {
+                if dir_entry.path().starts_with(".") || dir_entry.file_name.eq_ignore_ascii_case("node_modules") {
```
azjezz, line 40: we shouldn't ignore node_modules, while usually it does not contain PHP code, it could.

people using node alongside PHP should manually add this to their exclude directory.

src/source.rs
```diff
@@ -33,85 +27,62 @@ pub async fn load(
...
-        })
-        .collect();
+        });
+    });
 
-    let extensions: HashSet<&String> = extensions.iter().collect();
+    for entry in entries {
+        if let Err(_) = entry {
+            continue;
+        }
```
azjezz, line 50: we should return the error here instead of ignoring it, if the directory is not readable for example, it should be excluded.

src/source.rs
giorgiopogliani: What do you mean it could?

src/source.rs
giorgiopogliani: What do you suggest? I wouldn't stop the loop, I would just log the error to the user

PR #35, opened 2024-12-27
Introduce composer plugin

veewee:
This PR will make it possible to install the mago binaries from within composer.
It contains:

* A composer command: `composer mago:install-binary` that can be used to install the binary for currently required mago version for current platform.
  * The downloaded files will be placed in the vendor directory of this package inside composer/bin/{version}
  * It will also create a `.platform` file that contains a relative link to the mago executable for current platform.
* Event listeners to download the assets when the package is updated / installed through composer.

This package requires  activation of the plugin:

```json
    "config": {
        "allow-plugins": {
            "carthage-software/mago": true
        }
    }
```

Once the assets are in place, that is where the composer plugin stops.

Next up is a mago PHP binary that will be symlinked in `vendor/bin/mago`.
This script will:

* Load the current platform executable from the `.platform file`
* Executes the mago command in a separate processs
* Currently open stdin, stdout and stderr are being linked to the executable, making it possible to use it just as you would in a regular command.

This setup allows to run mago on multiple platforms (docker, virtualbox, host, ...) just as one would expect from regular PHP tools.

Screenshot 2025-01-03 at 14 35 52

Screenshot 2025-01-03 at 14 36 25


composer.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "carthage-software/mago",
+    "description": "Mago is a toolchain for PHP that aims to provide a set of tools to help developers write better code.",
+    "type": "composer-plugin",
+    "license": "MIT",
```
azjezz:
```suggestion
    "license": "MIT OR Apache-2.0",
```

composer/bin/mago
```diff
@@ -0,0 +1,137 @@
...
+            2 => array('file', 'php://stderr', 'w'),
+        ),
+        $pipes
+    );
+
+    if (!is_resource($process)) {
+        echo('Unable to start mago process');
+        exit(1);
+    }
+
```
azjezz, line 43: can we have the php script move the real executable to its place or symlinking it, which will avoid the overhead on every use?

veewee: Yes and no...

That would make it not possible to execute on both the host system and e.g. inside docker at the same time which it can do right now.

I can of course move the logic the the composer command and even only download the binaries for that specific platform, but it will crash when its being used like that on multiple architectures.

I've thought about it as well but I'm not sure how we want it to be used in that kind of situations.

azjezz: > will crash when its being used like that on multiple architectures.

why? can't we have it execute on every install/update and download the binary then? 🤔

veewee: It will then only download the binary for the architecture you are currently run composer on.

If you are composer installing on e.g. Mac it will have the Darwin binary. But if you also work with docker on the project, you might need some Linux version as well for the same project: if you want to run the mago command from both host and guest. So in that situation where you composer installed locally, and execute Mago grom the link guest container, it will run the Darwin version on linux and crash. People are pretty used to having the php executables working cross system.

Since this pr also introduces a composer command for installing the assets, it's pretty easy to switch versions that way though. So maybe going for one version might be OK as well.

composer/bin/mago
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env php
+<?php
+
+declare(strict_types=1);
```
veewee, line 4: I've kept this file for a few reasons:

* Composer adds a binary proxy wrapper. It somehow expects this to be a PHP file.
* In windows, the executable will be suffixed with .exe. I'm not sure if it runs without that suffix.
* This gives us some control about the availability of the executable. For example when composer runs with `no-scripts`

composer/InstallMagoBinaryCommand.php
```diff
@@ -0,0 +1,268 @@
...
+    {
+        return 'https://api.github.com/repos/carthage-software/mago' . $path;
+    }
+
+    /**
+     * @return array{platform: string, executable: string, storage_dir: string}
+     */
+    private function detectPlatformInfo(ProcessExecutor $process_executor, string $version): array
+    {
+        $arch_name = php_uname('m');
```
Zuruuh: `php_uname('m')` will return an uppercase string on windows so you should lowercase `$arch_name` like you did with `$os` later :+1:

composer/InstallMagoBinaryCommand.php
```diff
@@ -0,0 +1,268 @@
...
+            ),
+        };
+
+        $os = strtolower(php_uname('s'));
+        $vendor = 'unknown';
+        $os_suffix = '';
+        $executable_extension = '';
+
+        switch ($os) {
+            case 'windows':
```
Zuruuh: `php_uname('s')` will return `Windows NT` on windows so you should check for this string instead, though you should still make sure that the osname used in the filename is still `windows`

veewee: Can you verify if I understood your comment correctly. 

Thanks for reviewing!

Zuruuh: Yes this looks fine to me :+1: I'll test again on windows tonight

azjezz: @Zuruuh is this good to go?

Zuruuh: Yeah I tried it and can confirm it works now 😄

azjezz (approved):

PR #43, opened 2025-01-18
Add preliminary support for PHP 5.4

LinusU:
I'm using an absolute ancient version of PHP at work (got handed a 10+ year old project I now need to maintain 🙈) and are constantly running into problems with me using features that aren't implemented. So this linter is actually great for me!

This PR adds the ability to configure the `php_version` all the way down to `"5.4"`.

There is still some rules missing in order to cover 100% of the features introduced between 5.4 and 7.3, but I think that this is a good start! 🐎 

Please let me know if I'm doing anything wrong, which I probably am 😅 

Here is a screenshot of it in action:

Screenshot 2025-01-18 at 16 50 34

### DIFF
--- a/crates/linter/src/plugin/compatibility/rules/php55/finally_feature.rs
+++ b/crates/linter/src/plugin/compatibility/rules/php55/finally_feature.rs
@@ -1,0 +1,71 @@
+use indoc::indoc;
+
+use mago_ast::ast::*;
+use mago_php_version::PHPVersion;
+use mago_reporting::*;
+use mago_span::*;
+use mago_walker::Walker;
+
+use crate::context::LintContext;
+use crate::definition::RuleDefinition;
+use crate::definition::RuleUsageExample;
+use crate::rule::Rule;
+
+#[derive(Clone, Copy, Debug)]
+pub struct FinallyFeatureRule;
+
+impl Rule for FinallyFeatureRule {
+    fn get_definition(&self) -> RuleDefinition {
+        RuleDefinition::enabled("Finally Feature", Level::Error)
+            .with_maximum_supported_php_version(PHPVersion::PHP54)
+            .with_description(indoc! {"
+                Flags any usage of the `finally` keyword, which was introduced in PHP 5.5.
+
+                In environments running older versions of PHP, you can use a `try`/`catch` block without a `finally` block instead.
+            "})
+            .with_example(RuleUsageExample::valid(
+                "Using a `try`/`catch` block without a `finally` block",
+                indoc! {r#"
+                    <?php
+
+                    try {
+                        // Code that might throw an exception.
+                    } catch (Exception $e) {
+                        // Handle the exception.
+                    }
+                "#},
+            ))
+            .with_example(RuleUsageExample::invalid(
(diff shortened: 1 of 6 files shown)


crates/linter/src/plugin/compatibility/rules/php55/finally_feature.rs
```diff
@@ -0,0 +1,71 @@
...
+use crate::definition::RuleUsageExample;
+use crate::rule::Rule;
+
+#[derive(Clone, Copy, Debug)]
+pub struct FinallyFeatureRule;
+
+impl Rule for FinallyFeatureRule {
+    fn get_definition(&self) -> RuleDefinition {
+        RuleDefinition::enabled("Finally Feature", Level::Error)
+            .with_maximum_supported_php_version(PHPVersion::PHP54)
```
azjezz, line 20:
```suggestion
            .with_maximum_supported_php_version(PHPVersion::PHP55)
```

this has became exlusive now.

crates/linter/src/plugin/compatibility/rules/php56/variadic_functions_feature.rs
```diff
@@ -0,0 +1,73 @@
...
+            }
+
+            let issue = Issue::new(
+                context.level(),
+                "Variadic functions are only available in PHP 5.6 and later.",
+            )
+            .with_annotation(
+                Annotation::primary(param.span()).with_message("Variadic parameter used here."),
+            )
+            .with_note("Use `func_get_args()` if you need compatibility with older PHP versions.");
```
azjezz, line 68:
```suggestion
            .with_help("Use `func_get_args()` if you need compatibility with older PHP versions.");
```

notes are used to provide more information, help are used to provide a solution, or a workaround.

azjezz (commented):
I'm fine with this approach, but I believe we should define the minimum PHP versions we officially support:

- `>= 7.4`: This version is well-tested and reliable with Mago.
- `>= 5.4`: This can be allowed, but with caution.

The reasoning is that PHP 7 and PHP 8 introduced significant changes, particularly around operator precedence and associativity (left, right, or none). While Mago ( esp. the parser ) might work with your codebase, it could potentially fail or behave unexpectedly with others.

azjezz:
We could improve support for PHP 5.x to the point where no warnings would be necessary, but achieving this would require:

- Compiling a comprehensive list of syntax changes between PHP 5.x and PHP 8.
- Updating the lexer and parser to adapt based on the specified PHP version.

When I mention making the parser version-aware, I'm primarily referring to handling elements like operator precedence. The parser would still process features from future PHP versions, provided they don't conflict with syntax from earlier versions. For example, features like property hooks can be parsed safely even if the version is set to PHP 5.4.

azjezz:
@LinusU I'll be closing this PR.

Mago has changed quite a lot since this PR was opened, it can't be merged anymore. and after fighting with a lot of precedence issues in the parser recently, I think adding PHP 5.x support would make the parser much more complicated.

LinusU:
@azjezz sounds good!

Sorry for never getting back to this, at the moment I'm not doing any more PHP at work, but happy to hear that your tool has evolved! 🤩

PR #62, opened 2025-01-30
feat(reporting): add gitlab code quality format

vvvinceocam:
## 📌 What Does This PR Do?

Adds [Code Quality](https://docs.gitlab.com/ee/ci/testing/code_quality.html#code-quality-report-format) (aka GitLab) report format.

## 🔍 Context & Motivation

In my workplace, we use GitLab. To integrate mago into our CI pipeline, I’ve implemented the [Code Quality](https://docs.gitlab.com/ee/ci/testing/code_quality.html#code-quality-report-format) report format. It’s a subset of the [Code Climate](https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md) format.

## 🛠️ Summary of Changes

- **Feature:** Added `gitlab` reporting format.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [x] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

N/A

## 📝 Notes for Reviewers

I’ve named the format `gitlab` instead of `code-quality` (or something along those lines) because I believe it’s clearer for GitLab users.

For generating unique issue identifiers, I used the `blake3` hash algorithm. I chose it for its portability, performance, and excellent distribution properties. However, this choice is open to discussion.


Cargo.toml
```diff
@@ -125,6 +126,7 @@ tempfile = { workspace = true }
 tracing-subscriber = { workspace = true }
 indicatif = { workspace = true }
 colored = { workspace = true }
+blake3 = { workspace = true }
```
azjezz: we don't need blake in cli crate, this can be removed.

Cargo.toml
```diff
@@ -86,6 +86,7 @@ self_update = { version = "0.42.0", features = ["archive-tar", "archive-zip", "c
 openssl = { version = "0.10", features = ["vendored"] }
 tempfile = "3.15.0"
 colored = "3.0.0"
+blake3 = "1.5.5"
```
azjezz, line 89: do we need this depedency? can't we just use the `Hasher` from the `ahash` crate? we don't need something secure here.

azjezz (changes requested):

Cargo.toml
vvvinceocam: `ahash` doesn't provide any guarantee about portability. So, produced hash may differ between OSs, CPU's architecture. 

Practically, this would mean that a report generated on a laptop running macOS with an Apple Silicon chip would have different issue IDs compared to one generated on a CI worker running Linux on an x86_64 processor.

To be completely honest, I’ve never had to compare Code Quality reports. I usually just consult them in Merge Request reports. But it feels to me like it goes against the spirit of such a format to not be "portable".

The tools I’ve looked at that generate Code Quality reports tend to use cryptographic hashes like SHA256.

azjezz: make sense 👍 will go with that.

crates/reporting/src/internal/emitter/gitlab.rs
```diff
@@ -0,0 +1,87 @@
...
+
+                    let file_path = interner.lookup(&source.identifier.0).to_string();
+                    let line = source.line_number(annotation.span.start.offset) + 1;
+
+                    (file_path, line)
+                }
+                None => ("<unknown>".to_string(), 0),
+            };
+
+            let description = issue.message.as_str();
```
azjezz: should we include the help and notes here like we do in github format? 🤔

azjezz:
rebase needed.

crates/reporting/src/internal/emitter/gitlab.rs
vvvinceocam: Good point !

PR #84, opened 2025-02-23
Linter/strictness extra

vvvinceocam:
## 📌 What Does This PR Do?

This PR adds a bunch of rules into `strictness` plugin to avoid additional cases of loose comparison/type juggling.

The rules are:
- `no-elvis-operator`: detects uses of shorthand ternary operator `?:`, aka elvis operator;
- `no-empty-construct`: detects uses of `empty()` language construct;
- `require-strict-behavior`: detects uses of functions relying on loose comparison unless their `$strict` parameter is set to `true`.

~~As these rules can be considered "drastic", they are put in a distinct plugin `strictness-extra`, allowing them to be made optional and activated at the user's discretion.~~

~~The plugin is disabled by default.~~
 
## 🔍 Context & Motivation

Type jugging and loose comparison are notorious source of bugs . Moreover, expressions with weak or broad semantic fail to communicate actual developer's intent, making code harder to review and maintain.

With this in mind:
- Elvis operator (`?:`) must be replaced by coalesce operator (`??`) or explicit ternary operator with strict comparison.
- `empty()` language construct must be replaced by more specific function (like `is_null()`) or explicit strict comparison.
- `in_array()`, `array_search()` and `base64_decode()` functions must always be called with their `$strict` parameter set to `true` to rely on strict comparison.
- Similarly, when the `array_keys()` function is used with a filter value, its third parameter`$strict` must be set to `true`.

## 🛠️ Summary of Changes

- **Feature:** Added  `no-elvis-operator`, `no-empty-construct`, and `require-strict-behavior` rules to the `strictness` plugin.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [x] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

-

## 📝 Notes for Reviewers

Hi ! :wave: 

I hope you find the proposed rules relevant. Please let me know if you have any thoughts or suggestions on them. Naturally, the new rules can be reassigned to existing plugins or the plugin can be renamed for better organization and clarity.

I may not have used the various functions for displaying messages, warnings and hints  optimally. I'm happy to improve them, and the content of the messages as well.


crates/linter/src/plugin/strictness_extra/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+            name: "Strictness Extra",
+            description: "Provides extra rules that enforce strictness in the codebase.",
+            enabled_by_default: false,
+        }
+    }
+
+    fn get_rules(&self) -> Vec<Box<dyn Rule>> {
+        vec![Box::new(NoElvisOperator), Box::new(NoEmptyConstruct), Box::new(RequireStrictBehavior)]
+    }
+}
```
azjezz: Should be merged with `StrictnessPlugin`, rules can be disabled by default by setting `level` to `None`.

crates/linter/src/plugin/strictness_extra/rules/no_elvis_operator.rs
```diff
@@ -0,0 +1,46 @@
...
+
+                The shorthand ternary operator (`?:`), aka elvis operator, relies on loose comparison.
+            "})
+            .with_example(RuleUsageExample::invalid(
+                "Using the `?:` operator",
+                indoc! {r#"
+                    <?php
+
+                    $value = $foo ?: $default;
+                    "#},
```
azjezz:
```suggestion
                indoc! {r#"
                    <?php

                    $value = $foo ?: $default;
                "#},
```

crates/linter/src/plugin/strictness_extra/rules/no_empty_construct.rs
```diff
@@ -0,0 +1,59 @@
...
+                indoc! {r#"
+                            <?php
+
+                            // ...
+
+                            if (!empty($myArray)) {
+                                // ...
+                            }
+                        "#},
+            ))
```
azjezz:
```suggestion
            .with_example(RuleUsageExample::invalid(
                "Using the `empty()` construct",
                indoc! {r#"
                    <?php

                    // ...

                    if (!empty($myArray)) {
                        // ...
                    }
                "#},
            ))
```

crates/linter/src/plugin/strictness_extra/rules/no_empty_construct.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    fn lint_node(&self, node: Node<'_>, context: &mut LintContext<'_>) -> LintDirective {
+        let Node::EmptyConstruct(construct) = node else { return LintDirective::default() };
+
+        let issue = Issue::error("Use of the `empty()` function.")
```
azjezz:
```suggestion
        let issue = Issue::error("Use of the `empty` construct")
```

azjezz (changes requested):
Hi @vvvinceocam,

Thank you for the PR!

I like those new rules, but i don't see the need for a new plugin, the rules can be part of `strictness` plugin, and be disabled by default.

However, i don't see why we would disable them by default, we have stricter rules that are enabled by default already, and considering that all of the rules have a warning level, they won't be breaking any workflow.

crates/linter/src/plugin/strictness/rules/no_elvis_operator.rs
```diff
@@ -0,0 +1,46 @@
...
+
+        let issue = Issue::error("Use of the short-ternary operator (`?:`).")
+            .with_annotation(
+                Annotation::primary(expression.span()).with_message("Ambigous check due to `?:` loose comparison."),
+            )
+            .with_help("Use null coalesce operator (`??`) or ternary operator with explicit strict comparison.");
+        context.report(issue);
+
+        LintDirective::Prune
+    }
```
azjezz: we can use this same rule to detect `Conditional` expression where `then` is `None`. this happens when there is a space between `?` and `:`, e.g:

```php
$a = $b ?: $c;  // `Expression::Binary(Binary { operator: BinaryOperator::Elvis(_), .. })`
$a = $b ? : $c; // `Expression::Conditional(Conditional { then: None, .. })`

```

crates/linter/src/plugin/strictness/rules/no_empty_construct.rs
```diff
@@ -0,0 +1,59 @@
...
+            .with_annotation(
+                Annotation::primary(construct.span()).with_message("Ambigous check due to `empty()` loose semantic."),
+            )
+            .with_help(indoc! {"
+                Use explicit checks instead of `empty()` to clearly convey your intent.
+
+                Specifically, use strict comparison operators and explicit functions like `is_null()`
+                or `array_key_exist()`. When handling multiple cases, consider combining these predicates
+                with logical operators.
+            "});
```
azjezz: can you try spliting this between `with_note` and `with_help`? we would like to keep all messages as a single line.

azjezz (changes requested):

crates/linter/src/plugin/strictness/rules/no_elvis_operator.rs
vvvinceocam: PHP, please, PHP... I didn't see that coming. :sweat_smile:

As a result, I'm thinking of renaming the rule to `no_shorthand_ternary`. Sounds good to you?

azjezz: Yep!

crates/linter/src/plugin/strictness/rules/no_empty_construct.rs
vvvinceocam: Better this way?

azjezz: Yes!

azjezz (approved):
LGTM! Thanks @vvvinceocam

Issue #100, opened 2025-03-03
New formatter cases


innocenzi:
Hello, I just tried Mago again on Tempest's codebase, and here are a few formatting issues:

## ~empty parameter list is broken~ fixed in #103 

I'd expect them to stay the same as "before" or very close, but the "after" looks broken.

**Before**

```php
$path = $this->viewCache->getCachedViewPath(
    path: $view->path,
    compiledView: fn () => $this->cleanupCompiled($this->compiler->compile($view->path)),
);
```

**After**

```php
$path = $this->viewCache->getCachedViewPath(path: $view->path, compiledView: fn(
) => $this->cleanupCompiled($this->compiler->compile($view->path)));
```

## ~seemingly arrow functions~ FIXED IN #106

**Before**
```php
$callable = new CommandBusMiddlewareCallable(
    fn (object $command) => $this->container->get($middlewareClass)($command, $callable),
);
```

**After**
```php
$callable = new CommandBusMiddlewareCallable(fn (object $command) => $this->container->get(
    $middlewareClass,
)($command, $callable));
```

---

The one below, neither cases look good. Our code is too hard to read, the condition should be in a multiline closure. But the formatted code is even worse:

**Before**
```php
$command = array_find(
    array: $this->consoleConfig->commands,
    callback: fn (ConsoleCommand $consoleCommand) =>
        $consoleCommand->handler->getDeclaringClass()->getName() === $command[0]
            && $consoleCommand->handler->getName() === $command[1],
);
```

**After**
```php
$command = array_find(
    array: $this->consoleConfig->commands,
    callback: fn (ConsoleCommand $consoleCommand) => $consoleCommand->handler->getDeclaringClass()->getName() ===
        $command[0] &&
    $consoleCommand->handler->getName() === $command[1],
);
```

## ~Seemingly broken cases~ - FIXED IN #107 

**Before**

```php
    private function getStubFileFromConfigType(ConfigType $configType): StubFile
    {
        try {
            $stubPath = dirname(__DIR__) . '/Stubs';

            return match ($configType) {
                ConfigType::CONSOLE => StubFile::from($stubPath . '/console.config.stub.php'),
                ConfigType::CACHE => StubFile::from($stubPath . '/cache.config.stub.php'),
                ConfigType::LOG => StubFile::from($stubPath . '/log.config.stub.php'),
                ConfigType::COMMAND_BUS => StubFile::from($stubPath . '/command-bus.config.stub.php'),
[...]

PR #106, opened 2025-03-04
feat(formatter): improve arrow function formatting.

azjezz:
## 📌 What Does This PR Do?

This PR improves the formatting of arrow functions in two main ways:

- Consistent formatting with return statements: The expression on the right-hand side of an arrow function is now formatted using the same logic as return statement values. This means that in some cases, the expression might be wrapped in parentheses for better readability, especially when it spans multiple lines.

- Improved argument list breaking: When an arrow function is the only argument in a function call, and that arrow function has a simple expression or a function-call-like expression with simple arguments, the formatter will now break the argument list into multiple lines before breaking the arrow function's expression. This makes the code easier to read and understand.

## 🔍 Context & Motivation

This PR addresses issues raised in #100 where a user reported that the formatting of arrow functions could be difficult to read in certain cases. The changes in this PR aim to improve the readability and consistency of arrow function formatting.

## 🛠️ Summary of Changes

- **Feature:** Format arrow function expressions using the same logic as return statement values.
- **Feature:** Improve the breaking behavior of argument lists containing single arrow function arguments.
- **Tests:** Added new tests to cover the updated arrow function formatting behavior.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Related to #100

### DIFF
--- a/crates/formatter/tests/cases/arrow_return/after.php
+++ b/crates/formatter/tests/cases/arrow_return/after.php
@@ -1,0 +1,15 @@
+<?php
+
+class Example
+{
+    public function something(): void
+    {
+        $command = array_find(
+            array: $this->consoleConfig->commands,
+            callback: fn(ConsoleCommand $consoleCommand) => (
+                $consoleCommand->handler->getDeclaringClass()->getName() === $command[0] &&
+                $consoleCommand->handler->getName() === $command[1]
+            ),
+        );
+    }
+}
--- a/crates/formatter/tests/cases/arrow_return/after.php
+++ b/crates/formatter/tests/cases/arrow_return/after.php
@@ -11,5 +11,9 @@ class Example
                 $consoleCommand->handler->getName() === $command[1]
             ),
         );
+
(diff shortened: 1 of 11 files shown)


crates/formatter/tests/cases/arrow_return/after.php
```diff
@@ -0,0 +1,19 @@
...
+{
+    public function something(): void
+    {
+        $command = array_find(
+            array: $this->consoleConfig->commands,
+            callback: fn(ConsoleCommand $consoleCommand) => (
+                $consoleCommand->handler->getDeclaringClass()->getName() === $command[0] &&
+                $consoleCommand->handler->getName() === $command[1]
+            ),
+        );
```
azjezz, line 13: @innocenzi I think this looks much better than the before you provided :D

Issue #145, opened 2025-03-10
Add `strictness/require_strict_types` fix


innocenzi:
It would be nice for this rule to be fixable by `mago --fix --unsafe`. It detected 10s of file which need manual changes.

```
warning[strictness/require-strict-types]: Missing `declare(strict_types=1);` statement at the beginning of the file
   ┌─ src/Tempest/Mapper/src/Exceptions/ViewNotFound.php:1:1
   │  
 1 │ ╭ <?php
 2 │ │ 
 3 │ │ namespace Tempest\Mapper\Exceptions;
 4 │ │ 
   · │
12 │ │     }
13 │ │ }
   │ ╰─^
   │  
   = The `strict_types` directive enforces strict type checking, which can prevent subtle bugs.
   = Help: Add `declare(strict_types=1);` at the top of your file.
```

PR #147, opened 2025-03-10
feat(linter): add automatic fix for `strictness/require_strict_types`

innocenzi:
## 📌 What Does This PR Do?

This pull request introduces an automatic fix for the rule `strictness/require_strict_types`. The fix adds `declare(strict_types=1);` after the opening tag.

## 🔍 Context & Motivation

When multiple files lack this statement, it's cumbersome to add them manually. It would be convenient that Mago take care of that.

## 🛠️ Summary of Changes

Added a suggestion to the reported issue when `declare(strict_types=1)` is missing.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Closes https://github.com/carthage-software/mago/issues/145

## 📝 Notes for Reviewers

- This is my first time installing and setting up Rust
- I couldn't find where to test the suggestion. My understanding is that the tests are generated thanks to the `RuleDefinition` examples, but I don't know how to test the fix plan from that.

### DIFF
--- a/crates/linter/src/plugin/strictness/rules/require_strict_types.rs
+++ b/crates/linter/src/plugin/strictness/rules/require_strict_types.rs
@@ -1,5 +1,4 @@
 use indoc::indoc;
-use mago_fixer::FixPlan;
 use mago_fixer::SafetyClassification;
 use toml::Value;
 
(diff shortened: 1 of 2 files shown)


crates/linter/src/plugin/strictness/rules/require_strict_types.rs
```diff
@@ -141,7 +143,22 @@ impl Rule for RequireStrictTypesRule {
...
+                .with_help("Add `declare(strict_types=1);` at the top of your file.")
+                .with_suggestion(program.source, {
+                    let mut plan = FixPlan::new();
+                    let opening_tag = program.statements.iter().find_map(|statement| match statement {
+                        Statement::OpeningTag(tag) => Some(tag),
+                        _ => None,
+                    });
+
+                    plan.insert(
+                        opening_tag.map(|tag| tag.span().end).unwrap_or_else(|| program.span().start).offset,
```
innocenzi: I don't know if it's a good solution to fall back to the start of the program. Maybe it's just better not to allow the fix if we don't find an `OpeningTag`?

azjezz: if no opening tag is found, we should insert `<?php declare(strict_types=1); ?>\n` at the start of the file, as that is an html file most likely.

azjezz (changes requested):

crates/linter/src/plugin/strictness/rules/require_strict_types.rs
azjezz: Actually, we might want to do that even if an opening tag is there, if it is not the first statement, e.g:

```
<html> 
    <?php echo $data; ?>
</html>
```

having `declare(strict_types=1);` before the `echo` is weird.

azjezz: so i think the best solution is:

1. if first statement is opening tag: insert after it
2. if first statement is inline:
  2.a. if inline is a shebang, look for first opening tag, and insert after it.
  2.b. insert opening tag, declare, and closing tag before inline,

crates/linter/src/plugin/strictness/rules/require_strict_types.rs
```diff
@@ -141,7 +143,22 @@ impl Rule for RequireStrictTypesRule {
                 )
                 .with_annotation(Annotation::primary(program.span()))
                 .with_note("The `strict_types` directive enforces strict type checking, which can prevent subtle bugs.")
-                .with_help("Add `declare(strict_types=1);` at the top of your file."),
+                .with_help("Add `declare(strict_types=1);` at the top of your file.")
+                .with_suggestion(program.source, {
```
azjezz: instead of using `with_suggestion`, you should use `context.propose`, example: https://github.com/carthage-software/mago/blob/91146250af4f02e7ad1884b48b19a1098d90ad53/crates/linter/src/plugin/consistency/rules/no_tag_pair_terminator.rs#L55-L57

azjezz (changes requested):

innocenzi:
> so i think the best solution is:
> 
>     1. if first statement is opening tag: insert after it
> 
>     2. if first statement is inline:
>        2.a. if inline is a shebang, look for first opening tag, and insert after it.
>        2.b. insert opening tag, declare, and closing tag before inline,

I tried that, and that works, though I couldn't manage to try 2.b. because PHP files without tags in them are not linted by Mago

crates/linter/src/plugin/strictness/rules/require_strict_types.rs
```diff
@@ -134,15 +135,64 @@ impl Rule for RequireStrictTypesRule {
...
+
+                // If first statement is inline:
+                let inline = program.statements.iter().find_map(|statement| match statement {
+                    Statement::Inline(inline) => Some(inline),
+                    _ => None,
+                });
+
+                if let Some(inline) = inline {
+                    // If inline is a shebang, look for first opening tag, and insert after it.
+                    if inline.value.to_string().starts_with("#!") {
```
azjezz: we already store this info in the ast, see inline.kind

azjezz (approved):

PR #193, opened 2025-04-17
fix: support compiling mago-formatter to wasm32-unknown-unknown

dsherret:
## 📌 What Does This PR Do?

Allows `mago-formatter` to compile under wasm32-unknown-unknown.

## 🔍 Context & Motivation

To allow it to be used as a dprint plugin.

## 🛠️ Summary of Changes

1. Disable `ahash` using the `getrandom` crate and instead use comptime randomness, which I believe is fine for these crates?
1. Upgrade blake3 which causes it to not use the `getrandom` crate.
1. I think we can now update the wasm crate to not require the "js" feature of getrandom and then remove it as a dependency.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,7 +24,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.15",
  "once_cell",
  "version_check",
  "zerocopy 0.7.35",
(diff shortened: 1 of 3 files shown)


azjezz:
I think the reason we had `getrandom` pinned is because it is required to compile `mago_wasm` crate ( `crates/wasm` ), I will need to verify that `mago_wasm` still compiles correctly with this change.

azjezz (approved):

Issue #191, opened 2025-04-16
Global mago.toml file


trymeouteh:
### 🚀 Describe the Feature

To have the ability to set a global `mago.toml` file which Mago will use for formatting rules unless your 
current working directory includes a `mago.toml` file.

https://github.com/carthage-software/mago/discussions/190

### ✅ Use Case / Example

- Set your own default formatting rules for all PHP code (except for projects that have their own rules set)

### 💡 Proposed Solution

- To be able to place a `mago.toml` file in the Home directoryn which is used as a the global `mago.toml` file.

### 📌 Priority

Nice to have

PR #200, opened 2025-05-16
feat(config): use global config

JungoQuentin:
## 📌 What Does This PR Do?

Add test to cover order of precedence for configuration sources

## 🔍 Context & Motivation

In order to treat this issue : #191 

In order to address this issue (#191), I first added tests to ensure it works as intended

## 🛠️ Summary of Changes

- **Test:** assert order of precedence for configuration sources

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [x] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

related to #191

## 📝 Notes for Reviewers

### DIFF
--- a/crates/formatter/tests/cases/issue_208/after.php
+++ b/crates/formatter/tests/cases/issue_208/after.php
@@ -1,0 +1,9 @@
+<?php
+
+class Foo
+{
+    public function getCreatedAt(): null|\DateTimeImmutable
+    {
+        return $this->created_at;
+    }
+}
--- a/crates/formatter/tests/cases/issue_208/before.php
+++ b/crates/formatter/tests/cases/issue_208/before.php
@@ -1,0 +1,11 @@
+<?php
+
+class Foo
+{
+    public function getCreatedAt(): null|\DateTimeImmutable
+    {
+
+
+        return $this->created_at;
+    }
+}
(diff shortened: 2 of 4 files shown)


JungoQuentin:
If I understand the docstring above correctly, the environment variables should be overridden by the other sources. The tests attempt to assert that, but they fail.

If that is the case, I can easily fix it in this PR!

```rs
    /// This function attempts to load the configuration from the following sources, in order of precedence:
    ///
    /// 1. A TOML file specified by the `file` argument.
    /// 2. A TOML file named `mago.toml` in the current directory.
    /// 3. Environment variables with the prefix `MAGO_`.
```

azjezz:
> If that is the case, I can easily fix it in this PR!

Actually, the doc is wrong.

env vars should always override any other configuration. it is useful to have it this way to allow users to override a specific config for testing purposes, e.g `MAGO_THREADS=1 mago lint` to check performance under a single thread.

env vars can also be used in CI env to specify configurations that will always apply regardless of the content of the configuration file.

azjezz:
@JungoQuentin could you please fix the documentation to specify that env vars always override other sources, and update the tests accordingly? thank you.

azjezz:
Priority should ideally be:

1. env
2. specified configuration file
3. configuration file within the current working directory
4. configuration file in the home directory ( #191 )

JungoQuentin:
Thanks for the clarification @azjezz !

For adding the global configuration file, I propose the following logic:

- The global configuration overrides the default values.
- The workspace configuration overrides the default+global values.
- Using the `--file` option cancels out what is set in the last two configurations.
- Environment variables override the values, no matter what.

PR #237, opened 2025-07-20
fix(formatter): Preserve closing echo tags

### LINKED ISSUE #229: Mago format removes closing part of echo tag
### 🐞 Describe the Bug

Mago format incorrectly removes the closing part of an echo tag, which creates invalid PHP syntax. For example, `<?= 'hello' ?>` is re-formatted to `<?= 'hello'`

### 🔄 Steps to Reproduce

1. Run `mago format`
2. The command succeeds, but mis-formats files ending with echo tags
3. Mago format should produce valid PHP code

```toml
[source]
paths = ["src", "test"]
```

### 📜 Command Output

```shell
$ MAGO_LOG=trace vendor/bin/mago format test.php
DEBUG Configuration specifies 14 threads.
DEBUG Configuration specifies a stack size of 37748736 bytes.
 INFO Formatted 1 source files successfully.
```

```php
<?= 'hello' ?>
```

### 🖥️ Operating System

macOS

Composer (`composer require carthage-software/mago`)

### 📝 Additional Context

_No response_

edsrzf:
## 📌 What Does This PR Do?

Fixes #229, preserving closing echo tags even when `remove_trailing_close_tag` is true.

## 🔍 Context & Motivation

Fixing a bug

## 🛠️ Summary of Changes

- **Bug Fix:** Preserve closing echo tags, even 

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #229

## 📝 Notes for Reviewers

I tried following the existing pattern of adding an `in_` field to the `Formatter` struct. We don't need to keep a `was_` copy of the old value because tags can't be nested, as far as I'm aware.


edsrzf:
I've re-pushed to fix the clippy issue that showed up in CI.

azjezz:
Thank you @edsrzf !

PR #376, opened 2025-09-03
fix(docblock): normalize @param variable token; drop trailing punctua…

### LINKED ISSUE #351: bug: False positive: invalid-param-tag error when @param has trailing comma
## 🐞 Describe the Bug

Mago incorrectly reports an "invalid-param-tag" error when a `@param` tag has a trailing comma after the parameter name. The linter fails to recognize that the parameter exists in the function signature, treating `$installmentPlans,` (with comma) as a different parameter than `$installmentPlans` (without comma).

## 🔄 Steps to Reproduce

1. Create a PHP file with a function that has multiple parameters documented with `@param` tags
2. Add a trailing comma after one of the parameter names in the `@param` tag (e.g., `@param list<InstallmentPlan> $installmentPlans,`)
3. Run `mago analyse` on the file
4. Mago reports: "Parameter `$installmentPlans,` is not defined in this function"
5. Expected: Mago should either ignore the trailing comma or report a different error about incorrect formatting, but should recognize that the parameter exists

## ⚙️ Configuration (mago.toml)

```toml
# Mago configuration file
# For more information, see https://mago.carthage.software/#/getting-started/configuration
php-version = "8.3.0"

[source]
paths = ["config/", "src/", "migrations/"]
includes = ["vendor"]
excludes = ["src/BindHQ/Bundle/TestsBundle/Resources/LoadSystemDataReferences.php"]

[formatter]
print-width = 180
tab-width = 4
use-tabs = false
null-type-hint = "question"
always-break-named-arguments-list = false
preserve-breaking-argument-list = true
preserve-breaking-array-like = true
preserve-breaking-attribute-list = true
preserve-breaking-parameter-list = true

[linter]
integrations = ["symfony", "php-unit"]

[analyzer]
mixed-issues = false
ignore = ["non-existent-method"]
```

## 📜 Command Output

```shell
error[invalid-param-tag]: The @param tag references an unknown parameter.
   ┌─ tests/unit/BindHQ/InstallmentPlan/Form/Traits/InstallmentPlanVersionTraitTest.php:87:15
   │
87 │      * @param list<InstallmentPlan> $installmentPlans,
   │               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Parameter `$installmentPlans,` is not defined in this function
   │
   = Each `@param` tag in a docblock must correspond to a parameter in the function's signature.
   = Help: Please check for typos or add the parameter to the function signature.
```

## 📂 PHP Code Sample

```php
<?php

class Example
{
    /**
     * @param string $foo,
     * @param int $bar,
     */
    public function doSomething(
        string $foo,
        int $bar
    ): void {
        // Function implementation
    }
}
```

## 🖥️ Operating System

macOS

## 📦 How did you install Mago?

Composer (`composer require carthage-software/mago`)

z4nder:
## 📌 What Does This PR Do?

Normalize PHPDoc variable token parsing and description anchoring across docblock tags. This fixes cases where trailing punctuation (e.g., `,` or `)`) was considered part of the variable name (e.g., `"$foo,"`), and ensures by-ref (`&$x`) and variadic (`...$x`) tokens.

## 🔍 Context & Motivation

While parsing tags like `@param string $foo, desc`, the variable token could include trailing punctuation (`$foo,`), leading to incorrect identifiers and follow-on diagnostics. This PR introduces a single, zero-allocation helper that extracts only the normalized identifier token (`$foo` / `...$ids`) and moves any trailing punctuation into the description.

## 🛠️ Summary of Changes

- **Bug Fix:** Normalize variable token parsing in docblocks:
  - Exclude trailing punctuation (`,`, `)` …) from the variable name.
  - Always anchor the description immediately after the normalized name (works with and without whitespace, e.g., `", desc"` and `",desc"`).
- **Refactor:** Centralize variable token parsing via `parse_var_ident` and reuse it in:
  - `@param` (`parse_param_tag`)
  - `@param-out` (`parse_param_out_tag`)
  - `@var` (`parse_var_tag`)
  - `@assert` / `@assert-if-true` / `@assert-if-false` (`parse_assertion_tag`)
- **Tests:** Add focused tests for ref/variadic tokens and trailing punctuation with and without whitespace.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [x] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): **Docblock parser (`crates/docblock`) and analyzer diagnostics anchoring**

## 🔗 Related Issues or PRs

[Fixes #351](https://github.com/carthage-software/mago/issues/351)

## 📝 Notes for Reviewers

- Identifier rules are intentionally **ASCII-only**: `[_A-Za-z][_A-Za-z0-9]*`, matching PHP variable naming. 
- Added tests cover: `&$refVar`, `...$ids)`, `$foo,`, and no-whitespace variants (e.g., `",desc"`, `")more"`).

### DIFF
--- a/crates/docblock/src/tag.rs
+++ b/crates/docblock/src/tag.rs
@@ -110,6 +110,70 @@ pub struct VarTag {
     pub variable_name: Option<String>,
 }
 
+/// Parses a PHPDoc var and returns the identifier token.
+///
+/// Accepts `$name`, `...$name` and `&$name`
+/// ASCII character after the name (e.g., `,`, `)`, whitespace), so trailing
+///
+/// ## Examples
+///
+/// - `"$foo"`        → `Some("$foo")`
+/// - `"&$foo"`       → `Some("$foo")`
+/// - `"...$ids)"`    → `Some("...$ids")`
+/// - `"$bar,"`       → `Some("$bar")`
+/// - `"$"`           → `None`
+/// - `"...$"`        → `None`
+/// - `"$1x"`         → `None`
+///
+/// # Arguments
+///
+/// * `raw` - Raw token starting at the variable position.
+///
+/// # Returns
+///
+/// `Some(&str)` with a subslice contains the `$`/`...$` prefix
+/// and the identifier, or `None` if the token does not start with `$`/`...$`
+///
+/// # Notes
+///
+/// - Identifier rules are ASCII-only: `[_A-Za-z][_A-Za-z0-9]*`.
+/// - The returned slice length can be used to anchor any trailing description.
+#[inline]
+fn parse_var_ident(raw: &str) -> Option<&str> {
+    // tolerate "&$x" in docblocks
+    let raw = raw.strip_prefix('&').unwrap_or(raw);
+
+    // accept "$name" or "...$name"
+    let (prefix_len, rest) = if let Some(r) = raw.strip_prefix("...$") {
+        (4usize, r)
+    } else if let Some(r) = raw.strip_prefix('$') {
+        (1usize, r)
+    } else {
+        return None;
+    };
+
+    // PHP identifier rules (ASCII + underscore): [_A-Za-z][_A-Za-z0-9]*
+    let bytes = rest.as_bytes();
+    if bytes.is_empty() {
+        return None;
+    }
+
+    let is_start = |b: u8| b == b'_' || b.is_ascii_alphabetic();
+    let is_cont = |b: u8| is_start(b) || b.is_ascii_digit();
+
+    if !is_start(bytes[0]) {
+        return None;
+    }
+
+    let mut len = 1usize;
+    while len < bytes.len() && is_cont(bytes[len]) {
+        len += 1;
+    }
+
+    // Some from ...$ OR $ until ascii
+    Some(&raw[..prefix_len + len])
+}
+
 /// Parses the content string of a `@template` or `@template-covariant` tag.
 ///
 /// Extracts the template name, an optional modifier (`as`, `of`, `super`),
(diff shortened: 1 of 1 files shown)


azjezz:
@z4nder after merging this, i discovered that it would break `codex`.

we left `...` token before the variable on purpose.

see: https://github.com/carthage-software/mago/blob/f42cf303041dbe4d5aabffe2fc891d6454a526db/crates/codex/src/scanner/function_like.rs#L364

we need a way to tell if `...` was used in the `@param` in order to turn `@param array<string> $foo` into `@param string $foo` ( because we need the single type ).

see: https://github.com/carthage-software/mago/blob/f42cf303041dbe4d5aabffe2fc891d6454a526db/crates/codex/src/scanner/function_like.rs#L413-L425

azjezz:
we can actually do better than keeping `...`, introduce a `Variable` structure that has flags like `is_by_reference` and `is_variadic` maybe?

azjezz:
we are going to need to re-write most of that docblock in the future either way to avoid allocations, as currently it allocates a lot of things on the heap that can be just a reference.

z4nder:
> @z4nder after merging this, i discovered that it would break `codex`.
> 
> we left `...` token before the variable on purpose.


I was checking and the name parse is not removing the `...`. I couldn't quite understand the case in which it broke. Would there be any input in docblock for me to validate?

Checking this test case, the name still keeps returning with `...`
```rust
#[test]
    fn test_param_variadic_trailing_paren_without_space() {
        let content = " list<int> ...$items)more";
        let span = test_span_for(content);
        let result = parse_param_tag(content, span).unwrap();
        assert_eq!(result.name, "...$items");
        assert_eq!(result.description, ")more");
    }
```

z4nder:
> we can actually do better than keeping `...`, introduce a `Variable` structure that has flags like `is_by_reference` and `is_variadic` maybe?

Sounds like a good idea, I'll test how this could be done. Are you considering changing the String type to the new Variable or adding it as a new field?

Similar to this
```rust
pub struct Variable<'a> {
    pub token: &'a str,        
    pub is_variadic: bool,
    pub is_by_reference: bool,
}


pub struct ParameterTag {
    pub span: Span,
    pub name: Variable,
    pub type_string: TypeString,
    pub description: String,
}

```

azjezz:
Yes, that's the idea, maybe renaming `name` to `variable` as well, and `Variable.token` to `Variable.name`

azjezz:
>I was checking and the name parse is not removing the .... I couldn't quite understand the case in which it broke. Would there be any input in docblock for me to validate?

Ops, ignore me then.

z4nder:
fn raw_name(&self) -> std::borrow::Cow<'_, str>

> Yes, that's the idea, maybe renaming `name` to `variable` as well, and `Variable.token` to `Variable.name`

https://github.com/carthage-software/mago/pull/402

Issue #233, opened 2025-07-10
Global Config: Settings always fallback to global config


trymeouteh:
### 🐞 Describe the Bug

If you have a `mago.toml` file in the root of a project directory and a `mago.toml` file in your Home directory, and if the project `mago.toml` file edits a settings that is not edited in the global `mago.toml` file, Mago will format the code using the settings from the global `mago.toml` file when it should just use the default value for that setting.

### 🔄 Steps to Reproduce

1. Create global `mago.toml` file and use the settings in the `mago.toml` file provided.
2. Create folder for a PHP project
3. Create project `mago.toml` file inside the project folder and use the settings in the `mago.toml` file provided.
4. Create `index.php` file in the project file using this for code
```
<?php

function foo()
{
	return 'Hello, world!';
}
```
5. Format `index.php` using Mago

### ⚙️ Configuration (mago.toml)

Global `mago.toml` file

```
[format]
function_brace_style = "same_line"
```

Project `mago.toml` file

```
[format]
use_tabs = true
```

With these settings, all settings from the global `mago.toml` file should be ignored since there is a `mago.toml` file in the project folder. However it will still use settings from the global `mago.toml` file if they are not present in the project `mago.toml` file.

### 📜 Command Output

```shell
~/Desktop/mago-test-project$ mago fmt
INFO All source files are already formatted.
```

### 📂 PHP Code Sample (If Applicable)

The results

```
<?php

function foo() {
	return 'Hello, world!';
}
```

What the results should be

```
<?php

function foo()
{
	return 'Hello, world!';
}
```

### 🖥️ Operating System

Linux

### 📦 How did you install Mago?

Homebrew (brew install mago)

### 📝 Additional Context

Having the global config as a fallback for settings is not a good idea since a project on Github that uses Mago will want to have their `mago.toml` in the project directory be the configuration file that determines all the formatting setting for that project regardless what another uses global `mago.toml` settings are.

PR #385, opened 2025-09-05
feat: support yaml config file

### LINKED ISSUE #350: Yaml and json config file
### 🚀 Describe the Feature

I think it would be nice if mago supported loading config from `mago.yaml` (and `.json` which could even have a schema to provide autocomplete there 👍 )

### ✅ Use Case / Example

I think more people are familiar with yaml or json compared to toml, at least in the php ecosystem :)

### 💡 Proposed Solution

Im not familiar with rust, but it seems like the config crate has support to infer the type from the file extension.

### 📌 Priority

Nice to have

### 📝 Additional Context

_No response_

NeoIsRecursive:
## 📌 What Does This PR Do?

This PR allows mago to read yaml configs by using `config::File::with_name` instead of `config::File::from`, which makes the config crate resolve the format based on the file extension.

This is sort of like a proof of concept, and since I don't know any rust this is probably subpar.

If this is something you're interested in having in mago, I'd happily work on this some more ☺️

## 🔍 Context & Motivation

#350 

In my opinion it is way more readable:

```yaml
php-version: "8.4.0"

source:
  paths:
    - "src"
    - "tests"
    - "workbench/app/"
    - "workbench/database/factories/"
    - "workbench/database/seeders/"
  includes:
    - "vendor"
  excludes: []

formatter:
  print-width: 120
  tab-width: 4
  use-tabs: false
  null-type-hint: "question"

linter:
  integrations:
    - "laravel"
  rules:
    ambiguous-function-call:
      enabled: false
    literal-named-argument:
      enabled: false
    halstead:
      effort-threshold: 7000
```

## 🛠️ Summary of Changes

- **Feature:** Added support for reading config from `mago.yaml`.

## 📂 Affected Areas

Everything?

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

I think the description sums it up :)

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1,4 +1,5 @@
 use std::env::home_dir;
+use std::fmt::Debug;
 use std::path::Path;
 use std::path::PathBuf;
 
@@ -6,7 +7,6 @@ use config::Case;
 use config::Config;
 use config::Environment;
 use config::File;
-use config::FileFormat;
 use config::Value;
 use config::ValueKind;
 use serde::Deserialize;
@@ -103,7 +103,7 @@ impl Configuration {
         allow_unsupported_php_version: bool,
     ) -> Result<Configuration, Error> {
         let workspace_dir = workspace.clone().unwrap_or_else(|| CURRENT_DIR.to_path_buf());
-        let workspace_config_path = workspace_dir.join(CONFIGURATION_FILE);
+        let workspace_config_path = workspace_dir.join(CONFIGURATION_FILE_NAME);
 
         let mut configuration = Configuration::from_workspace(workspace_dir);
         let mut builder = Config::builder().add_source(Config::try_from(&configuration)?);
@@ -111,7 +111,7 @@ impl Configuration {
         if let Some(file) = file {
             tracing::debug!("Sourcing configuration from {}.", file.display());
 
-            builder = builder.add_source(File::from(file).required(true).format(FileFormat::Toml));
+            builder = builder.add_source(File::from(file).required(true));
         } else {
             let global_config_roots = [std::env::var_os("XDG_CONFIG_HOME").map(PathBuf::from), home_dir()];
             for global_config_root in global_config_roots {
@@ -119,16 +119,19 @@ impl Configuration {
                     continue;
                 };
 
-                let global_config_path = global_config_root.join(CONFIGURATION_FILE);
+                let global_config_path = global_config_root.join(CONFIGURATION_FILE_NAME);
 
                 tracing::debug!("Sourcing global configuration from {}.", global_config_path.display());
 
-                builder = builder.add_source(File::from(global_config_path).required(false).format(FileFormat::Toml));
+                match global_config_path.to_str() {
+                    Some(s) => builder = builder.add_source(File::with_name(s).required(false)),
+                    None => {}
+                }
             }
 
             tracing::debug!("Sourcing workspace configuration from {}.", workspace_config_path.display());
 
-            builder = builder.add_source(File::from(workspace_config_path).required(false).format(FileFormat::Toml));
+            builder = builder.add_source(File::with_name(workspace_config_path.to_str().unwrap()).required(false));
         }
 
         configuration = builder
@@ -240,6 +243,7 @@ impl Configuration {
 
 #[cfg(all(test, not(target_os = "windows")))]
 mod tests {
+    use core::str;
     use std::fs;
 
     use pretty_assertions::assert_eq;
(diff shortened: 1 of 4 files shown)


src/config/mod.rs
```diff
@@ -103,32 +103,34 @@ impl Configuration {
...
-                builder = builder.add_source(File::from(global_config_path).required(false).format(FileFormat::Toml));
+                if let Some(s) = global_config_path.to_str() {
+                    builder = builder.add_source(File::with_name(s).required(false))
+                }
             }
 
             tracing::debug!("Sourcing workspace configuration from {}.", workspace_config_path.display());
 
-            builder = builder.add_source(File::from(workspace_config_path).required(false).format(FileFormat::Toml));
+            builder = builder.add_source(File::with_name(workspace_config_path.to_str().unwrap()).required(false));
```
Bleksak: i'd remove this unwrap, the function returns a Result so you can probably just use the question mark operator

Bleksak (changes requested):
looks good to me, just remove the unwrap

src/config/mod.rs
Bleksak: actually, you can just leave it as before and use File::from(...)

NeoIsRecursive: Thanks a lot for taking the time to look at this, really appreciate it.

Does `from` also "resolve" the format by the extension? I cant find it in the [docs](https://docs.rs/config/latest/config/index.html)?

Bleksak: I'm pretty sure it does, as they both behave the same, the only difference is how they take the file source.

This is "with_name":
```rs
impl File<FileSourceFile, FileFormat> {
    /// Given the basename of a file, will attempt to locate a file by setting its
    /// extension to a registered format.
    pub fn with_name(base_name: &str) -> Self {
        Self {
            format: None,
            required: true,
            source: FileSourceFile::new(base_name.into()),
        }
    }
}
```

and this is From Path:
```rs
impl<'a> From<&'a Path> for File<FileSourceFile, FileFormat> {
    fn from(path: &'a Path) -> Self {
        Self {
            format: None,
            required: true,
            source: FileSourceFile::new(path.to_path_buf()),
        }
    }
}
```

NeoIsRecursive: Oh, okay! Then it is probably nicer to keep it like before but without the format method everywhere, right? 🧐

Bleksak: Sure :+1:

azjezz:
We might as well support json.

It would be nice if we can also update the init command to generate yaml and json if --json or --yaml are passed, or we can ask the user what they like, and default to toml.

My only question is how does this behave if both mago.toml and mago.yml are defined?

I'm currently AFK so I can't really review this properly.

NeoIsRecursive:
> We might as well support json.

Yes! enabled json as well.

> My only question is how does this behave if both mago.toml and mago.yml are defined?

I believe it will be "tried" in this order: `toml` -> `yaml` -> `json` (https://docs.rs/config/latest/src/config/file/format/mod.rs.html#55 and https://docs.rs/config/latest/src/config/file/source/file.rs.html#40).

~~I can try to add a test for it?~~

EDIT:

```rust
#[test]
fn test_toml_has_presedence_when_multiple_configs_present() {
  let workspace_path = temp_dir().join(process::id().to_string());
  std::fs::create_dir_all(&workspace_path).unwrap();

  create_tmp_file("threads = 3", &workspace_path, "toml");
  create_tmp_file("threads: 2\nphp-version: \"7.4.0\"", &workspace_path, "yaml");
  create_tmp_file("{\"threads\": 1,\"php-version\":\"8.1.0\"}", &workspace_path, "json");

  let config = Configuration::load(Some(workspace_path), None, None, None, false).unwrap();

  assert_eq!(config.threads, 3);
  assert_eq!(config.php_version.to_string(), DEFAULT_PHP_VERSION.to_string())
}
```

It seems to be sort of random which file is loaded :/

src/config/mod.rs
```diff
@@ -111,32 +111,32 @@ impl Configuration {
...
                 tracing::debug!("Sourcing global configuration from {}.", global_config_path.display());
 
-                builder = builder.add_source(File::from(global_config_path).required(false).format(FileFormat::Toml));
+                builder = builder.add_source(File::from(global_config_path).required(false));
             }
 
             tracing::debug!("Sourcing workspace configuration from {}.", workspace_config_path.display());
 
-            builder = builder.add_source(File::from(workspace_config_path).required(false).format(FileFormat::Toml));
+            builder = builder.add_source(File::from(workspace_config_path).required(false));
```
azjezz, line 139: Why not simply:

```suggestion
            builder = builder
            .add_source(File::from(path).required(false).format(FileFormat::Toml)) 
            .add_source(File::from(path).required(false).format(FileFormat::Yaml)) 
            .add_source(File::from(path).required(false).format(FileFormat::Json)) 
```

Makes it clear what formats we support + we control precedence using the order.

NeoIsRecursive: Wouldn't that merge all the configs?

Thinking a bit about this one #233.

I think one would assume that there is only "one source" and that environment variables can override?

PR #386, opened 2025-09-05
feat(codex): add support for `#[\Deprecated]` on traits

DanielEScherzer:
## 📌 What Does This PR Do?

Add support for `#[\Deprecated]` on traits, see https://wiki.php.net/rfc/deprecated_traits

## 🔍 Context & Motivation

PHP 8.5 adds support for using the `#[\Deprecated]` attribute on traits. When the attribute is present, PHP will emit deprecation warnings when the trait gets used. Since mago emits deprecation warnings when a trait documented with `@deprecated` is used, it seems sensible to also emit warnings when the deprecation is indicated via attribute.

## 🛠️ Summary of Changes

- **Feature:** Expanded existing `analysis:deprecated-trait` warnings

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): Scanner

## 🔗 Related Issues or PRs

Related to php/php-src#19045

### DIFF
--- a/crates/analyzer/tests/cases/use_inheritance_error_path.php
+++ b/crates/analyzer/tests/cases/use_inheritance_error_path.php
@@ -1,5 +1,16 @@
 <?php
 
+namespace {
+    class Attribute
+    {
+    }
+
+    #[Attribute]
+    class Deprecated
+    {
+    }
+}
+
 namespace Fixture {
     class NotATrait
     {
@@ -11,6 +22,11 @@ namespace Fixture {
 
     /** @deprecated */
     trait DeprecatedTrait
+    {
+    }
+
+    #[\Deprecated]
+    trait DeprecatedTraitFromAttribute
     {
     }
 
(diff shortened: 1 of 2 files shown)


DanielEScherzer:
This is my first time writing Rust code, I tested this locally but please let me know if it doesn't work

azjezz:
@DanielEScherzer Thank you! 

For future reference: we should probably update `semantics` crate to trigger an error if this attribute is used on a trait.

This error is helpful to catch php version incompatibility, and `semantics` allows the error to be caught really early.

DanielEScherzer:
> @DanielEScherzer Thank you!
> 
> For future reference: we should probably update `semantics` crate to trigger an error if this attribute is used on a trait.
> 
> This error is helpful to catch php version incompatibility, and `semantics` allows the error to be caught really early.

Something like a check in https://github.com/carthage-software/mago/blob/746f3cd4556e0c16bf514dd7915b9c5fdfa038ac/crates/semantics/src/internal/checker/attribute.rs#L9 ?

I can certainly try and take a look at working on that, but since this was my first time writing Rust code I figured a small self-contained patch would be easier

Edit: I realized you might be asking that `semantics` be updated as part of this PR - can that be done separately?

azjezz:
@DanielEScherzer can be left to a future PR, but yes, it should be done there, we should probably pass a parameter to that function telling it where the attribute is being used, and error for deprecated on traits, also error for other "native" attributes in wrong places

PR #397, opened 2025-09-06
feat(stubs): add php-decimal extension support

### LINKED ISSUE #395: php-decimal support
### 🚀 Describe the Feature

Coverage for php-decimal extension.

### ✅ Use Case / Example

Currently, mago doesn’t know anything about Decimal namespace (extension), so analyzer trips up and misdiagnoses legitimate function calls.

### 💡 Proposed Solution

Mago has support for DS that significantly improves its capabilities, Decimal support would work in the same way.

### 📌 Priority

Nice to have

### 📝 Additional Context

_No response_

Bleksak:
## 📌 What Does This PR Do?

This PR adds a php-decimal-ext stubs. This fixes some false positives in the analyzer.

fixes #395 

## 🛠️ Summary of Changes

- **Feature:** Added stubs for php-decimal-ext for better analysis.

## 📂 Affected Areas

- [x] Other (please specify): Stubs

## 🔗 Related Issues or PRs

fixes #395


crates/prelude/assets/extensions/decimal.php
```diff
@@ -0,0 +1,400 @@
+<?php
+
+namespace Decimal;
+
+final class Decimal implements \JsonSerializable
+{
+    /* away from zero */
```
azjezz: Please remove docs

crates/prelude/assets/extensions/decimal.php
```diff
@@ -0,0 +1,400 @@
...
+
+    public const DEFAULT_PRECISION = 28;
+    public const MIN_PRECISION = 1;
+    public const MAX_PRECISION = 0;
+
+    /**
+     * @throws \BadMethodCallException if already constructed.
+     * @throws \TypeError if the value is not a decimal, string, or integer.
+     * @throws \DomainException is the type is supported but the value could not be converted to decimal.
+     */
```
azjezz:
```suggestion
    /**
     * @throws \BadMethodCallException
     * @throws \TypeError
     * @throws \DomainException
     */
```
 same here

Bleksak: Will get to it later, why don't we allow doc? Wouldn't it be beneficial for the lsp hover later?

azjezz: we will figure that out later. i prefer if we can link php.net docs instead.

removing docs from the rest of extensions removed ~6MBs from the binary size, so we almost cut it in half.

azjezz (approved):

PR #420, opened 2025-09-10
feat: dont merge configs and support yaml and json

NeoIsRecursive:
This PR addresses #233 and #350, this PR is an alternative to #385 and if the configs are meant to be merged, I'll close this and adjust in #385 🙂

The config file will be resolved like this:

file -> workspace -> global

toml -> yaml -> json

And it will then be merged with the env variables, as before.


azjezz:
Configuration files are not meant to be merged. however, sources are meant to be merged, meaning one should be able to override a specific config using an env variable, e.g `MAGO_THREADS=x` should override `threads = y` in `config.toml`, while the rest of the file should be loaded.

NeoIsRecursive:
@azjezz what do you think about this?
If going this way mago could probably use a less "powerful" config crate, right (if there is any performance to gain there, of course)?

vasilvestre:
I haven't found documentation about yaml/json config, is it supported now ?

azjezz:
@vasilvestre yes, supported, but none of the examples use it in the documentation 😅

vasilvestre:
> @vasilvestre yes, supported, but none of the examples use it in the documentation 😅 

Should I open a new issue to discuss this ?
Symfony and Sylius has a system to show different configuration in "tabs", maybe something similar can be done.
Or we could, at least, show the different format and inform users on the main "Configuration" documentation's page

PR #427, opened 2025-09-11
feat(docblock): parse method blocks

yigitcukuren:
## 📌 What Does This PR Do?

Adds support for parsing and recognizing `@method` docblock tags to prevent "non-existent method" errors for magic methods documented in class docblocks.

## 🔍 Context & Motivation

PHP classes often use magic methods (`__call`, `__callStatic`) to dynamically handle method calls. These methods are typically documented using `@method` docblock tags to provide type information and IDE support. However, static analysis tools would report these methods as "non-existent" since they don't exist as actual PHP methods.

This change allows mago to recognize `@method` annotations and treat documented methods as valid, eliminating false positive errors during static analysis.

## 🛠️ Summary of Changes

- **Feature:** Added `@method` docblock tag parsing support
- **Feature:** Added `MethodTag` struct to represent parsed method information
- **Feature:** Added `parse_method_tag` function with support for static methods
- **Enhancement:** Extended `ClassLikeDocblockComment` to store method and static method tags
- **Enhancement:** Modified class scanner to create pseudo methods from `@method` tags
- **Enhancement:** Updated method existence checks to include pseudo methods
- **Enhancement:** Enhanced method resolution to recognize pseudo methods

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [x] Analyzer, Docblock Parser, Codex

## 📝 Notes for Reviewers

- The implementation creates `FunctionLikeMetadata` entries for pseudo methods to maintain consistency with the existing method resolution system
- Both `@method` and `@psalm-method` tags are supported
- Static methods are handled separately using `@method static` syntax
- Pseudo methods are stored in `pseudo_methods` and `static_pseudo_methods` sets in `ClassLikeMetadata`
- The change is backward compatible and doesn't affect existing functionality

### Example Usage:
```php
/**
 * @method string getName() Get the user name
 * @method static User create(array $data) Create a new user
 * @method void setAge(int $age) Set user age
 */
class User {
    public function __call($method, $args) {
        // Magic method implementation
    }
}

$user = new User();
$user->getName(); // No longer reports "non-existent method"
$user->setAge(25); // No longer reports "non-existent method"
User::create(['name' => 'John']); // No longer reports "non-existent method"
```

### DIFF
--- a/crates/analyzer/src/resolver/method.rs
+++ b/crates/analyzer/src/resolver/method.rs
@@ -306,6 +306,16 @@ pub fn get_method_ids_from_object<'ctx, 'ast, 'arena, 'object>(
         }
 
         ids.push((class_metadata, method_id, outer_object, *name));
+    } else {
+        // Check if it's a pseudo method from @method docblock
+        let lowercase_method_name = ascii_lowercase_atom(method_id.get_method_name());
+        if class_metadata.pseudo_methods.contains(&lowercase_method_name)
+            || class_metadata.static_pseudo_methods.contains(&lowercase_method_name)
+        {
+            // It's a pseudo method, treat it as existing but don't add to resolved methods
+            // since we don't have full metadata for analysis
+            ids.push((class_metadata, method_id, outer_object, *name));
+        }
     }
 
     if let Some(intersection_types) = object_type.get_intersection_types() {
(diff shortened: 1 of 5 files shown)


yigitcukuren:
Closing this since #429 covers it.

Issue #405, opened 2025-09-08
Roadmap: Tracking Issue for `1.0.0-rc.1`


azjezz:
## Overview

This issue serves as the primary tracking ticket for the features and improvements required before we can tag the `1.0.0-rc.1` release. The focus for this milestone is to achieve feature-completeness on the static analyzer's core capabilities, introduce the foundational Language Server Protocol (LSP) support, and enhance usability.

Discussion and contributions on these items are welcome.

---

### Static Analyzer Enhancements

- [x] **`@method` Tag Support**: Implement support for the `@method` PHPDoc tag to define "virtual" methods on a class.
- [x] **`@property` Tag Support**: Implement support for the `@property`, `@property-read`, and `@property-write` tags to define virtual properties.
- [x] **Magic Method Support**: Add analysis support for the `__get`, `__set`, `__call`, and `__callStatic` magic methods to correctly trace dynamic property access and method calls.
- [x] **Type Alias Support**: Implement support for custom type aliases using the `@type` docblock annotation, similar to Psalm and PHPStan. #302 
- [x] **Object Shape Support**: Implement support for object shape type `object{foo: string}` #285 
- [x] **Inheritdoc**: Implement support for `@inheritDoc` #265  
- [x] **Indexed Access Type Support**: Add support for resolving types like `T[K]`, where `T` is a known array shape and `K` is a literal key. This allows for precise type inference when accessing specific array keys. #404
- [ ] **Unused Definition Detection**: Implement checks to report unused private methods, properties, constants, and potentially other unused symbols.

### Tooling & Integrations

- [x] ~**Initial LSP Implementation**: Ship the first version of the Language Server Protocol implementation. This must include:~
    - [ ] ~Diagnostics for linter issues.~
    - [ ] ~Diagnostics for analyzer issues.~
    - [ ] ~Formatting support (`textDocument/formatting`).~
- [x] **SARIF Reporting Format**: Add `sarif` as a supported reporting format for better integration with CI/CD and security scanning tools. #340 

### Usability & Configuration

- [x] **JSON & YAML Support**: Add support for using `mago.json` or `mago.yaml` for both the main configuration and for baseline files, in addition to the existing TOML format. #350 #385 

### Architectural Changes

- [x] **Relocate Type Linter Rules**: Move all linter rules that perform types checks (e.g., `parameter-type` rules) into the analyzer. #374

aszenz:
[...]

PR #429, opened 2025-09-12
feat(analyzer): add support for @method annotations in doc comments

Bleksak:
## 📌 What Does This PR Do?

this PR adds support for analysis of `@method` annotations inside class doc comments

## 🛠️ Summary of Changes

- **Feature:** Add support for analysis of `@method` annotations inside class documentation comments.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation

## 🔗 Related Issues or PRs

related to #405

### DIFF
--- a/crates/analyzer/src/code.rs
+++ b/crates/analyzer/src/code.rs
@@ -148,6 +148,7 @@ pub enum IssueCode {
     NoValidCatchTypeFound,
     NoValue,
     NonClassUsedAsAttribute,
+    NonDocumentedMethod,
     NonExistentAttributeClass,
     NonExistentCatchType,
     NonExistentClass,
@@ -396,6 +397,7 @@ impl IssueCode {
             Self::NoValidCatchTypeFound => "no-valid-catch-type-found",
             Self::NoValue => "no-value",
             Self::NonClassUsedAsAttribute => "non-class-used-as-attribute",
+            Self::NonDocumentedMethod => "non-documented-method",
             Self::NonExistentAttributeClass => "non-existent-attribute-class",
             Self::NonExistentCatchType => "non-existent-catch-type",
             Self::NonExistentClass => "non-existent-class",
@@ -1007,7 +1009,7 @@ impl IssueCode {
         ]
     }
 
-    pub const fn get_method_issue_codes() -> [Self; 11] {
+    pub const fn get_method_issue_codes() -> [Self; 12] {
         [
             Self::AmbiguousObjectMethodAccess,
             Self::DeprecatedMethod,
@@ -1016,6 +1018,7 @@ impl IssueCode {
             Self::InvalidStaticMethodCall,
             Self::MethodAccessOnNull,
             Self::MixedMethodAccess,
+            Self::NonDocumentedMethod,
             Self::NonExistentMethod,
             Self::PossibleMethodAccessOnNull,
             Self::UnimplementedAbstractMethod,
@@ -1023,7 +1026,7 @@ impl IssueCode {
         ]
     }
 
-    pub const fn get_method_issue_code_values() -> [&'static str; 11] {
+    pub const fn get_method_issue_code_values() -> [&'static str; 12] {
         [
             "ambiguous-object-method-access",
             "deprecated-method",
@@ -1032,6 +1035,7 @@ impl IssueCode {
             "invalid-static-method-call",
             "method-access-on-null",
             "mixed-method-access",
+            "non-documented-method",
             "non-existent-method",
             "possible-method-access-on-null",
             "unimplemented-abstract-method",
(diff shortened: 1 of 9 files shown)


Bleksak:
@yigitcukuren Hi, can you please try this? Does it fully cover your use case?

I tried your PR but your parser was incomplete, so I continued with my work instead.

crates/analyzer/tests/cases/method_docblock.php
```diff
@@ -0,0 +1,29 @@
...
+$b = $t->b(10);
+/** @mago-expect analysis:type-confirmation */
+Mago\confirm($b, 'int');
+$c = Test::c(10);
+/** @mago-expect analysis:type-confirmation */
+Mago\confirm($c, 'int');
+$z = Test::d(fn(int $p) => $p * 2);
+/** @mago-expect analysis:type-confirmation */
+Mago\confirm($z, 'int');
+
```
azjezz: Can you add test cases for incorrect argument types, incorrect argument count, non existing methods.. etc? 

Also, if a method does not exist in class, and `__call` is defined, we probably should not emit any errors at all, maybe a warning is enough.

crates/codex/src/scanner/class_like.rs
```diff
@@ -817,6 +821,71 @@ fn scan_class_like<'ctx, 'arena>(
...
+                    function_parameter_metadata.flags.set(MetadataFlags::BY_REFERENCE, true);
+                }
+
+                if argument.default.is_some() {
+                    function_parameter_metadata.flags.set(MetadataFlags::HAS_DEFAULT, true);
+
+                    // TODO(azjezz): not sure how to get the type from a literal value
+                    // function_parameter_metadata.default_type = argument.default.map(
+
+                    // );
```
azjezz: we probably shouldn't bother. default values are useful for generics, but it is unlikely that those methods use generics. I say we can ignore default values for now.

crates/codex/src/scanner/docblock.rs
```diff
@@ -217,6 +219,11 @@ impl ClassLikeDocblockComment {
                     require_implements
                         .push(TypeString { value: tag.description.to_string(), span: tag.description_span });
                 }
+                TagKind::Method | TagKind::PsalmMethod => {
```
azjezz, line 224: does phpstan support `@phpstan-method` ? if yes, we should probably look for it too.

crates/docblock/src/tag.rs
```diff
@@ -10,6 +10,28 @@ pub struct Variable {
...
+    pub is_static: bool,
+    pub name: String,
+    pub argument_list: Vec<Argument>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize, PartialOrd, Ord)]
+pub struct Argument {
+    pub type_hint: Option<TypeString>,
+    pub variable: Variable,
+    pub default: Option<String>,
```
azjezz: lets replace this with `has_default` for now.
```suggestion
    pub has_default: bool,
```

crates/docblock/src/tag.rs
```diff
@@ -698,6 +734,198 @@ pub fn split_tag_content(content: &str, input_span: Span) -> Option<(TypeString,
...
+pub fn parse_method_tag(content: &str, span: Span) -> Option<MethodTag> {
+    let mut is_static = false;
+
+    let mut visibility = None;
+
+    let mut tokens = content.split_whitespace().peekable();
+    let mut acc_len = 0;
+
+    while let Some(token) = tokens.peek().cloned() {
+        // TODO(azjezz): the span calculation here is wrong because there can be more than 1 whitespace
```
azjezz: we should probably not use `split_whitespaces` because of this. and and keep track of the number of whitespaces ourselves. otherwise error reporting will be incorrect and we might even have panics in other parts of mago when trying to index using an invalid index.

crates/codex/src/scanner/docblock.rs
Bleksak: i don't think so, couldn't find it anywhere in the documentation

yigitcukuren:
> @yigitcukuren Hi, can you please try this? Does it fully cover your use case?
> 
> I tried your PR but your parser was incomplete, so I continued with my work instead.

No problem @Bleksak, I tried it and these changes cover my case without any issues. 🙏

crates/analyzer/src/code.rs
```diff
@@ -148,6 +148,7 @@ pub enum IssueCode {
     NoValidCatchTypeFound,
     NoValue,
     NonClassUsedAsAttribute,
+    NonDocumentedMethod,
     NonExistentAttributeClass,
```
azjezz, line 152: This file should not be edited manually.

Please rebase, add the code to [the script](https://github.com/carthage-software/mago/blob/fb090217a8c4f4b7c7360f24e6822a0d4983dfef/scripts/regen-analyzer-issue-codes.php#L56) and run `just regen-analyzer-issue-codes`.

azjezz (changes requested):

crates/analyzer/src/code.rs
Bleksak: fixed!

azjezz:
Thank you @Bleksak 💛

PR #439, opened 2025-09-14
feat(linter): implement no-redundant-use rule to remove unused imports

### LINKED ISSUE #154: Remove unused import
This is a useful rule that I found myself missing very often. Would be great for Mago to support it.

rachidify:
## 📌 What Does This PR Do?

Adds a new linter rule `no-redundant-use` that detects and removes unused `use` statements. The rule identifies both entirely unused import statements and specific unused items within group imports

## 🔍 Context & Motivation

To Automatically detect and remove the unused/redundant imports

## 🛠️ Summary of Changes

- **Feature:** Added `redundancy/no-redundant-use` rule.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #154

### DIFF
--- a/crates/linter/src/rule/mod.rs
+++ b/crates/linter/src/rule/mod.rs
@@ -182,6 +182,7 @@ define_rules! {
     NoRedundantFile(no_redundant_file @ NoRedundantFileRule),
     NoRedundantContinue(no_redundant_continue @ NoRedundantContinueRule),
     NoRedundantBlock(no_redundant_block @ NoRedundantBlockRule),
+    NoRedundantUse(no_redundant_use @ NoRedundantUseRule),
     NoPhpTagTerminator(no_php_tag_terminator @ NoPhpTagTerminatorRule),
     NoNoop(no_noop @ NoNoopRule),
     NoMultiAssignments(no_multi_assignments @ NoMultiAssignmentsRule),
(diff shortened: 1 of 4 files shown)


aialloc:
@azjezz Sorry for the delay and thanks for letting me take a shot at this. could you please take a look?

azjezz (changes requested):
we must also check the the name ( not for functions and constants ) is no used in any docblock.

this currently is considered unused:

```
<?php

namespace X;

use Foo;

/** @param Foo $foo */
function x($foo): string {
  return $foo->getSomething();
}
```

But it is used by mago codex.

azjezz:
I think a better approach is to target `Program` instead of `Use` node, and lint all of the program at once.

also, if we have a `use Foo;` in top-level ( outside a namespace ), or inside an unnamed namespace ( e.g `namespace { ... statements ... }` ), we should report it as redundant ( same applies to `use function` and `use const` ).

azjezz (approved):

azjezz:
Thank you @rachidify 💛

azjezz:
Note: multiple false-positives were reported by this rule. It has been completely re-written in https://github.com/carthage-software/mago/commit/8f42bc2a74ca83a807e8655b2787906b8327b48c

PR #446, opened 2025-09-15
feat(formatter) support for spaces in call arguments

joehoyle:
## 📌 What Does This PR Do?

Adds an option to add spaces around the arguments list in function calls etc.

## 🔍 Context & Motivation

Adds an option in the search of supporting the WordPress PHP Coding Standard

## 🛠️ Summary of Changes

- **Feature:** Added `formatter/space-within-call-arguments ` setting.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

If this is the right approach, there are several other rules to add around spacing (e.g. arguments in control structure arguments such as `if`, `switch` etc.

### DIFF
--- a/crates/formatter/src/internal/format/call_arguments.rs
+++ b/crates/formatter/src/internal/format/call_arguments.rs
@@ -84,6 +84,12 @@ pub(super) fn print_argument_list<'arena>(
 
     let mut contents = vec![in f.arena; clone_in_arena(f.arena, &left_parenthesis)];
 
+    if  f.settings.space_within_call_arguments {
+        contents.push(Document::Space(Space {
+            soft: true,
+        }));
+    }
+
     // First, run all the decision functions with unformatted arguments
     let should_break_all = force_break || should_break_all_arguments(f, argument_list, for_attribute);
     let should_inline = !force_break && should_inline_breaking_arguments(f, argument_list);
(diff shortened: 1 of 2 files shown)


azjezz:
Hi @joehoyle, thanks so much for taking the time to put this PR together and contribute to Mago! We really appreciate the effort.

However, we can't merge this change. I want to be transparent about the reasoning, as it's a core part of Mago's philosophy.

Mago is designed to be a highly opinionated formatter, similar in spirit to tools like Prettier or rustfmt. We actually used to have this exact option, alongside about 50 other granular spacing settings. However, as we prepared for our beta release, we made a conscious decision to remove almost all of them to simplify the tool and promote a single, consistent style across the PHP community.

You can see the large-scale removal and the reasoning behind it in these links:

- https://github.com/carthage-software/mago/commit/4b0382872b981eaf9dc47fe521a27cb8a425852b
- https://github.com/carthage-software/mago/pull/313

Our philosophy is that having fewer options makes the tool easier to maintain and, more importantly, ends debates over minor stylistic preferences.

While we can't accept this particular change due to that philosophical direction, we truly appreciate you taking the time to contribute.

joehoyle:
Ah understood @azjezz !

PR #468, opened 2025-09-22
feat(analyzer): add support for shaped objects

### LINKED ISSUE #285: `error[invalid-var-tag]` for Object properties - eg. `@var object{id:int}`
### 🐞 Describe the Bug

I got below error when I ran `mago lint` or `mago analyze` for laminas/laminas-hydrator

```
error[invalid-var-tag]: Could not resolve the type for the @var tag.
   ┌─ test/HydratorObjectPropertyTest.php:28:28
   │
28 │             /** @var object{id:int} */
   │                            ^ Unexpected token `LeftBrace`
   │
   = The parser encountered a token that was not expected at this position.
   = Help: Review the type syntax near the unexpected token.
```

It seems that current mago cannot support with Object properties var tag.
https://psalm.dev/docs/annotating_code/type_syntax/object_types/#named-objects

### 🔄 Steps to Reproduce

1. Run on `mago lint` or `mago analyze` on laminas/laminas-hydrator
2. `error[invalid-var-tag]: Could not resolve the type for the @var tag.` 
3. Expected - no errors.

```toml
# Mago configuration file
# For more information, see https://mago.carthage.software/#/getting-started/configuration
php_version = "8.1.0"

[source]
paths = ["benchmark/", "src/", "test/"]
includes = ["vendor"]
excludes = []

[format]
print_width = 120
tab_width = 4
use_tabs = false

[linter]
default_plugins = true
plugins = ["php-unit"]
```

### 📜 Command Output

```shell
$  MAGO_LOG=trace mago lint test/HydratorObjectPropertyTest.php
DEBUG Configuration specifies 8 threads.
DEBUG Configuration specifies a stack size of 37748736 bytes.
 WARN Linting aborted - no rules configured.
error[invalid-var-tag]: Could not resolve the type for the @var tag.
   ┌─ test/HydratorObjectPropertyTest.php:28:28
   │
28 │             /** @var object{id:int} */
   │                            ^ Unexpected token `LeftBrace`
   │
   = The parser encountered a token that was not expected at this position.
   = Help: Review the type syntax near the unexpected token.

```

```php
<?php

declare(strict_types=1);

namespace LaminasTest\Hydrator;

use Laminas\Hydrator\ObjectPropertyHydrator;
use PHPUnit\Framework\Attributes\CoversClass;
use PHPUnit\Framework\TestCase;

#[CoversClass(ObjectPropertyHydrator::class)]
class HydratorObjectPropertyTest extends TestCase
{
    private ObjectPropertyHydrator $hydrator;

    protected function setUp(): void
    {
        $this->hydrator = new ObjectPropertyHydrator();
    }

    public function testMultipleInvocationsWithDifferentFiltersFindsAllProperties(): void
    {
        $instance = new class {
            /** @var int */
            public $id;
            /** @var int[] */
            public $array;
            /** @var object{id:int} */
            public $object;

            public function __construct()
            {
                $this->id     = 4;
                $this->array  = [4, 3, 5, 6];
                $this->object = new class {
                    /** @var int */
                    public $id = 4;
                };
            }
        };

    }
}
```

### 🖥️ Operating System

Linux

Cargo (`cargo install mago`)

### 📝 Additional Context

```
$ mago --version
mago 1.0.0-alpha.7
```

Bleksak:
## 📌 What Does This PR Do?

this PR adds support for shaped objects documented via doc blocks

## 🛠️ Summary of Changes

- **Feature:** Added support for shaped objects.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation

## 🔗 Related Issues or PRs

fixes #285 
related to #405 

## 📝 Notes for Reviewers

not sure if this is all that was required, I took array_like and just removed <K,V> paths and kept only {...},


crates/analyzer/tests/cases/object_shape.php
```diff
@@ -0,0 +1,32 @@
...
+    public int $number;
+}
+
+/**
+ * @param object{number: int, str: string} $obj
+ * @mago-expect analysis:type-confirmation
+ */
+function basic_object_test(object $obj): void
+{
+    Mago\confirm($obj, 'object{number: int, str: string}');
```
azjezz: `confirm` should not be used in tests. It will always report `analysis:type-confirmation` regardless if the type matches or not. we cant see the output.

instead, create functions that consume the object, and ones that return it. and confirm that returning/passing of the object results in no error.

crates/codex/src/ttype/atomic/object/shaped.rs
```diff
@@ -0,0 +1,129 @@
...
+use crate::ttype::atomic::array::key::ArrayKey;
+use crate::ttype::union::TUnion;
+
+/// Metadata for a PHP object (map/dictionary-like).
+///
+/// Corresponds to `object{'key': TVal, 1: TVal2 ...}` shape.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Eq, Hash, PartialOrd, Ord, Default)]
+pub struct TShapedObject {
+    /// Specific types known for certain keys (`ArrayKey`). The bool indicates if the element is optional.
+    pub known_items: Option<BTreeMap<ArrayKey, (bool, TUnion)>>,
```
azjezz:
```suggestion
    pub known_properties: Option<BTreeMap<ArrayKey, (bool, TUnion)>>,
```

crates/codex/src/ttype/atomic/object/shaped.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub non_empty: bool,
```
azjezz:
```suggestion
```

This does not apply to objects.

crates/codex/src/ttype/comparator/union_comparator.rs
```diff
@@ -79,6 +82,34 @@ pub fn is_contained_by(
...
+            for (key, item) in known_items {
+                let Some(property) = class_like.properties.get(&concat_atom!("$", &key.to_atom())) else {
+                    continue;
+                };
+
+                let Some(type_metadata) = &property.type_metadata else {
+                    continue;
+                };
+
+                if type_metadata.type_union != item.1 {
```
azjezz, line 107: We should compare the types fully, not using `!=`. if a function expects `object{username: string}` and we give it `User` which has property `username` with type `non-empty-string`, it should be allowed in.

crates/codex/src/ttype/comparator/union_comparator.rs
```diff
@@ -79,6 +82,34 @@ pub fn is_contained_by(
...
+                    continue;
+                }
+
+                matched_count += 1;
+            }
+
+            if matched_count == known_items.len() {
+                type_match_found = true;
+            }
```
azjezz: This does not take into account optional properties.

crates/codex/src/ttype/builder.rs
```diff
@@ -604,6 +605,51 @@ fn get_shape_from_ast(
...
+            let mut next_offset = 0;
+
+            for field in &shape.fields {
+                let field_is_optional = field.is_optional();
+
+                let key = match field.key.as_ref() {
+                    Some(field_key) => {
+                        let key = match field_key.key {
+                            ShapeKey::String { value, .. } => ArrayKey::String(atom(value)),
+                            ShapeKey::Integer { value, .. } => ArrayKey::Integer(value),
```
azjezz: we should error here. objects can't have integer property names.

We should probably have a new ast for `object{}`, because the name of the property can't contain `-` for example, which is currently allowed because we allow it for arrays.

crates/codex/src/ttype/atomic/object/shaped.rs
azjezz:
```suggestion
    pub known_items: Option<BTreeMap<Atom, (bool, TUnion)>>,
```

Property names are all strings. not array keys.

crates/codex/src/ttype/atomic/object/shaped.rs
```diff
@@ -0,0 +1,129 @@
...
+use crate::ttype::TType;
+use crate::ttype::TypeRef;
+use crate::ttype::atomic::array::key::ArrayKey;
+use crate::ttype::union::TUnion;
+
+/// Metadata for a PHP object (map/dictionary-like).
+///
+/// Corresponds to `object{'key': TVal, 1: TVal2 ...}` shape.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Eq, Hash, PartialOrd, Ord, Default)]
+pub struct TShapedObject {
```
azjezz:
```suggestion
pub struct TObjectWithProperties {
```

maybe?

crates/type-syntax/src/parser/internal/object.rs
```diff
@@ -0,0 +1,230 @@
...
+        additional_fields: {
+            if !stream.is_at(TypeTokenKind::Ellipsis)? {
+                None
+            } else {
+                Some(ShapeAdditionalFields {
+                    ellipsis: stream.consume()?.span,
+                    parameters: parse_generic_parameters_or_none(stream)?,
+                })
+            }
+        },
```
azjezz: this does not apply to objects.

crates/type-syntax/src/parser/internal/object.rs
```diff
@@ -0,0 +1,230 @@
...
+                        }
+                    }
+
+                    found_key
+                };
+
+                let field = ShapeField {
+                    key: if has_key {
+                        Some(ShapeFieldKey {
+                            key: parse_shape_field_key(stream)?,
```
azjezz, line 86: object keys can only be identifiers probably. or tokens that are keywords ( but don't contain `-` )

azjezz:
Reconciler support is missing, for example, the following should not emit any errors:

```php
/**
 * @psalm-assert-if-true object{name: string, username: string} $_object
 */
function is_user(object $_object): bool {
  return false;
}

/** @param object{name: string, username: string} $object */
function get_user_name(object $object): string {
  return $object->name;
}

function foo(object $object): void {
  if (is_user($object)) {
    echo get_user_name($object);
  }
}
```

However, i think it is okay to not ship reconcilation support now, and we can add it in a follow up PR.

crates/analyzer/src/expression/assignment/static_property_assignment.rs
```diff
@@ -57,7 +60,7 @@ pub(crate) fn analyze<'ctx, 'arena>(
                 Issue::error("Invalid property assignment value").with_annotation(
                     Annotation::primary(property_access.class.span()).with_message(format!(
                         "{}::{} with declared type {}, cannot be assigned type {}",
-                        resolved_property.declaring_class_id,
+                        resolved_property.declaring_class_id.unwrap_or(atom("")),
```
azjezz, line 63:
```suggestion
resolved_property.declaring_class_id.unwrap_or(atom("<object>")),
```

having an empty string in the error message for class name is weird, let's use `<object>`.

crates/analyzer/src/expression/unary.rs
```diff
@@ -1251,6 +1253,23 @@ fn cast_type_to_object<'ctx, 'ast, 'arena>(
             TAtomic::Object(_) => {
                 possibilities.push(t.clone());
             }
+            TAtomic::Array(TArray::Keyed(keyed_array)) => {
+                let mut known_properties = BTreeMap::new();
+                if let Some(known_items) = &keyed_array.known_items {
+                    for (key, item) in known_items {
+                        let property_name = match key {
+                            ArrayKey::Integer(value) => atom(&value.to_string()),
```
azjezz, line 1261:
```suggestion
                            ArrayKey::Integer(value) => i64_atom(value),
```

we don't want to allocate on the heap

crates/analyzer/src/resolver/property.rs
```diff
@@ -149,25 +150,64 @@ pub fn resolve_instance_properties<'ctx, 'ast, 'arena>(
...
+
+                for prop_name in &property_names {
+                    let key = atom(prop_name.trim_start_matches('$'));
+                    let Some((_, value)) = known_properties.get_key_value(&key) else {
+                        if !block_context.inside_isset {
+                            report_ambiguous_access(context, property_selector, object_expression.span());
+                        }
+
+                        continue;
+                    };
```
azjezz, line 182: If `optional` is true, we should trigger a warning at least.

PR #497, opened 2025-09-28
fix: make session argument optional

mathe42:
## 📌 What Does This PR Do?

Small fix in the session_id() stub.
See https://www.php.net/manual/de/function.session-id.php for refrence

## 🔍 Context & Motivation

Unexpected errors get thrown by mago anlyze.
```
error[too-few-arguments]: Too few arguments provided for function `session_id`.
    ┌─ php_source\lib\index.php:151:25
    │
151 │ $session_id = session_id();
    │               ----------^^ More arguments expected here
    │               │
    │               For this function call
    │
    = Expected at least 1 argument(s) for non-optional parameters, but received 0.
    = Help: Provide all required arguments.
```

## 🛠️ Summary of Changes
- **Bug Fix:** Fixed Stub
- 
## 📂 Affected Areas

- [x] Analyze
- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

### DIFF
--- a/crates/prelude/assets/extensions/standard.php
+++ b/crates/prelude/assets/extensions/standard.php
@@ -2123,7 +2123,7 @@ function escapeshellarg(string $arg): string
  *
  * @pure
  */
-function passthru(string $command, &$result_code): null|false
+function passthru(string $command, &$result_code = null): null|false
 {
 }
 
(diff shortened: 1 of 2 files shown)


mathe42:
After creating this PR found more issues of same kind in the stub. Fixed them...

crates/prelude/assets/extensions/standard.php
```diff
@@ -2123,7 +2123,7 @@ function escapeshellarg(string $arg): string
  *
  * @pure
```
azjezz: should be removed. this function is not pure.

PR #509, opened 2025-09-30
feat(prelude): add gd extension support

yigitcukuren:
## 📌 What Does This PR Do?

Adds stubs for the gd extension.

## 🛠️ Summary of Changes

Feature: Introduced gd.php stub file with functions, constants, and classes.

## 📂 Affected Areas

- [x] Other (please specify): Stubs


crates/prelude/assets/extensions/gd.php
```diff
@@ -0,0 +1,697 @@
...
+
+function imagesavealpha(\GdImage $image, bool $enable): bool
+{
+}
+
+function imagelayereffect(\GdImage $image, int $effect): bool
+{
+}
+
+function imagecolorallocatealpha(\GdImage $image, int $red, int $green, int $blue, int $alpha): int|false
```
azjezz:
```suggestion
function imagecolorallocatealpha(GdImage $image, int $red, int $green, int $blue, int $alpha): int|false
```

same applies for all other functions.

crates/prelude/assets/extensions/gd.php
```diff
@@ -0,0 +1,697 @@
...
+
+function imagecreatefromgd2(string $filename): \GdImage|false
+{
+}
+
+function imagecreatefromgd2part(string $filename, int $x, int $y, int $width, int $height): \GdImage|false
+{
+}
+
+function imagecreatefrombmp(string $filename): \GdImage|false
```
azjezz: should we add `@psalm-ignore-falsable-return`? It is supported by Mago.

This makes it so Mago does not complain about `false` when using the result as a `GdImage`

azjezz (changes requested):

yigitcukuren:
> redundant `@param` and `@return` that don't add value should be removed, as they only increase binary size and make compilation slower with no added value.

removed 👍🏽

PR #525, opened 2025-10-06
feat(analyzer): add docblock parameter inheritance

LeTamanoir:
## 📌 What Does This PR Do?

Implements docblock parameter type inheritance from parent classes and interfaces, allowing child implementations to automatically inherit `@param` types from their parents when not explicitly overridden.

## 🔍 Context & Motivation

When implementing interfaces or extending classes in PHP, developers often omit @param docblock tags in child methods since the parameter types are already documented in the parent. However, the analyzer was not inheriting these types, leading to:

- False positives for mixed-type warnings
- Inability to analyze callback signatures properly (e.g., callable(string): bool)

This change enables the analyzer to automatically inherit docblock parameter types from parent methods, improving type inference without requiring redundant documentation.

## 🛠️ Summary of Changes

- Feature: Added docblock parameter type inheritance in the populator phase
     - New function inherit_docblock_parameter_types() in crates/codex/src/populator.rs
     - Walks through class hierarchies using existing overridden_method_ids data
     - Inherits @param types from parent interfaces/classes when child methods lack explicit docblock types

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

- https://github.com/carthage-software/mago/issues/526

## 📝 Notes for Reviewers

It's my first Rust PR, consider me a total noob

### DIFF
--- a/crates/analyzer/src/statement/function_like/mod.rs
+++ b/crates/analyzer/src/statement/function_like/mod.rs
@@ -186,6 +186,64 @@ pub fn analyze_function_like<'ctx, 'ast, 'arena>(
     Ok(artifacts)
 }
 
+/// Retrieves @param type from parent method docblocks for a specific parameter.
+///
+/// This function supports docblock inheritance: when a child class method doesn't have a type
+/// for a parameter, it will inherit the type from parent class/interface methods.
+///
+/// # Arguments
+///
+/// * `context`: The main analysis context, providing access to the codebase metadata
+/// * `block_context`: The current block context containing scope information
+/// * `parameter_name`: The name of the parameter (e.g., "$callback")
+///
+/// # Returns
+///
+/// An `Option<&TypeMetadata>` containing the inherited type from parent method, if found
+fn get_inherited_parameter_type<'ctx>(
+    context: &Context<'ctx, '_>,
+    block_context: &BlockContext<'ctx>,
+    parameter_name: &str,
+) -> Option<&'ctx TypeMetadata> {
+    // Check if we're in a method context
+    let function_like = block_context.scope.get_function_like()?;
+    let class_like = block_context.scope.get_class_like()?;
+    let method_name = function_like.name?;
+
+    // Get parent methods that this method overrides
+    let parent_method_map = class_like.overridden_method_ids.get(&method_name)?;
+
+    // Prioritize: direct parent class first, then interfaces
+    let mut parent_methods: Vec<_> = parent_method_map.iter().collect();
+    parent_methods.sort_by_key(|(parent_class, _)| {
+        // Direct parent class comes first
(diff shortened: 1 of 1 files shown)


crates/analyzer/src/expression/assignment/mod.rs
```diff
@@ -1451,4 +1451,68 @@ mod tests {
             IssueCode::ExpressionIsTooComplex,
         ]
     }
+
+    test_analysis! {
```
azjezz: Please create a test case in `crates/analyzer/tests/cases` and add it to `crates/analyzer/tests/mod.rs` instead.

This old way of testing is not recommanded anymore, as it does not support builtin PHP symbols, we try from time to time to move `test_analysis! { }` tests to `cases/` but its going slowly 😅

LeTamanoir: Will move the test 👌 , is it okay if I name it with the github issue ID ? I see a lot already done that way

LeTamanoir: also was wondering, where do you think is the best place to add this kind of code, I don't see any docblock stuff done in the populator so I'm unsure whether this should be added there 🤔

azjezz: Yes, using github issue id is fine.

And best place to test this is analyzer, even though its not part of it :D

LeTamanoir: Ok, will update the code and test :+1:

LeTamanoir: ok I added the following branching in `add_parameter_types_to_context` (`crates/analyzer/src/statement/function_like/mod.rs`):
```rust
        let declared_parameter_type = if let Some(inherited_type_metadata) =
            get_inherited_parameter_type(context, block_context, &parameter_variable_str.to_string())
        {
            expand_type_metadata(context, block_context, artifacts, function_like_metadata, inherited_type_metadata)
        } else if let Some(type_metadata) = parameter_metadata.get_type_metadata() {
            expand_type_metadata(context, block_context, artifacts, function_like_metadata, type_metadata)
        } else {
            get_mixed()
        };
```

I feel like I'm doing smth wrong, wouldn't it make it simpler if the docblock meta coming from `parameter_metadata.get_type_metadata()` already contained merged docblock data (resolve inheritance etc..) ?

this kind of logic could be benefic to other attributes such as `@throws`, `@return` (and maybe others?)

---

I also added other code samples in the test (commented for now) that I'd wish mago could detect, maybe adding generic docblock propagation for all kind of attributes could help ?

azjezz:
This has been implemented in main.

Issue #523, opened 2025-10-06
analyser: Psl\Type\optional() not quite working as expected


bendavies:
### 🐞 Describe the Bug

I would expect the following code to trigger an error, but it does not

### 🔄 Steps to Reproduce

mago analyse

### ⚙️ Configuration (mago.toml)

```toml
n/a
```

### 📜 Command Output

```shell
help[type-inspection]: Type information for arguments of `Mago\inspect()` call.
   ┌─ mago.php:29:5
   │
29 │     \Mago\inspect($array['address']); // type: array{'city': string, 'country'?: string, 'street': string} <-- looks correct, `country` is optional
   │     ^^^^^^^^^^^^^ ----------------- Argument #1 type: `array{'city': string, 'country'?: string, 'street': string}`
   │     │
   │     Type inspection point
   │
   = The `Mago\inspect()` function is a static analysis debugging utility; it has no effect at runtime.
   = Help: Remember to remove `Mago\inspect()` calls before deploying to production.

help[type-inspection]: Type information for arguments of `Mago\inspect()` call.
   ┌─ mago.php:30:5
   │
30 │     \Mago\inspect($array['address']['country']); // type: string <-- looks wrong
   │     ^^^^^^^^^^^^^ ---------------------------- Argument #1 type: `string`
   │     │
   │     Type inspection point
   │
   = The `Mago\inspect()` function is a static analysis debugging utility; it has no effect at runtime.
   = Help: Remember to remove `Mago\inspect()` calls before deploying to production.
```

### 📂 PHP Code Sample (If Applicable)

```php
<?php declare(strict_types=1);


namespace {

    function get_mixed(): mixed
    {
        return 1;
    }

    function i_take_string(string $value): void
    {
        echo "Received string: $value\n";
    }


    $array_type = Psl\Type\shape([
        'name'    => Psl\Type\string(),
        'age'     => Psl\Type\int(),
        'address' => Psl\Type\shape([
            'street'  => Psl\Type\string(),
            'city'    => Psl\Type\string(),
            'country' => Psl\Type\optional(Psl\Type\string()),
        ]),
    ]);

    $array = $array_type->assert(get_mixed());

    \Mago\inspect($array['address']); // type: array{'city': string, 'country'?: string, 'street': string} <-- looks correct, `country` is optional
    \Mago\inspect($array['address']['country']); // type: string <-- looks wrong

    i_take_string ($array['address']['country']); // <--  should error, but does not

}

```

[...]

PR #527, opened 2025-10-06
feat(analyzer): add support for flow-php/types library

bendavies:
## 📌 What Does This PR Do?

Adds support for Flow-PHP's type system library (`flow-php/types`), enabling Mago to understand and analyze type assertions using Flow-PHP's `type_structure()` function.

## 🔍 Context & Motivation

Flow-PHP provides a type system library that has almost 1:1 mapping to PSL types but with slightly different APIs. This PR adds analyzer support so that Mago can properly infer types when using Flow-PHP's type assertions, similar to how PSL's `shape()` is already supported.

Key differences from PSL:
- PSL uses `shape(array $elements, bool $allow_unknown)` where optionals are wrapped in `optional()`
- Flow-PHP uses `type_structure(array $required, array $optional, bool $allow_extra)` with separate arrays

## 🛠️ Summary of Changes

- **Feature:** Added special function handlers for Flow-PHP type functions:
  - `Flow\Types\DSL\type_structure()` - Maps to PSL's `shape()` with support for required/optional element separation
- **Tests:** Comprehensive integration test covering:
  - Basic structures with required fields
  - Optional elements (2nd parameter)
  - `allow_extra` flag (3rd parameter)
  - Nested structures

## 📂 Affected Areas

- [x] Analyzer
- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation

## 📝 Notes for Reviewers

⚠️ **Note:** This PR was developed with heavy assistance from AI (Claude Code). I am not a Rust developer, so extra scrutiny on the Rust implementation, error handling, and code patterns would be greatly appreciated.

- The handler implementation mirrors PSL's `shape()` handler but adapts to Flow-PHP's separate required/optional arrays
- Flow-PHP's `type_structure()` only supports keyed arrays (string keys)
- Test includes `@mago-expect` assertions to verify optional fields are correctly flagged as possibly-undefined


norberttech:
That would be awesome! 
I have been looking into migrating to Mago since a while and this was pretty much the biggest blocker. 
My other projects would be unblocked as well 💪

crates/analyzer/tests/cases/flow_php_integration.php
```diff
@@ -0,0 +1,129 @@
...
+        ],
+        [],
+        true,
+    );
+
+    $array = $array_type->assert(get_mixed());
+    $flexible = $flexible_type->assert(get_mixed());
+
+    /* @mago-expect analysis:type-confirmation */
+    Mago\confirm($flexible, 'array{required_field: string, ...}');
```
azjezz: This actually does do anything, Mago will trigger `type-confirmation` issue regardless if the type matches or not.. we should probably add a utility that triggers only on failure.

However, for now, i suggest creating a function that expects `array{required_field: string, ...}` and passing it as an argument, which should yield no errors.

azjezz (changes requested):

crates/analyzer/tests/cases/flow_php_integration.php
bendavies: i tried this and it didnt work. prehaps due to https://github.com/carthage-software/mago/issues/523 ?

but my goal was to assert the exact shape. would a function with that signature do that?


btw, did the same here which you just merged
https://github.com/carthage-software/mago/pull/524/files#diff-902bbd72d0062ff88b0631f693d3d3e52ebff6096373ceb95382d4a03de74c7eR126

azjezz: You can assert in an out, i.e passing it as an argument, and returning it as a return type.

#523 has been fixed. you can try rebasing. but `Mago\confirm` does nothing here.

bendavies: thanks, i'll take a look

crates/analyzer/tests/cases/flow_php_integration.php
```diff
@@ -0,0 +1,124 @@
...
+
+    if (isset($array['address']['country'])) {
+        i_take_string($array['address']['country']);
+    }
+
+    $flexible = $flexible_type->assert(get_mixed());
+    \Mago\inspect($flexible);
+
+    $flexible2 = $flexible_type2->assert(get_mixed());
+    \Mago\inspect($flexible2);
```
bendavies: @azjezz can you help me with this please?

`$flexible` type is incorrect (`array<string, string>`)
`$flexible2` type is correct (`array{'required_field': string, ...}`)

`flexible_type` uses positional arguments
`flexible_type2` uses named arguments for `allow_extra`

i.e when `optional_elements` is specified. the function handler doesn't work

bendavies: ignore, solved it

crates/analyzer/tests/cases/flow_php_integration.php
bendavies: @azjezz please take another look.
added
```php
    /**
     * @param array{required_field: string, ...} $value
     * @return array{required_field: string, ...}
     */
    function i_take_flexible_array(array $value): array
    {
        return $value;
    }
```

which i think is what you mean?

PR #522, opened 2025-10-06
feat(linter): lint redundant readonly modifiers

dotdash:
In a readonly class, properties don't need their own readonly qualifiers.

## 📌 What Does This PR Do?

This adds a linter rule that checks for redundant `readonly` modifiers

## 🔍 Context & Motivation

Similar to the check for `final` modifiers, this adds a feature to detect redundant code.

## 🛠️ Summary of Changes

- **Feature:** Added a linter rule for redundant `readonly` properties.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

None

## 📝 Notes for Reviewers

None

PR #530, opened 2025-10-07
feat(linter): suggest `private` instead of `protected` visibility in final classes and enum

dotdash:
## 📌 What Does This PR Do?

Adds a lint and fix to the linter to detect protected visibility items in final classes and enums, suggesting them to be made private instead.

## 🔍 Context & Motivation

Without the possibility of child classes, there's no point in protected visibility.

## 🛠️ Summary of Changes

- **Feature:** Added a `no-protected-in-final` rule to the linter, including automatic fixing.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

This was suggested in PR #522

## 📝 Notes for Reviewers

I've taken care to remove write visibilites that become redundant when fixing the read visibility and added a secondary annotation for those cases. Please let me know if this should be handled differently.


azjezz:
You rock @dotdash !

PR #575, opened 2025-10-18
feat(analyzer): add a conditional return type for json_encode()

### LINKED ISSUE #425: Analyzer doesn't take the use of `JSON_THROW_ON_ERROR` into account.
### 🐞 Describe the Bug

When using `json_encode()` with the `JSON_THROW_ON_ERROR` flag, the analyzer still claims that the return value would be `string|false`.

When the `JSON_THROW_ON_ERROR` flag is used the return values is always `string` as an exception is thrown in case of an error.

### 🔄 Steps to Reproduce

.

```toml
.
```

### 📜 Command Output

```shell
.
```

```php

```

### 🖥️ Operating System

Linux

Installation Script (`curl | bash`)

### 📝 Additional Context

_No response_

mytskine:
## 📌 What Does This PR Do?

The default return type of PHP's json_encode() is string|false. But when the function receives the flag JSON_THROW_ON_ERROR, then the return type becomes string. So a new handler was created for ext/json, with a special case for the function json_encode().

## 🔍 Context & Motivation

Closes #425

## 🛠️ Summary of Changes

- **Bug Fix:** Fixed incorrect return type of PHP `json_encode(..., JSON_THROW_ON_ERROR)`.

## 📂 Affected Areas

- [x] Analyzer

## 🔗 Related Issues or PRs

#425

### DIFF
--- a/crates/analyzer/src/invocation/special_function_like_handler/standard/json.rs
+++ b/crates/analyzer/src/invocation/special_function_like_handler/standard/json.rs
@@ -1,0 +1,58 @@
+use mago_codex::ttype::atomic::TAtomic;
+use mago_codex::ttype::atomic::scalar::TScalar;
+use mago_codex::ttype::atomic::scalar::bool::TBool;
+use mago_codex::ttype::atomic::scalar::string::TString;
+use mago_codex::ttype::union::TUnion;
+
+use crate::artifacts::AnalysisArtifacts;
+use crate::context::Context;
+use crate::context::block::BlockContext;
+use crate::invocation::Invocation;
+use crate::invocation::special_function_like_handler::SpecialFunctionLikeHandlerTrait;
+use crate::invocation::special_function_like_handler::utils::get_argument;
+
+const JSON_THROW_ON_ERROR: i64 = 4194304;
+
+#[derive(Debug)]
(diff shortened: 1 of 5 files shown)


crates/analyzer/src/invocation/special_function_like_handler/standard/json.rs
```diff
@@ -0,0 +1,56 @@
...
+                let int_literal = int_argument_type.get_single_literal_int_value()?;
+
+                Some(if int_literal & JSON_THROW_ON_ERROR > 0 {
+                    TUnion::from_atomic(TAtomic::Scalar(TScalar::String(TString::unspecified_literal_with_props(
+                        false,
+                        false,
+                        true,
+                        false,
+                    ))))
+                } else {
```
azjezz, line 42: The resulting type should be a generic `string`, not a `literal-string`.

I think `non-empty-string` works too, AFAIK, all values ( even "empty" ones ) will produce a non-empty string.

azjezz (changes requested):

mytskine:
I replaced both call to `TString::unspecified_literal_with_props()` with `TString::new()` so that the string is just a non-empty string. Then I rebased on the latest main branch.

PR #600, opened 2025-10-28
fix(analyser): fix the try statement missing a return statement when calling a function returning never

### LINKED ISSUE #598: try-blocks are seen as missing a `return` statement when calling a function returning `never`
### 🐞 Describe the Bug

Calling a function that returns `never` is not treated as a valid way of exiting a function that returns non-void, instead a `missing-return-statement` diagnostic is produced.

### 🔄 Steps to Reproduce

1. Run `mago analyze` on the given example.
2. Observe the bogus `missing-return-statement` diagnostic.

```toml
php-version = "8.4.0"

[source]
paths = ["src"]
includes = ["vendor"]
excludes = []
```

### 📜 Command Output

```shell
❯ MAGO_LOG=trace cargo r -- --workspace play analyze src/never_in_try.php
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/mago --workspace play analyze src/never_in_try.php`
2025-10-27T20:15:07.774040Z DEBUG main mago::config: Sourcing global configuration from /home/bsteinbrink/mago.toml.
2025-10-27T20:15:07.774103Z DEBUG main mago::config: Sourcing workspace configuration from play/mago.toml.
2025-10-27T20:15:07.776476Z  INFO main mago::config: Overriding workspace directory with play.
2025-10-27T20:15:07.776499Z DEBUG main mago::config: Configuration specifies 32 threads.
2025-10-27T20:15:07.776513Z DEBUG main mago::config: Configuration specifies a stack size of 12582912 bytes.
error[missing-return-statement]: Missing return statement in function 'xxx'
   ┌─ src/never_in_try.php:7:10
   │  
 7 │   function xxx(): int
   │            ^^^ This function is declared to return 'int'...
 8 │ ╭ {
 9 │ │   try {
10 │ │     return 0;
11 │ │   } catch (Exception) {
12 │ │     a();
13 │ │   }
14 │ │ }
   │ ╰─' ...but this path can exit without returning a value.
   │  
   = A function that does not explicitly return a value will implicitly return `null`.
   = Help: Add a `return` statement that provides a value of type 'int' to all paths, or change the function's return type to 'int|null' and return `null` explicitly.

```

```php

function a(): never {
  die();
}

function xxx(): int
{
  try {
    return 0;
  } catch (Exception) {
    a();
  }
}
```

### 🖥️ Operating System

Linux

Built from Source

### 📝 Additional Context

_No response_

JeanDavidDaviet:
## 📌 What Does This PR Do?

The `all_catches_leave` variable is calculated before are analyzed, so it doesn't know that the function `a()` has return type never.
This PR re-calculates the control actions after analyzing the catch block statements, which now knows a() returns never.
As a newcomer, thanks to @dotdash for the hint to where to look to fix the bug

## 🔍 Context & Motivation

This is based on the reproducer in #598, in the repro, the catch call a function called `a()` and this function has a `never
` return type. 

```bash
error[missing-return-statement]: Missing return statement in function 'xxx'
   ┌─ src/never_in_try.php:7:10
   │  
 7 │   function xxx(): int
   │            ^^^ This function is declared to return 'int'...
 8 │ ╭ {
 9 │ │   try {
10 │ │     return 0;
11 │ │   } catch (Exception) {
12 │ │     a();
13 │ │   }
14 │ │ }
   │ ╰─' ...but this path can exit without returning a value.
   │  
   = A function that does not explicitly return a value will implicitly return `null`.
   = Help: Add a `return` statement that provides a value of type 'int' to all paths, or change the function's return type to 'int|null' and return `null` explicitly.

```
## 🛠️ Summary of Changes

- **Bug Fix:** Add a single line to recalculate all_catches_leave after updating catch block control actions.

## 📂 Affected Areas

- [x] Linter

## 🔗 Related Issues or PRs

Fixes #598 

## 📝 Notes for Reviewers

`just check` and `just test` are passing.

### DIFF
--- a/crates/analyzer/tests/cases/issue_598.php
+++ b/crates/analyzer/tests/cases/issue_598.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+function a(): never
+{
+    die();
+}
+
+function xxx(): int
+{
+    try {
+        return 0;
+    } catch (Exception) {
+        a();
+    }
+}
--- a/crates/analyzer/tests/mod.rs
+++ b/crates/analyzer/tests/mod.rs
@@ -263,6 +263,7 @@ test_case!(issue_549);
 test_case!(issue_551);
 test_case!(issue_577);
 test_case!(issue_574);
+test_case!(issue_598);
 
 #[test]
 fn test_all_test_cases_are_ran() {


crates/analyzer/src/statement/try.rs
```diff
@@ -209,6 +209,8 @@ impl<'ast, 'arena> Analyzable<'ast, 'arena> for Try<'arena> {
                     true,
                 );
             }
+            
+            all_catches_leave = catch_actions.iter().all(|actions| !actions.contains(&ControlAction::None));
```
dotdash, line 213: The previous value of `all_catches_leave` is already used on line 136. I was unsure whether it is enough to just recalculate it here, or if we already need the correct value earlier and this needs some more involved fix.

PR #615, opened 2025-11-07
fix(analyzer): don't assert that $y is not null on isset($x[$y])

### LINKED ISSUE #586: `isset($x[$y])` incorrectly asserts that `$y` is not null
### 🐞 Describe the Bug

The fix for #573 made it so that for `isset($x[$y])` to succeed, it is asserted that `$y` needs to be non-null. But this is wrong. PHP silently casts `null` to `''` (the empty string) when using it to perform an array access.

From the [docs](https://www.php.net/manual/en/language.types.array.php):

> Additionally the following key casts will occur:
> 
> [String](https://www.php.net/manual/en/language.types.string.php)s containing valid decimal [int](https://www.php.net/manual/en/language.types.integer.php)s, unless the number is preceded by a + sign, will be cast to the [int](https://www.php.net/manual/en/language.types.integer.php) type. E.g. the key "8" will actually be stored under 8. On the other hand "08" will not be cast, as it isn't a valid decimal integer.
> [Float](https://www.php.net/manual/en/language.types.float.php)s are also cast to [int](https://www.php.net/manual/en/language.types.integer.php)s, which means that the fractional part will be truncated. E.g. the key 8.7 will actually be stored under 8.
> [Bool](https://www.php.net/manual/en/language.types.boolean.php)s are cast to [int](https://www.php.net/manual/en/language.types.integer.php)s, too, i.e. the key [true](https://www.php.net/manual/en/reserved.constants.php#constant.true) will actually be stored under 1 and the key [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) under 0.
> [Null](https://www.php.net/manual/en/language.types.null.php) will be cast to the empty string, i.e. the key null will actually be stored under "".

### 🔄 Steps to Reproduce

1. Run `mago analyze` on the given example
2. Observe that the `if ($y === null)` as marked with `impossible-condition`
3. Run the given code using the PHP interpreter
4. Observe that the condition is actually true and `123` appears in the output

```toml
php-version = "8.4.0"

[source]
paths = ["src"]
includes = ["vendor"]
excludes = []
```

### 📜 Command Output

```shell
❯ cargo run -- --workspace play analyze src/keynull.php
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
     Running `target/debug/mago --workspace play analyze src/keynull.php`
2025-10-20T10:49:00.839715Z DEBUG main mago::config: Sourcing global configuration from /home/bsteinbrink/mago.toml.
2025-10-20T10:49:00.839777Z DEBUG main mago::config: Sourcing workspace configuration from play/mago.toml.
2025-10-20T10:49:00.842131Z  INFO main mago::config: Overriding workspace directory with play.
2025-10-20T10:49:00.842152Z DEBUG main mago::config: Configuration specifies 32 threads.
2025-10-20T10:49:00.842166Z DEBUG main mago::config: Configuration specifies a stack size of 12582912 bytes.
help[redundant-comparison]: Redundant `===` comparison: left-hand side is never identical to right-hand side.
  ┌─ src/keynull.php:8:9
  │
8 │     if ($y === null) {
  │         ^^     ---- Right operand is `null`
  │         │       
  │         Left operand is `string`
  │
  = The `===` operator will always return `false` in this case.
  = Help: Consider simplifying or removing this comparison as it always evaluates to `false`.

warning[impossible-condition]: This condition (type `false`) will always evaluate to false.
  ┌─ src/keynull.php:8:9
  │
8 │     if ($y === null) {
  │         ^^^^^^^^^^^ Expression of type `false` is always falsy
  │
  = Because this condition is always false, the code block it controls will never be executed.
  = Help: Check the logic of this expression. If the code block is intended to be unreachable, consider removing it. Otherwise, revise the condition.

warning: found 2 issues: 1 warning(s), 1 help message(s)
```

```php

function gar(?string $y): void {
  $x = ['' => 2];

  if (isset($x[$y])) {
    echo $x[$y], PHP_EOL;
    if ($y === null) {
      echo 123, PHP_EOL;
    }
  }
}

gar(null);
```

### 🖥️ Operating System

Linux

Built from Source

### 📝 Additional Context

_No response_

Bleksak:
fixes #586

### DIFF
--- a/crates/analyzer/src/assertion.rs
+++ b/crates/analyzer/src/assertion.rs
@@ -113,25 +113,6 @@ pub fn scrape_assertions(
             }
             Construct::Isset(isset_construct) => {
                 for value in isset_construct.values.iter() {
-                    // First, handle assertions for nullable array indices
-                    // For isset($arr[$key]) where $key is nullable, assert that $key is non-null
-                    let mut current_expr = value;
-                    while let Expression::ArrayAccess(array_access) = current_expr {
-                        if let Some(index_id) = get_expression_id(
-                            array_access.index,
-                            assertion_context.this_class_name,
-                            assertion_context.resolved_names,
-                            Some(assertion_context.codebase),
-                        ) && let Some(index_type) = artifacts.get_expression_type(array_access.index)
-                            && index_type.is_nullable()
-                        {
-                            if_types.entry(index_id).or_insert_with(|| vec![vec![Assertion::IsNotType(TAtomic::Null)]]);
-                        }
-
-                        current_expr = array_access.array;
-                    }
-
-                    // Then handle the value itself
                     if let Some(value_id) = get_expression_id(
                         value,
                         assertion_context.this_class_name,
--- a/crates/analyzer/src/utils/expression/array.rs
+++ b/crates/analyzer/src/utils/expression/array.rs
@@ -105,7 +105,7 @@ pub(crate) fn get_array_target_type_given_index<'ctx, 'arena>(
         return get_never();
     }
 
-    if index_type.is_null() {
+    if index_type.is_null() && !index_type.is_keyed_array() {
         context.collector.report_with_code(
             IssueCode::NullArrayIndex,
             Issue::error(format!(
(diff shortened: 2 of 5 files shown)


azjezz:
@Bleksak thanks! can you please rebase? 💛

PR #618, opened 2025-11-08
fix(prelude): Fix generic typings for DOMNamedNodeMap and DOMNode::$attributes

Zuruuh:
## 📌 What Does This PR Do?

See https://github.com/JetBrains/phpstorm-stubs/blob/master/dom/dom_c.php#L101 and https://github.com/JetBrains/phpstorm-stubs/blob/master/dom/dom_c.php#L1393 for comparison :+1:
## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation

### DIFF
--- a/crates/prelude/assets/extensions/dom.php
+++ b/crates/prelude/assets/extensions/dom.php
@@ -355,6 +355,7 @@ namespace {
 
         /**
          * @readonly
+         * @var DOMNamedNodeMap<DOMAttr>
          */
         public null|DOMNamedNodeMap $attributes;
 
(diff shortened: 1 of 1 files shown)


crates/prelude/assets/extensions/dom.php
```diff
@@ -355,6 +355,7 @@ class DOMNode
 
         /**
          * @readonly
+         * @var DOMNamedNodeMap<DOMAttr>
```
Girgias, line 358: This is incorrect, it can be a named node map of `DOMDocumentType` which is NOT a `DOMAttr`.

Honestly, the thing that should be promoted is to migrate to the new 8.4 DOM API rather than fixing the stubs of the old API in correct ways, as it fixes a lot of this nonsense.

Issue #693, opened 2025-12-03
`literal-named-argument` should not be reported for Pest's `test` and `it` methods


innocenzi:
I think a safelist should be added for methods or functions that should be ignored by this rule:

```php
it('gets subscriptions', function () {
// ...
```

```
warning[literal-named-argument]: Literal argument `'gets subscriptions'` should be passed as a named argument for clarity.
   ┌─ src/Modules/Invoices/InboundEmails/MicrosoftGraph/ListSubscriptionsRequestTest.php:11:4
   │
11 │ it('gets subscriptions', function () {
   │    ^^^^^^^^^^^^^^^^^^^^ This literal is being passed positionally.
   │
   = Passing literals positionally can make code less clear, especially with booleans, numbers, or `null`.
   = Help: Consider using a named argument instead: `function_name(param: 'gets subscriptions')`.
```

EDIT: actually, the first argument can almost always be ignored. It's more consistent than a safelist

PR #694, opened 2025-12-03
fix(linter): ignore first argument during `literal-named-argument` check

innocenzi:
Closes https://github.com/carthage-software/mago/issues/693

This pull request updates the behavior of the `literal-named-argument` check to ignore the first parameter. In most situations, the first parameter of functions and methods is clear. It only gets confusing when the other arguments are string/int/boolean.

For instance, `data_set($array, 'foo')` could use a named argument. But `config('autorouter.base_url')` is clear enough.


azjezz:
I agree with this, but one change: adding an option to preseve the old behavior :)

PR #703, opened 2025-12-04
feat(linter): add `property-name` rule

innocenzi:
This pull request adds support for class properties (including promoted ones) in the `variable-name` rule.

I added an option to specifically disable checking for class properties but not promoted properties to "support" the situation where we have to use inherited properties from underlying classes, or if a framework expects a specific property (eg. `$maxExceptions` in Laravel jobs).

By the way, I think this rule should be named `variable-name-casing` and that the `camel` parameter should actually be a `style` parameter with either `camel-case` or `kebab-case` as options. I can PR that if you want? It would be breaking, but it's been introduced just one version ago and was disabled by default, so I think it'd be fine.

### DIFF
--- a/crates/linter/src/rule/consistency/mod.rs
+++ b/crates/linter/src/rule/consistency/mod.rs
@@ -14,6 +14,7 @@ pub mod no_alias_function;
 pub mod no_hash_comment;
 pub mod no_php_tag_terminator;
 pub mod no_trailing_space;
+pub mod property_name;
 pub mod trait_name;
 pub mod variable_name;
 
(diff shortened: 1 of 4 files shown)


azjezz:
I think this is better as a new rule, rather than a setting here, i want to enforce property as `$camelCase` in my project, but i dont care about variable names for example

innocenzi:
I added the option to disable checking variables, I think it's better than kind of duplicating the rule? This way, with that rule you have total control over what you want to check.

azjezz:
this now means that the rule can be completely disabled using its configurations, i don't like this, i think having a seperate rule is better...

PR #713, opened 2025-12-07
fix(formatter): make `empty_line_after_opening_tag` default consistent

magic-akari:
## 📌 What Does This PR Do?

Fixes inconsistency between the `Default` trait implementation and serde default annotation for `empty_line_after_opening_tag` formatter setting.

## 🔍 Context & Motivation

While integrating the WASM tooling and migrating test cases, I discovered that `empty_line_after_opening_tag` had conflicting default values:
- The serde default annotation pointed to `default_true()` (returning `true`)
- The `Default` trait implementation set it to `false`

This inconsistency could cause different behavior depending on whether the struct was instantiated via `Default::default()` or deserialized without the field present, leading to unpredictable formatting results.

## 🛠️ Summary of Changes

- **Bug Fix:** Aligned `empty_line_after_opening_tag` default value in `Default` implementation to match serde default (`true`)

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

The documented default in the code comments was already `true`, and the serde annotation was correct. Only the `Default` impl was wrong. This ensures consistent behavior across different initialization methods.

### DIFF
--- a/crates/formatter/tests/cases/html_template/after.php
+++ b/crates/formatter/tests/cases/html_template/after.php
@@ -7,6 +7,7 @@
 <body>
     <h1>Welcome to the Jungle!</h1>
     <?php
+
     $level1 = 'Level 1';
     ?>
     <div class="level1">
@@ -21,11 +22,13 @@
             <?php echo $level2; ?>
             <ul>
                 <?php
+
                 $items = ['item1', 'item2', 'item3'];
                 foreach ($items as $item): ?>
                 <li>
                     <?php echo $item; ?>
                     <?php
+
                     $level3 = 'Level 3';
                     ?>
                     <span class="level3">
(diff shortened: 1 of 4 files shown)


azjezz:
Thank you for the bug fix!

Unrelated: i looked at https://github.com/wasm-fmt/mago_fmt and its cool! one note: it is using mago 0.x version, which has a lot of bugs and would produce invalid php code in many cases, i would highly recommand switching to 1.0.0-rc.x, even if not "stable", it is more stable ( espically the formatter ) compared to 0.x

PR #750, opened 2025-12-14
fix(formatter): sort top-level uses with `expand-use-groups = false`

### LINKED ISSUE #730: [Formatter] Top-level `use`s aren't sorted with `expand-use-groups = false`
### Description

With `expand-use-groups = false`, the top-level `use` statements aren't sorted. However the groups themselves are.

### Playground Link

_No response_

### Reproduction

1. Create a file with the contents below
2. Run `mago format` on it and see that the `use` statements aren't sorted

Before:

```php
<?php

use Sample\B;
use Example\{C, A};
use function Sample\e;
use function Example\{f, d};
use const Sample\H;
use const Example\{I, G};
```

After:

```php
<?php

use Sample\B;
use Example\{A, C};
use function Sample\e;
use function Example\{d, f};
use const Sample\H;
use const Example\{G, I};
```

### Environment

Mago 1.0.0-rc.11
OS: MacOS
Installed via Composer

edsrzf:
## 📌 What Does This PR Do?

Fixes #730 by setting `name` and `namespace`. Otherwise all the items sort equally.

## 🔍 Context & Motivation

Fixing a bug.

## 🛠️ Summary of Changes

- **Bug Fix:** Fix sorting of uses when groups aren't expanded.
- **Refactor:** Factor out name/namespace extraction functions.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #730

### DIFF
--- a/crates/formatter/tests/cases/use_sorting_with_grouping_and_separation/after.php
+++ b/crates/formatter/tests/cases/use_sorting_with_grouping_and_separation/after.php
@@ -1,0 +1,10 @@
+<?php
+
+use Example\{A, C};
+use Sample\B;
+
+use function Example\{d, f};
+use function Sample\e;
+
+use const Example\{G, I};
+use const Sample\H;
--- a/crates/formatter/tests/cases/use_sorting_with_grouping_and_separation/before.php
+++ b/crates/formatter/tests/cases/use_sorting_with_grouping_and_separation/before.php
@@ -1,0 +1,8 @@
+<?php
+
+use Sample\B;
+use function Sample\e;
+use const Sample\H;
+use function Example\{f, d};
+use Example\{C, A};
+use const Example\{I, G};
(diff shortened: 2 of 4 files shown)


crates/formatter/tests/cases/use_sorting_with_grouping/settings.inc
```diff
@@ -0,0 +1,6 @@
+FormatSettings {
+    sort_uses: true,
+    expand_use_groups: false,
+    separate_use_types: false,
```
azjezz, line 4: can we also add a test where this is true?

azjezz:
Thank you @edsrzf

PR #759, opened 2025-12-16
allow ReflectionProperty constructor to accept key-of<properties-of<T>>

Bleksak:
this pr allows us to pass key-of<properties-of\<T>> to ReflectionProperty

fixes this issue: https://mago.carthage.software/playground#019b28ed-b476-569a-99b0-a315a3c60667

### DIFF
--- a/crates/prelude/assets/extensions/reflection.php
+++ b/crates/prelude/assets/extensions/reflection.php
@@ -572,8 +572,10 @@ class ReflectionProperty implements Reflector
     public string $class;
 
     /**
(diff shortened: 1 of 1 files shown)


azjezz:
This is not the right solution, properties-of<T> should be expanded to `array<non-empty-string, mixed>`, and key-of that is `non-empty-string` so the code should work, this is a bug!

Bleksak:
Unknowingly found a bug 😃 thank you for fixing it.

PR #802, opened 2025-12-21
Statement/docblocks: always populate docblock variables

### LINKED ISSUE #801: analyzer: var doc comment is still tied to a statement
### Description

2 playground links:
https://mago.carthage.software/playground#019b3e0b-0821-787c-55f2-138cb5a3624c

reports 2 errors about Config and Session

https://mago.carthage.software/playground#019b3e0c-1386-d5f2-0c3b-e09a7ef5926e

reports no issue at all

both should behave the same though

### Playground Link

_No response_

### Reproduction

_No response_

### Environment

playground 1.0.1

Bleksak:
partially fixes #801

### DIFF
--- a/crates/analyzer/src/statement/mod.rs
+++ b/crates/analyzer/src/statement/mod.rs
@@ -68,47 +68,28 @@ impl<'ast, 'arena> Analyzable<'ast, 'arena> for Statement<'arena> {
             }
         }
 
-        let should_populate_docblock_variables = matches!(
-            self,
-            Statement::Block(_)
-                | Statement::Expression(_)
-                | Statement::Try(_)
-                | Statement::Continue(_)
-                | Statement::Break(_)
-                | Statement::Return(_)
-                | Statement::Echo(_)
-                | Statement::EchoTag(_)
-                | Statement::Unset(_)
-                | Statement::Inline(_)
-                | Statement::OpeningTag(_)
-                | Statement::Declare(_)
-                | Statement::Noop(_)
-                | Statement::ClosingTag(_)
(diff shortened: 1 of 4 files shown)


Bleksak:
@azjezz i checked some of the Psl code, and it's exactly this pattern. Is Psl code wrong or my code? Mago did not pick up on the doc comments before so I think Psl code must be wrong

azjezz:
This breaks `foreach` `@var`, hence why Psl is failing.

azjezz:
We should have a special case for `foreach` like we do for assignment, where we do not populate `$k` and `$v` in `foreach ($it as $k => $v)`

Bleksak:
But isn't that correct behavior? If you change the var type to be of T|null then it works(and is actually correct since you assigned null before, and if the array is empty, then it stays null)

Bleksak:
fixed - inside foreach the vars are nonnull, outside they can be null
is there any other statement that would have it like this?

azjezz:
@Bleksak can you please add tests for things that broke in previous iteration in Psl? this ensures we don't accidently break them in the future :)

Bleksak:
tests added, found another bug (that's why var T is in the Iterate class of the test)

azjezz:
the PR description mentions "partially fixes #801" why is it partial? what do you think this PR doesn't handle?

Bleksak:
well, it still is tied to a statement.. it just fixes this case (and probably many more), i'm not sure if that is enough, that's probably up to you to decide

Bleksak:
i removed the `@var T` and just used `@var array-key`

PR #821, opened 2025-12-22
feat(linter): add rule for no assignment in function parameters

chrisopperwall-qz:
## 📌 What Does This PR Do?

This PR adds a no_assign_in_argument rule to lint similar use cases to no_assign_in_condition, where an assignment expression within a function call argument is not allowed. 

## 🔍 Context & Motivation

This rule would be helpful to warn or error in cases where it's clearer to assign a variable in its own statement, and then use that variable in the function call, instead of nesting the assignment expression in the function call argument.

## 🛠️ Summary of Changes

- **Feature:** Added `correctness/no_assign_in_argument ` rule.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [x] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

The inspiration for this change is this hhast-lint rule which discourages assignment expressions in function call arguments because
* users may assume that new variable isn't available in the current scope/context.
* it looks like keyword args, which exist now via a different syntax in PHP 8.

https://github.com/hhvm/hhast/blob/v3.27.4/src/Linters/NoBasicAssignmentFunctionParameterLinter.php

I'm a little new to rust, so I mostly followed the example of `correctness/no_assign_in_condition`, so if there's any feedback I'm happy to make changes!


crates/linter/src/rule/correctness/no_assign_in_argument.rs
```diff
@@ -0,0 +1,242 @@
...
+                $x = 5;
+                foo($x);
+            "},
+            bad_example: indoc! {r"
+                <?php
+
+                foo($x = 5);
+            "},
+            category: Category::Correctness,
+
```
azjezz:
```suggestion
```

crates/linter/src/rule/correctness/no_assign_in_argument.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    fn check<'arena>(&self, ctx: &mut LintContext<'_, 'arena>, node: Node<'_, 'arena>) {
+        let argument_list = match node {
+            Node::FunctionCall(call) => &call.argument_list,
+            Node::MethodCall(call) => &call.argument_list,
+            Node::NullSafeMethodCall(call) => &call.argument_list,
+            Node::StaticMethodCall(call) => &call.argument_list,
+            _ => return,
+        };
```
azjezz:
```suggestion
        let Node::ArgumentList(argument_list) = node else {
          return;
        };
```

crates/linter/src/rule/correctness/no_assign_in_argument.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    fn targets() -> &'static [NodeKind] {
+        const TARGETS: &[NodeKind] = &[
+            NodeKind::FunctionCall,
+            NodeKind::MethodCall,
+            NodeKind::NullSafeMethodCall,
+            NodeKind::StaticMethodCall,
+        ];
+
+        TARGETS
```
azjezz:
```suggestion
        const TARGETS: &[NodeKind] = &[
            NodeKind::ArgumentList,
        ];

        TARGETS
```

crates/linter/src/rule/correctness/no_assign_in_argument.rs
```diff
@@ -0,0 +1,242 @@
...
+            let value = argument.value();
+
+            if let Some(assignment) = get_assignment_from_expression(value) {
+                let mut issue = Issue::new(self.cfg.level(), "Avoid assignments in function call arguments.")
+                    .with_code(self.meta.code)
+                    .with_annotation(
+                        Annotation::primary(assignment.span()).with_message("This is an assignment"),
+                    )
+                    .with_annotation(
+                        Annotation::secondary(argument.span()).with_message("This is the argument"),
```
azjezz:
```suggestion
                        Annotation::secondary(argument_list.span()).with_message("In this argument list"),
```

crates/linter/src/rule/correctness/no_assign_in_argument.rs
```diff
@@ -0,0 +1,242 @@
...
+    fn default() -> Self {
+        Self { level: Level::Warning }
+    }
+}
+
+impl Config for NoAssignInArgumentConfig {
+    fn level(&self) -> Level {
+        self.level
+    }
+}
```
azjezz, line 52: New rules should be disabled by default.

```suggestion
impl Config for NoAssignInArgumentConfig {
    fn default_enabled() -> bool {
        false
    }

    fn level(&self) -> Level {
        self.level
    }
}
```

docs/tools/linter/rules/correctness.md
```diff
@@ -188,39 +184,22 @@ before a complete statement at the top level of a block or file.
...
-$arr = [ // @mago-format-ignore-next
-    1,
-    2,
-];
+echo 'Hello, world!';
 ```
 
 
-## <a id="ineffective-format-ignore-region"></a>`ineffective-format-ignore-region`
+## <a id="no-assign-in-argument"></a>`no-assign-in-argument`
```
azjezz, line 301: Please make sure to generate documentation using our script, not manually.

```
$ scripts/linter-rules-docs.php
```

azjezz:
You can use `just fix` to fix clippy and/or formatting that are causing build failure now

crates/linter/src/rule/correctness/no_assign_in_argument.rs
azjezz: This is more accurate as it will also catch arguments passed to instantiation ( e.g `$x = new Foo($b = 1);` and any other place where an argument list might be used.

Also, it might be a good idea to also target `PartialArgumentList` to support upcoming PHP 8.6, e.g `$x = foo(?, ?, $b = 1);`

chrisopperwall-qz:
Thanks, @azjezz! I'll address those comments soon

crates/linter/src/rule/correctness/no_assign_in_argument.rs
chrisopperwall-qz: I gave the PartialArgumentList support and attempt in [9a09671](https://github.com/carthage-software/mago/pull/821/commits/9a09671a4e4de403f4a9c38c598cc93b93eaf251) 😄

crates/linter/src/rule/correctness/no_assign_in_argument.rs
```diff
@@ -0,0 +1,281 @@
...
+        count = 1,
+        code = indoc! {r"
+            <?php
+
+            foo($a, $b = 5, $c);
+        "}
+    }
+
+    test_lint_failure! {
+        name = multiple_arguments_with_assignment_in_parial_arguments,
```
azjezz:
```suggestion
        name = multiple_arguments_with_assignment_in_partial_arguments,
```

azjezz:
Thank you @chrisopperwall-qz !

Issue #833, opened 2025-12-26
use mago in pre-commit.com hooks


matlads:
### Feature

I'd like to add code to mago so that integration with [pre-commit.com](https://pre-commit.com) style hooks is easier. `pre-commit` is a framework for managing and maintaining _multi-language_ pre-commit hooks. One of the supported languages is *rust*, which mago is written in.

Since the mago git repo results in an installable package or exposes an executable, it can be used with pre-commit. All that is required in the repo is too a small YAML file `.pre-commit-hooks.yaml` . An example would be something like so:

```yaml
---
# mago pre-commit-hooks
 - id: mago-lint
   name: mago lint
   description: Lint using mago an extremely fast PHP linter, formatter, and static analyzer, written in Rust.
   entry: mago lint
   language: rust
   types: [php]

 - id: mago-fmt
   name: mago fmt
   description: format files using mago an extremely fast PHP linter, formatter, and static analyzer, written in Rust.
   entry: mago fmt
   language: rust
   types: [php]
```

### Example

```php
pre-commit run mago-lint --verbose --all-files
pre-commit run mago-fmt --verbose --all-files
```

azjezz:
see https://github.com/carthage-software/mago/pull/846#issuecomment-3697131896

PR #846, opened 2025-12-29
feat: add .pre-commit-hooks.yaml

matlads:
## 📌 What Does This PR Do?

This PR introduces a yaml file that allows mago to easily be used by users of pre-commit.com style of pre-commit hooks.

## 🔍 Context & Motivation

pre-commit allows the user to run a large set of hooks that help in maintaining a git repo and git managed code base. Ensuring that integrating `mago` is easy to do, only helps `mago` and the users of pre-commit hooks

## 🛠️ Summary of Changes

- **Feature:** Allow mago-lint and mago-fmt to be used in pre-commit hooks

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): Integration

## 🔗 Related Issues or PRs
Fixes: #833 

## 📝 Notes for Reviewers

N/A

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,15 @@
+---
+
+ - id: mago-lint
+   name: mago lint
+   description: An extremely fast PHP linter, formatter, and static analyzer, written in Rust.
+   entry: mago lint
+   language: rust
+   types: [php]
+
+ - id: mago-fmt
+   name: mago fmt
+   description: An extremely fast PHP linter, formatter, and static analyzer, written in Rust.
+   entry: mago fmt
+   language: rust
+   types: [php]


azjezz:
I don't think this is a good addition. 

Simply, because it won't work for the majority of projects.

The reason is, the binary `mago` might be in a different location, not globally available as `mago`, e.g, if Mago is installed using composer, it would be in `vendor/bin/mago` ( which is not really a binary, but rather a thin PHP wrapper ); Another reason is that projects could be placing `mago.toml` in a different directory, which would require `mago --config path/to/mago.toml lint`..etc.

Overall, i see no benifits of adding this, and i would rather people solve this problem in their own projects.

Issue #854, opened 2026-01-04
Drupal support (mostly formatter)


klausi:
Hi,

I would like to explore using mago for Drupal code. Drupal has a bit of a custom formatting standard and Drupal-specific rules for analyzing via PHPStan.

Example: #852 

Do you want to support Drupal in mago? If yes I can keep sending pull requests.

I fully understand if you don't want to support Drupal. It will increase your maintenance burden (but I'm happy to help!). I can also look into forking mago for Drupal or check if mago has some plugin system so that I can extend it from the outside.

Thanks for starting this project, the runtime performance is amazing!

For people testing with Drupal here is an examplemago.toml:
```toml
# Welcome to Mago!
# For full documentation, see https://mago.carthage.software/tools/overview
php-version = "8.4.0"

[source]
workspace = "."
paths = ["../behat", "phpstan", "web/modules/custom", "web/profiles"]
includes = ["vendor"]
excludes = []
extensions = ["php", "inc", "module", "profile", "install", "theme"]

[formatter]
classlike-brace-style = "same_line"
empty-line-after-class-like-open = true
empty-line-before-class-like-close = true
else-catch-on-newline = true
function-brace-style = "same_line"
method-brace-style = "same_line"
preserve-breaking-argument-list = true
preserve-breaking-array-like = true
preserve-breaking-attribute-list = true
preserve-breaking-conditional-expression = true
preserve-breaking-member-access-chain = true
preserve-breaking-parameter-list = true
print-width = 120
tab-width = 2
uppercase-literal-keyword = true
use-tabs = false

[linter]
integrations = ["drupal", "symfony", "phpunit"]

[linter.rules]
ambiguous-function-call = { enabled = false }
halstead = { effort-threshold = 7000 }
literal-named-argument = { enabled = false }
lowercase-keyword = { enabled = false }
no-empty = { enabled = false }

[analyzer]
find-unused-definitions = true
find-unused-expressions = true
analyze-dead-code = true
memoize-properties = true
allow-possibly-undefined-array-keys = true
check-throws = false
check-missing-override = true
find-unused-parameters = false
strict-list-index-checks = false
no-boolean-literal-comparison = false
check-missing-type-hints = true
register-super-globals = true
```
You need to run the dev version of mago and compile it.

azjezz:
Hi @klausi !

Yes! We would love to!

I don't have experience with using Drupal, so i don't really know what formatting setting, or linter rules make sense, but we welcome PRs!

[...]

PR #860, opened 2026-01-05
feat(formatter): Add following_clause_on_newline option to put else, catch, finally on a new line

klausi:
## 📌 What Does This PR Do?

This PR introduces a new configuration option: `following_clause_on_newline`. 

When enabled, this setting forces `else`, `elseif`, `catch`, and `finally` keywords to be placed on a new line, rather than sharing the line with the preceding closing brace (`}`).

## 🔍 Context & Motivation

This change is required to support coding standards that mandate control structure keywords appear on their own lines, most notably **Drupal** coding standards. 

While the default behavior (and standards like PER-CS) places these keywords on the same line, this option provides an opt-in mechanism for users whose style guides require the alternative formatting.

## 🛠️ Summary of Changes

- **Configuration:** Added `following_clause_on_newline` to `FormatSettings` (defaults to `false` to preserve backward compatibility).
- **Formatter Logic:** Updated `misc.rs` and `mod.rs` to inject line breaks for `if`/`else`/`elseif` and `try`/`catch`/`finally` blocks when the option is active.
- **Tests:** Added a new test suite, `following_clause_on_newline`, covering various control structure scenarios.
- **Documentation:** Updated `configuration-reference.md` to document the new option.

## 📂 Affected Areas

- [x] Formatter Engine
- [x] Documentation

## 🔗 Related Issues

#854

## 📝 Notes for Reviewers

- **Do-While Loops:** The `do-while` structure is intentionally **excluded** from this rule. The `while` keyword will remain on the same line as the closing brace, as this structure is conventionally kept on one line even in standards that require `else` on a new line.


crates/formatter/src/settings.rs
```diff
@@ -89,6 +89,12 @@ pub struct FormatSettings {
     #[serde(default = "BraceStyle::same_line")]
     pub control_brace_style: BraceStyle,
 
+    /// Whether to place `else`, `elseif`, `catch` and `finally` on a new line.
+    ///
+    /// Default: false
+    #[serde(default = "default_false")]
+    pub else_catch_on_newline: bool,
```
azjezz: I think a better name is `following_clause_on_newline` as this does not apply to just `else` and `catch`, but any following clause.

klausi: thanks for reviewing, done!

PR #881, opened 2026-01-13
fix(analyzer): add is_expanded to named_object to prevent cycles

### LINKED ISSUE #880: bug(analyzer): static return does not see parent methods
### Description

playground is broken atm

```php
<?php

declare(strict_types=1);

class X
{
    public function clone(): static
    {
        return clone $this;
    }
}

class XX extends X
{
    public function a_method(): string
    {
        return 'abc';
    }
}

$xx = new XX();
$a = $xx->clone()->a_method();
```

### Playground Link

_No response_

### Reproduction

_No response_

### Environment

playground

Bleksak:
## 📌 What Does This PR Do?

fixes #880

## 📝 Notes for Reviewers
Not sure if this is a correct fix for this issue, but it works. The underlying issue is probably that the code is getting executed multiple times and the named object will get overwritten many times.

### DIFF
--- a/SPONSORS.md
+++ b/SPONSORS.md
@@ -21,7 +21,7 @@ To become a sponsor, please visit [the sponsorship page](https://github.com/spon
 
 ---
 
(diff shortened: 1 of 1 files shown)


dynasource:
thx

PR #898, opened 2026-01-16
Require import of global symbols

MidnightDesign:
## 📌 What Does This PR Do?

Adds a formatter rule that requires global symbols like constants and functions to be imported.

## 🔍 Context & Motivation

Using fully qualified symbol names is a small performance and maintainability win because PHP doesn't have to check the current namespace for it. Importing symbols vs. `\strlen()` just looks nicer IMO.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

It's arguable whether this is a formatter or a linter feature. Maybe the import of already-fully-qualified symbols is a formatter thing and the no-lookup part is a linter thing. To be discussed.

## 📝 Notes for Reviewers

I haven't implemented anything yet. Just added the tests as a basis for discussion. (I also don't know if I even _could_ implement it given my limited Rust experience.)


Bleksak:
This should either be a linter or analyzer rule. Not a formatter setting. The linter can already detect these cases, it would just need a --fix option and possibly mark as unsafe. However I think both the formatter and linter don't have access to the codebase. So they don't know that "strlen" or any other symbol is in the global namespace.

MidnightDesign:
(Yeah, this probably should have been an issue instead of a PR. I just feel like a PR with failing test cases shows a bit more effort. : )

Whatever part of the toolchain this fit is definitely up for debate. My perspective was just that this is a missing feature compared to PHP CS Fixer, which is just a formatter.

Anyway, this is a legitimate use case, right?

azjezz:
This is now fully covered ( including fixes ) by the linter.

PR #901, opened 2026-01-16
fix(member_access): Fix multi line indentation in if clause

klausi:
## 📌 What Does This PR Do?

Fixes an issue where chained method calls inside control structure conditions (like `if` statements) were being formatted with incorrect indentation (double indent).

## 🔍 Context & Motivation

When a method chain (e.g., `$this->foo()->bar()`) appeared inside an `if` condition, the formatter was applying two levels of indentation: one from the condition wrapper and one from the chain formatter itself. Additionally, heuristics for "short chains" were preventing proper inlining, causing readable code to break awkwardly across multiple lines.

This change ensures that chains inside conditions respect the parent indentation context and inline short chains appropriately, matching the expected output in the `if_multiline_operator` test case and general PHP coding standards.

This test code:
```php
if (!$this->windows && !$this->getSession()
    ->getDriver() instanceof BrowserKitDriver) {
    if (!$this->getSession()->isStarted()) {
        $this->getSession()->start();
    }

    $this->windows = $this->getSession()->getWindowNames();
}
```
with settings
```toml
```
is wrongly formatted to:
```php
if (!$this->windows && !$this->getSession()
        ->getDriver() instanceof BrowserKitDriver) {
    if (!$this->getSession()->isStarted()) {
        $this->getSession()->start();
    }

    $this->windows = $this->getSession()->getWindowNames();
}
```

## 🛠️ Summary of Changes

- **Bug Fix:** Prevented double indentation for member access chains when they appear inside a condition context.
- **Bug Fix:** Adjusted `is_already_broken` logic to allow short chains (length <= 3) to be formatted as chains when inside conditions.
- **Bug Fix:** Updated `print_member_access_chain` to force inlining of the first access for short chains in conditions, ensuring `$object->method()` stays together on the first line where possible.

## 📂 Affected Areas

- [x] Formatter

## 🔗 Related Issues or PRs

-

## 📝 Notes for Reviewers

- The logic specifically checks `FormatterState.in_condition` to apply these relaxations only where appropriate (inside `if(...)`, `while(...)`, etc.).
- Verified that all formatter tests pass without regressions.
- Not sure about the approach, let me know if there is a better way.

### DIFF
--- a/crates/formatter/src/internal/format/member_access.rs
+++ b/crates/formatter/src/internal/format/member_access.rs
@@ -314,7 +314,7 @@ impl<'arena> MemberAccessChain<'arena> {
             return true;
         }
 
-        if accesses_len <= 3 {
+        if accesses_len <= 3 && !f.in_condition {
             // we must have at least 2 breaks for chains of length 3 or less
             return false;
         }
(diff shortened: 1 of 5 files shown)


klausi:
Added some more test cases, did not find proper solution yet.

PR #908, opened 2026-01-19
fix(analyzer): add string concatenation combination threshold to stop exponential memory usage

### LINKED ISSUE #909: [analyzer] Exponential memory usage with string concatenation
### Description

When analyzing a file with a large number of string concatenation operations, memory usage is skyrocketing until an OOMKilled error. On my Mac with 48GB RAM, it is using all 48GB (and paging to disk).

I have isolated the issue to [this line](https://github.com/carthage-software/mago/blob/main/crates/analyzer/src/expression/binary/concat.rs#L87) and the recursive call within it (at least in my testing).

My reproduction in the playground causes my RAM to spike up to full usage for around 10 seconds. Adding just one more line of `Test " . $type . "` makes it spend multiple minutes at full throttle and gets OOM killed before it finishes.

I have attempted a fix on #908, but there may be a better long-term solution here since in theory there should only actually be two permutations in my error case (`$type`, is only actually ever one of 2 values throughout the entire string). 

### Playground Link

https://mago.carthage.software/playground#019bd5a9-9646-eb39-a1b8-fab15f36b108

### Reproduction

_No response_

### Environment

- Mago: mago 1.1.0
- OS: macOS 26.1 Tahoe
- Installed via: Shell installer

Noojuno:
## 📌 What Does This PR Do?

Introduced a new configuration setting to limit the maximum number of combinations tracked during string concatenation, preventing exponential blowup in large concatenation chains. The default value is set to 4096.

## 🔍 Context & Motivation

When running mago analyze on our large legacy monolith, we were hitting OOM errors and not able to run the analyzer at all.

## 🛠️ Summary of Changes

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [x] Documentation

## 🔗 Related Issues or PRs

Fixes #909

## 📝 Notes for Reviewers

As mentioned in my issue, there may be a better long-term solution than this, but this seemed like it's probably not a bad idea anyway to avoid memory leaks in future.

### DIFF
--- a/crates/analyzer/src/settings.rs
+++ b/crates/analyzer/src/settings.rs
@@ -6,7 +6,7 @@ use mago_php_version::PHPVersion;
 pub const DEFAULT_FORMULA_SIZE_THRESHOLD: u16 = 512;
 
 /// Default maximum number of combinations to track during string concatenation.
-pub const DEFAULT_STRING_CONCAT_COMBINATION_THRESHOLD: u16 = 4096;
+pub const DEFAULT_STRING_CONCAT_COMBINATION_THRESHOLD: u16 = 512;
 
 /// Configuration settings that control the behavior of the Mago analyzer.
 ///


crates/analyzer/src/settings.rs
```diff
@@ -5,6 +5,9 @@ use mago_php_version::PHPVersion;
 /// Default maximum logical formula size during conditional analysis.
 pub const DEFAULT_FORMULA_SIZE_THRESHOLD: u16 = 512;
 
+/// Default maximum number of combinations to track during string concatenation.
+pub const DEFAULT_STRING_CONCAT_COMBINATION_THRESHOLD: u16 = 4096;
```
azjezz: Actually, we can default to a much smaller number here, i think 1024 or 512 is better.
```suggestion
pub const DEFAULT_STRING_CONCAT_COMBINATION_THRESHOLD: u16 = 512;
```

azjezz:
Thanks @Noojuno !

PR #924, opened 2026-01-23
feat(linter): add no-isset rule to prevent usage of isset construct

djschilling:
## 📌 What Does This PR Do?

Adds a new linter rule `no-isset` that detects and warns against usage of the `isset()` language construct.

## 🔍 Context & Motivation

The `isset()` construct can lead to ambiguous code because it conflates two distinct checks: variable existence and null comparison. It returns `false` for both unset variables and variables set to `null`, making the developer's intent unclear. Using explicit null checks (`!== null`) or the null coalescing operator (`??`) is often clearer and more maintainable.

## 🛠️ Summary of Changes

- **Feature:** Added `clarity/no-isset` rule that warns when `isset()` is used
- **Tests:** Added 8 unit tests (3 success cases, 5 failure cases)

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

- Default severity level is `Warning` (not `Error`)
- Similar in structure to the existing `no-empty` rule
- Good examples suggest using `!== null` or `??`


crates/linter/src/rule/clarity/no_isset.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+impl Default for NoIssetConfig {
+    fn default() -> Self {
+        Self { level: Level::Warning }
+    }
+}
+
+impl Config for NoIssetConfig {
+    fn level(&self) -> Level {
```
Bleksak, line 44: Please implement the default_enabled and return false

djschilling: thanks for the review, i implemented `default_enabled `

azjezz:
Thanks @djschilling

djschilling:
@azjezz thanks for merging. I cannot find the rule in mago version 1.2.2. I thought it should be there, but `mago lint --list-rules  | grep "no-isset"` returns nothing

PR #950, opened 2026-01-26
Implement formatter playground

Zuruuh:
## 📌 What Does This PR Do?

Very simple implementation of a formatter playground. It doesn't support any configuration for now but definitely doable (would require server side changes though)

## 🔍 Context & Motivation

I'm working on integrating the formatter in my codebases and it would be easier to have a playground link to easily isolate, reproduce, and demonstrate the bugs I'd want to report.

## 🛠️ Summary of Changes

- Add a new (sub)tabs system to the playground
- Implement the formatter tab separately from the linter+analyzer one
- Added `lld` to the nix flake, since it's required to build the mago wasm binary

image

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [x] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

The code was mostly generated through AI to be honest since I have very few experience with vue, however I did re-read it, refactor, and tested it locally :+1:

### DIFF
--- a/docs/.vitepress/theme/components/playground/MagoPlayground.vue
+++ b/docs/.vitepress/theme/components/playground/MagoPlayground.vue
@@ -3,6 +3,7 @@ import { onMounted, onUnmounted, watch, ref } from 'vue';
 import { useMagoWasm } from '../../composables/useMagoWasm.js';
 import { createPlaygroundState } from '../../composables/usePlaygroundState.js';
 import { useUrlState } from '../../composables/useUrlState.js';
+import { useMagoTasks } from '../../composables/useMagoTasks.js';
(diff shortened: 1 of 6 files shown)


azjezz:
Hi! sorry for late reply.

I'm not really understanding what this does 🤔 

the current playground already has `Format` button to do this, am i missing something?

Zuruuh:
Hello, the point isn't really to *just* format the code but to make it easier to compare the code before and after it has been formatted. I think it makes it easier to find and report bugs for the formatter, and I took inspiration from the Biome toolchain, which already has it's own playground with a formatted output tab https://biomejs.dev/playground/

PR #959, opened 2026-01-28
refactor(codex): parallelize codebase population using level-based deps graph

azjezz:


azjezz:
Hm, this shows no actual benifits in release+pgo builds, it does however help a lot in debug builds - analyzing a large project that usually takes 5 seconds with release+pgo builds went from 41 seconds to 20 seconds using debug build.

The overhead of thread coordination, memory sync, and probably removal of branch prediction via PGO negates the gains... its probably worth exploring this again later.

One cool thing in this pr is the the dependency graph, which we might want to use for other things in the future.

PR #1008, opened 2026-02-02
chore: update .gitattributes to exclude more files from export

shyim:
## 📌 What Does This PR Do?

Ignore more files for composer based installs.

image

I wondered why it takes so long to install 😂


azjezz:
root `.md` files and `composer.lock` should be kept.

shyim:
done :)

PR #1015, opened 2026-02-03
feat(analyzer): add PSR-11 container support via plugin

Noojuno:
## 📌 What Does This PR Do?

Adds PSR-11 container plugin that infers types from class-string arguments.

## 🔍 Context & Motivation

Our organization makes heavy use of the PSR-11 container composer package for dependency injection.

PSR standard: https://www.php-fig.org/psr/psr-11/
Composer package: https://packagist.org/packages/psr/container

As it stands, our code that uses this container does not infer the type from the class-string, which causes `mixed-method-access` and `mixed-argument` errors.

Playground: https://mago.carthage.software/playground#019c2137-03e0-23ac-7921-f8edaa7e77ae

## 🛠️ Summary of Changes

- **Feature:** Added PSR-11 container plugin to correctly infer from class strings for dependency injection.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [x] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [x] Documentation
- [x] Other (please specify): Analyzer, Playground

### DIFF
--- a/crates/analyzer/src/plugin/libraries/mod.rs
+++ b/crates/analyzer/src/plugin/libraries/mod.rs
@@ -1,14 +1,16 @@
-//! Library-specific providers for PHP stdlib, PSL, and Flow-PHP.
+//! Library-specific providers for PHP stdlib, PSL, Flow-PHP, and PSR-11 Container.
 
 pub mod flow_php;
 pub mod psl;
+pub mod psr_container;
 pub mod stdlib;
 
 use crate::plugin::Plugin;
 
 pub use flow_php::FlowPhpPlugin;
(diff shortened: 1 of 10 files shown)


azjezz:
Thank you @Noojuno !

Can you please add documentation for this plugin in `docs/` as well?

KorvinSzanto:
Can this be called something more specific than `psr_container`? The container PSR is not only for service locators, and isn't guaranteed to return an instance of `T` when given `class-string<T>`. Maybe something like `psr_container_class_strings` or `psr_container_service_locator`?

Issue #1020, opened 2026-02-03
False positive for `prefer-first-class-callable`


dragosprotung:
### Description

First-class callable syntax can not be used in all scenarios.
`adminUrlGenerator` method might depend on runtime state.

### Playground Link

https://mago.carthage.software/playground#019c2499-1c99-98da-548c-698b242cc05d

### Environment

Mago 1.4

PR #1027, opened 2026-02-04
fix(linter): skip runtime-dependent call targets in prefer-first-class-callable

kzmshx:
## 📌 What Does This PR Do?

Fixes false positives in `prefer-first-class-callable` where the rule suggested converting closures with runtime-dependent call targets to first-class callable syntax, changing evaluation semantics.

## 🔍 Context & Motivation

The rule flagged patterns like:

```php
linkToUrl(fn(): string => adminUrlGenerator()->generateUrl());
```

and suggested `adminUrlGenerator()->generateUrl(...)`. However, this changes when `adminUrlGenerator()` is evaluated — from every invocation to closure creation time only. Similarly, null-safe method calls (`$obj?->method()`) and dynamic method names (`$obj->$method()`) have no equivalent first-class callable syntax.

## 🛠️ Summary of Changes

- **Bug Fix:** Added `is_convertible_to_first_class_callable()` helper that verifies the call target is safe for conversion (simple identifiers, variables, or static class references with non-dynamic method selectors).
- **Bug Fix:** Updated `prefer-arrow-function` skip logic to account for non-convertible call targets, so arrow function suggestions are not suppressed when first-class callable conversion is unsafe.
- **Tests:** Added 5 success and 3 failure test cases covering method chains, null-safe calls, dynamic method names, and closure variants.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes carthage-software/mago#1020

### DIFF
--- a/crates/linter/src/rule/best_practices/prefer_first_class_callable.rs
+++ b/crates/linter/src/rule/best_practices/prefer_first_class_callable.rs
@@ -219,20 +219,20 @@ pub(super) fn is_call_forwarding<'ast, 'arena>(
 }
 
 pub(super) fn is_convertible_to_first_class_callable<'ast, 'arena>(call: &'ast Call<'arena>) -> bool {
-    match call {
-        Call::Function(FunctionCall { function: Expression::Identifier(_) | Expression::Variable(_), .. }) => true,
-        Call::Method(MethodCall {
-            object: Expression::Variable(_),
-            method: ClassLikeMemberSelector::Identifier(_),
-            ..
-        }) => true,
-        Call::StaticMethod(StaticMethodCall {
-            class: Expression::Identifier(_) | Expression::Self_(_) | Expression::Static(_) | Expression::Parent(_),
-            method: ClassLikeMemberSelector::Identifier(_),
-            ..
-        }) => true,
-        _ => false,
-    }
+    matches!(
+        call,
+        Call::Function(FunctionCall { function: Expression::Identifier(_) | Expression::Variable(_), .. })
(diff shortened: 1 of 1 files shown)


crates/linter/src/rule/best_practices/prefer_first_class_callable.rs
```diff
@@ -205,3 +217,120 @@ pub(super) fn is_call_forwarding<'ast, 'arena>(
...
+                method: ClassLikeMemberSelector::Identifier(_),
+                ..
+            })
+            | Call::StaticMethod(StaticMethodCall {
+                class: Expression::Identifier(_) | Expression::Self_(_) | Expression::Static(_) | Expression::Parent(_),
+                method: ClassLikeMemberSelector::Identifier(_),
+                ..
+            })
+    )
+}
```
azjezz, line 236: is there a reason for `{identifier or variable}()`, `{variable}->{identifier}()` and `{identifier}::{identifier}()` only? we not stop for all calls?

azjezz: nvm, this is correct, can be more flexiable, but not worth it.

kzmshx: Honestly I missed those cases, but yeah, not worth the added complexity.

PR #1052, opened 2026-02-06
Laravel Pint formatter preset parity

nikspyratos:
## 📌 What Does This PR Do?

Bring the Pint formatter preset in line with Pint's actual defaults.

## 🔍 Context & Motivation

I had a lot of conflicts between the Pint preset and Pint itself, and a little bit even more so with my personal Pint config. So, I started from scratch with an empty config for both Mago's formatter and Pint, and set both to the pint/laravel preset and compared their changes for differences.

Turns out by default there are few currently - one or two rules that need changing, and one rule on Pint's side that needs disabling as AFAICT there is no Mago equivalent.

## 🛠️ Summary of Changes

- `space_after_logical_not_unary_prefix_operator: true`: Pint's Laravel default has `not_operator_with_successor_space: true`, so just solving the conflict.
- `space_within_grouping_parenthesis: false`: Pint's Laravel default has `spaces_inside_parentheses: true`, but I believe this seems to be mapping to PHPCS's `none` [config](https://mlocati.github.io/php-cs-fixer-configurator/#version:3.93|fixer:spaces_inside_parentheses) for that rule, as if I have this enabled in both tools then they loop in making opposite changes to it continuously.
- `inline_empty_*_braces: true`: Having all of these as true is equivalent to the default configured `single_line_empty_body` as true.

## 📂 Affected Areas

- [ ] Linter
- [X] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

Laravel 12, updated recently, Mago v1.5.0 via Composer.

[Current Pint Laravel preset for reference](https://github.com/laravel/pint/blob/main/resources/presets/laravel.php)

My testing of this was quite rudimentary: Just continuously running various configurations of Mago and Pint on a personal project (small-medium ish Laravel repo) and considering it a success if they don't change any files from each other anymore. 
Two things based on that testing:
- One thing I picked up from that testing was [this](https://github.com/carthage-software/mago/discussions/1051) around Pint's default `no_unneeded_control_parentheses` behaviour vs Mago adding parentheses around a return statement I have.
- As you can see in the recording, the line breaking from Mago on a class with many interfaces seems to make Pint's class_definition or similar unhappy. Although, I'm not quite sure why - but this would not go away unless I made the print-width setting wider. I'm unsure if that's an issue on Mago's side or Pint's, actually (probably the latter?)

https://github.com/user-attachments/assets/17d17555-be48-4f9f-a224-e8b3701580db

---

Mago is awesome, would love to see it gain traction in PHP space in 2026!


azjezz:
Thank you!

PR #1062, opened 2026-02-07
perf(database): avoid redundant canonicalize() calls in loader

aszenz:
**⚠️ Disclaimer: This code was generated by Claude Code. I verified the output/benchmarks but did not review the implementation details.**

---

Only call `canonicalize()` on symlinks during file discovery. Non-symlink paths under a canonical workspace are already canonical.

**Before:** ~833k `readlink` syscalls on a 30k file project  
**After:** ~30 `readlink` syscalls

~12% reduction in system time in benchmarks.

### DIFF
--- a/crates/database/src/loader.rs
+++ b/crates/database/src/loader.rs
@@ -166,7 +166,11 @@ impl<'a> DatabaseLoader<'a> {
         glob_excludes: &GlobSet,
         path_excludes: &HashSet<&Cow<'a, Path>>,
     ) -> Result<Vec<FileWithSpecificity>, DatabaseError> {
-        let mut paths_to_process: Vec<(std::path::PathBuf, String, usize)> = Vec::new();
+        // Tuple: (path, pattern, specificity, is_symlink)
+        // We track symlink status to avoid unnecessary canonicalize() calls.
+        // Since the workspace is already canonicalized, non-symlink children
+        // are already canonical and don't need resolution.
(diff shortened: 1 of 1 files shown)


crates/database/src/loader.rs
```diff
@@ -211,23 +221,39 @@ impl<'a> DatabaseLoader<'a> {
 
                 for entry in WalkDir::new(&dir_path).into_iter().filter_map(Result::ok) {
                     if entry.file_type().is_file() {
-                        paths_to_process.push((entry.into_path(), root.to_string(), specificity));
+                        // WalkDir follows symlinks by default, but we can check if the
```
dotdash, line 224: This seems exactly backwards to me. WalkDir "follows" the symlink by having the OS follow it. The resulting path will still have the original dir_path in it, which might contain a symlink. In fact if "foo" is a symlink, but you specified "foo/" as dir_path, then this won't be reported as a symlink either.

aszenz: Yeah it doesn't seem to be correct. Claude suggests another idea to cache these calls which seems to bring similar improvements but since @azjezz wants to refactor the whole database crate anyway for lsp, won't make sense to put too much effort on this micro-optimization.

aszenz:
> AFAICT, this changes observable behaviour because it ignores possible intermediate symlinks

Thanks for reviewing, I'm going to close this in light of above comment.

Issue #831, opened 2025-12-25
bug: psalm-assert-if-true results in redundant-type-comparison


rauanmayemir:
### Description

I have uses of

```
Type\property_name_of($input::class)->assert($field);
```

but mago reports it as
```
warning[redundant-type-comparison]: Redundant type assertion: `$field` of type `non-empty-string` is always not `non-empty-string`.
```

this is because `PropertyNameOfType` has a `@psalm-assert non-empty-string $value`, but it's not the only thing the function asserts.

### Environment

mago v1.0.3

PR #1079, opened 2026-02-09
fix(analyzer): suppress redundant-type-comparison for unconditional psalm-assert

kzmshx:
## 📌 What Does This PR Do?

Suppresses false positive `redundant-type-comparison` warnings for unconditional `@psalm-assert` annotations.

## 🔍 Context & Motivation

`@psalm-assert` describes the type-narrowing side effect of a function, not its full behavior. The annotated function may perform runtime validation beyond what the type system captures:

```php
/** @psalm-assert non-empty-string $value */
function assertNonEmptyString(mixed $value): void {
    if (!is_string($value) || $value === '') {
        throw new InvalidArgumentException('...');
    }
}

$field = getField(); // non-empty-string
assertNonEmptyString($field); // false positive: "Redundant type assertion"
```

The current suggestion "Consider removing this assertion" would remove the runtime safety guard. Conditional assertions (`@psalm-assert-if-true`/`if-false`) are unaffected — their redundancy warnings remain useful for detecting dead branches.

## 🛠️ Summary of Changes

- **Bug Fix:** Added `report_redundancy` parameter to `resolve_invocation_assertion()`, set to `false` for unconditional assertions and `true` for conditional ones.
- **Tests:** Added `issue_831.php` test case and updated `redundant_assert.php` to reflect the new behavior.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation

## 🔗 Related Issues or PRs

Fixes carthage-software/mago#831

## 📝 Notes for Reviewers

`ImpossibleTypeComparison` is intentionally preserved for unconditional assertions — calling `@psalm-assert Bar $value` with a `Foo` argument is a genuine bug.

### DIFF
--- a/crates/analyzer/src/invocation/post_process.rs
+++ b/crates/analyzer/src/invocation/post_process.rs
@@ -175,6 +175,7 @@ pub fn post_invocation_process<'ctx, 'arena>(
         &metadata.if_true_assertions,
         template_result,
         parameters,
+        true,
     );
 
     for (variable, assertions) in resolved_if_true_assertions {
@@ -190,6 +191,7 @@ pub fn post_invocation_process<'ctx, 'arena>(
         &metadata.if_false_assertions,
         template_result,
         parameters,
+        true,
     );
 
     for (variable, assertions) in resolved_if_false_assertions {
@@ -241,6 +243,7 @@ fn apply_assertion_to_call_context<'ctx, 'arena>(
         assertions,
         template_result,
         parameters,
+        false,
     );
 
     if type_assertions.is_empty() {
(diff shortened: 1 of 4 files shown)


azjezz:
#831 is unrelated.

---

This is wrong, we want to keep this check.

The idea is if something is verified statically to be correct, there should not be a runtime check for it, as it does not add any value, and only makes your code slower.

in the test example you shared, `get_field` is known to return `non-empty-string`, so there is no reason to verify that. if `get_field` would to return a potentionally empty string, then mago will emit an error there.

If the source of a value is not trusted ( e.g, from vendor, where it claims `non-empty-string` return, but might return empty sometimes ), you should widen the type first, e.g:

```
/** @var string $foo */
$foo = get_field();

// assert here, it won't be considered redundant.
```

PR #1101, opened 2026-02-11
chore(prelude): improve `bin2hex` stubs

veewee:
## 📌 What Does This PR Do?

Improves bin2hex stubs (in a similar way as psalm)

## 🔍 Context & Motivation

It is not strict enought

## 🛠️ Summary of Changes

```php
bin2hex('') -> literal('')
bin2hex('a') -> non-empty-string
```

## 📂 Affected Areas

- [ ] Analyzer
- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

### DIFF
--- a/crates/prelude/assets/extensions/standard.php
+++ b/crates/prelude/assets/extensions/standard.php
@@ -1017,6 +1017,8 @@ function constant(string $name): mixed {}
 
 /**
  * @pure
+ * 
+ * @return ($string is non-empty-string ? non-empty-string : '')
  */
 function bin2hex(string $string): string {}
 
--- a/crates/analyzer/tests/cases/bin2hex.php
+++ b/crates/analyzer/tests/cases/bin2hex.php
@@ -1,0 +1,17 @@
+<?php
+
+/**
+ * @return non-empty-string
+ */
+function test_non_empty_bin2hex(): string
+{
+    return bin2hex('a');
+}
+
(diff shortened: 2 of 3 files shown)


crates/prelude/assets/extensions/standard.php
```diff
@@ -1017,6 +1017,8 @@ function constant(string $name): mixed {}
 
 /**
  * @pure
+ * 
+ * @return ($string is non-empty-string ? non-empty-string : '')
```
tscni, line 1021: What if analysis can't determine that it's a non-empty-string?
```php
function foo(): string { // explicitly not non-empty-string
    return 'bar';
}

bin2hex(foo()); // --> '626172', not ''
```

veewee: Returning string instead of '' might be a better idea for that case.

azjezz: It already returns `non-empty-string|''` in this case, which gets normalized to `string`/

tscni: Ah I see, I misunderstood the behavior of "$x is Y" as "if $x is definitely Y"
Thanks a lot for the clarification :)

PR #1121, opened 2026-02-11
Redundant grouping parens test

Michael4d45:
## 📌 What Does This PR Do?

Fixes redundant grouping parentheses in member access chains. Expressions like `((($items ?? Fallback::X)))->value` are formatted as `($items ?? Fallback::X)->value` instead of `(($items ?? Fallback::X))->value`.

## 🔍 Context & Motivation

When formatting member access chains (`->property` or `?->property`) where the base expression is Binary (e.g. `??`) or Conditional, parentheses could be added in two places: by the base’s `wrap!` and by the member access logic. That produced double parentheses `((expr))->prop` instead of a single set `(expr)->prop`.

## 🛠️ Summary of Changes

- **Bug Fix:** Fixed redundant grouping parentheses in member access chains when the base is Binary or Conditional.
  - Use unwrapped base in `collect_member_access_chain` so we don’t treat outer grouping as essential.
  - Enter PropertyAccess/NullSafePropertyAccess before formatting the base in `print_member_access_chain` so only one place adds parens.
  - Add early `need_parens` check so Binary/Conditional add parens when they are the object of `->` or `?->`.
  - Add grandparent check in `binary_node_needs_parens` for `PropertyAccess` / `NullSafePropertyAccess` when the immediate parent is Expression.
  - In `called_or_accessed_node_needs_parens`, avoid adding parens for the wrapping Expression when the inner Binary/Conditional already adds them.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

- Changes are in `member_access.rs` and `parens.rs`.
- Added test case: `redundant_grouping_parens`.

### DIFF
--- a/crates/codex/src/metadata/mod.rs
+++ b/crates/codex/src/metadata/mod.rs
@@ -766,7 +766,7 @@ impl CodebaseMetadata {
 
         if let Some(overridden_map) = class_like.overridden_method_ids.get(method_name) {
             for (parent_class_name, parent_method_id) in overridden_map {
-                if class_like.name.eq_ignore_ascii_case(&parent_class_name) {
+                if class_like.name.eq_ignore_ascii_case(parent_class_name) {
                     continue; // Skip self-recursion if the method overrides itself
                 }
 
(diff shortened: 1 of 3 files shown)


azjezz:
Hi @Michael4d45 can you please rebase and remove @cursoragent from git commits? I don't care about using AI, but i don't want tools present in git history.

PR #1138, opened 2026-02-14
fix(analyzer): fix type for variadic parameters from list<T> to array<T>

dotdash:
For example when using named arguments, the array for a variadic function parameter may be of type array<T> rather than list<T>. Only when named arguments are not allowed can we assume the type to be list<T>.

## 📌 What Does This PR Do?

Fix the type for variadic parameters from `list<T>` to `array<T>`.

## 🔍 Context & Motivation

When using named arguments, the resulting array will be associative, not a list, so the current behaviour is wrong.

## 🛠️ Summary of Changes

- **Bug Fix:** Fixed the type for variadic parameters from `list<T>` to `array<T>`.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation

## 🔗 Related Issues or PRs

Discord discussion

## 📝 Notes for Reviewers

herp derp

### DIFF
--- a/crates/analyzer/src/statement/function_like/mod.rs
+++ b/crates/analyzer/src/statement/function_like/mod.rs
@@ -15,8 +15,6 @@ use mago_codex::misc::GenericParent;
 use mago_codex::ttype::TType;
 use mago_codex::ttype::TypeRef;
 use mago_codex::ttype::atomic::TAtomic;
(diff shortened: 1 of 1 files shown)


azjezz:
Tests? 😅

Issue #1139, opened 2026-02-14
Support recursive types


ddanielou:
### Feature

Correctly describing arbitrarily nested arrays is impossible at the moment. I ran into this issue while trying to fix `@param-out` in the definition of `parse_str()`. This is also an issue when trying to describe something as common as a JSON object.

Relevant PHPstan issue: https://github.com/phpstan/phpstan/issues/11684

### Example

```php
<?php

// Something like this maybe:
/** @type QueryString array<string, string|QueryString> */
```

azjezz:
I'm not sure we need recursive types, what we probably want is `tree<K, V>`, an array variant that when accessed with `K` would return `V|tree<K, V>` and you can assign to it `V`, `tree<K, V>` or `array<K, V>`.

I would just be a new variant to `TArray`.

PR #1140, opened 2026-02-14
fix(prelude): fix parse_str() @param-out

ddanielou:
## 📌 What Does This PR Do?
Correct parse_str() second parameter type.

## 🔍 Context & Motivation

Query strings can encode arrays, e.g. `arr[k1]=val1&arr[k2]=val2` which is decoded as `["k1" => "val1", "k2" => "val2"]`. Currently, the annotation of `parse_str()` only describes query string of the form `foo=bar&baz=qux`.

## 🛠️ Summary of Changes

- **Bug Fix:** fixed the `@param-out` annotation in `parse_str()`. 

## 📂 Affected Areas

- [x] Analyzer
- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

https://github.com/carthage-software/mago/issues/1139

## 📝 Notes for Reviewers

The new signature isn't perfect. It only supports a single level of nesting, while query strings can represent multidimensional arrays.

### DIFF
--- a/crates/analyzer/tests/mod.rs
+++ b/crates/analyzer/tests/mod.rs
@@ -402,6 +402,7 @@ test_case!(spread_list);
 test_case!(or_never);
 test_case!(abstract_method_inheritance);
 test_case!(variadic_array_type);
+test_case!(parse_str);
 
 // Github Issues
 test_case!(issue_659);


crates/prelude/assets/extensions/standard.php
```diff
@@ -1672,7 +1672,7 @@ function chr(int $codepoint): string {}
 function ord(string $character): int {}
 
 /**
- * @param-out array<string, string> $result
+ * @param-out array<string, string|array<array-key, string>> $result
```
azjezz:
```suggestion
 * @param-out array<string, string|array<array-key, string|array<array-key, mixed>>> $result
```

Actually, this can be nested many times..

ddanielou: Indeed, it can be nested infinitely many times… but how should this be represented? See https://github.com/carthage-software/mago/issues/1139
Also, is `mixed` correct here? Query strings can only represent strings or array of strings.

azjezz: Yes, we should define 3 levels deep as correct, then fallback to mixed. Most the time the users won't be dealing with this type directly beyond 3 levels deep.

Issue #884, opened 2026-01-14
[Formatter] Configuration "space-after-colon-in-enum-backing-type" does not exist


yoannblot:
### Description

The official documentation is talking about a configuration named `space-after-colon-in-enum-backing-type` which should "Add a space after the colon in enum backing types."
But this does not exist.

We have this error running the formatter : 
> ERROR Failed to build the configuration: unknown field `space-after-colon-in-enum-backing-type` for key `formatter`

I look into the source code, and there is no reference except this doc : 
https://github.com/carthage-software/mago/blob/fe51025116e40c613d0feb2c3bc47e3b39f8eadd/docs/tools/formatter/configuration-reference.md?plain=1#L126

### Environment

everywhere

PR #1151, opened 2026-02-17
fix(formatter): remove `space_after_colon_in_enum_backing_type`

ptondereau:
## 📌 What Does This PR Do?

We're removing from the doc `space_after_colon_in_enum_backing_type`

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

#884


ptondereau:
Actually, I don't know if you really want this because the formatter doesn't have `space-after-colon-in-return-type` either. So there's no reason enum backing types should be different.
Please let me know, and I can amend my PR to just remove it from the doc

azjezz:
Actually, we had this rule before, we removed it at some point when we were removing rarely used settings to reduce their number, but apparently forgot to update docs.. lets just remove the outdated docs instead

ptondereau:
> Actually, we had this rule before, we removed it at some point when we were removing rarely used settings to reduce their number, but apparently forgot to update docs.. lets just remove the outdated docs instead

done thanks

PR #1154, opened 2026-02-17
fix(analyzer): fix detection of possibly undefined keys in unions of sealed/unsealed arrays

dotdash:
## 📌 What Does This PR Do?

Fixes a bug where an unsealed array is not considered to be able to contain an unknown key, because the unsealed-ness is only considered when there are no known keys at all.

## 🔍 Context & Motivation

Fixes possibly undefined keys being reported as definitely undefined.

## 🛠️ Summary of Changes

- **Bug Fix:** Fixed detection of possibly undefined keys in unions of sealed/unsealed arrays

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [x] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Nyet

## 📝 Notes for Reviewers

It's a me, dotdashio

### DIFF
--- a/crates/analyzer/tests/cases/undefined_string_key_in_union.php
+++ b/crates/analyzer/tests/cases/undefined_string_key_in_union.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+/** @var array{foo: int, ...}|array{foo: int} $x */
+
+// @mago-expect analysis:undefined-string-array-index,undefined-string-array-index
+var_dump($x['bar']);
+
+// This used to produce a bogus impossible-nonnull-entry-check
+isset($x['bar']);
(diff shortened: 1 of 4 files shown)


crates/analyzer/tests/cases/undefined_string_key_in_union.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+/** @var array{foo: int, ...}|array{foo: int} $x */
+
+// @mago-expect analysis:undefined-string-array-index,undefined-string-array-index
+var_dump($x['bar']);
```
azjezz, line 8: why is it emitted twice? 🤔

PR #1159, opened 2026-02-17
feature(composer): download binary from wrapper script, just in time

edsrzf:
## 📌 What Does This PR Do?

This change removes the Composer plugin, and instead adds the platform detection and binary installation logic to the `vendor/bin/mago` wrapper script.

## 🔍 Context & Motivation

The benefits:
- Simpler binary installation process with no need to allow a Composer plugin.
- When the same working directory is accessed by multiple platforms (eg running Docker on MacOS with a volume mounted in), binaries for all platforms may coexist and be used when appropriate.

## 🛠️ Summary of Changes

- **Feature:** Removes the Composer plugin and moves platform-detection and downloading in the `vendor/bin/mago` wrapper script.
- **Dependencies:** Removes the `composer-plugin-api` dependency. Add `ext-curl` as a suggested dependency.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [x] Composer Plugin
- [x] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

None, but this was discussed on Discord in December 2025.

## 📝 Notes for Reviewers

I've kept the script self-contained (aside from the Symfony dependency), but am open to splitting out the logic into one or more classes if that's the preference.

Claude Code was used to author this PR, but I have thoroughly reviewed the changes and tested it myself.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,7 +18,6 @@
 - [ ] Linter
 - [ ] Formatter
 - [ ] CLI
-- [ ] Composer Plugin
 - [ ] Dependencies
 - [ ] Documentation
 - [ ] Other (please specify):


composer/bin/mago
```diff
@@ -3,22 +3,203 @@
...
+                    $totalMb = number_format($dlSize / 1048576, 1);
+                    fprintf(STDERR, "\r  %s / %s MB (%d%%)", $dlMb, $totalMb, $pct);
+                }
+                return 0;
+            });
+
+            $success = curl_exec($ch);
+            $statusCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+            $error = curl_error($ch);
+            curl_close($ch);
```
karoun, line 156: This function is deprecated in PHP 8.5 and causes failures when such warnings are turned into exceptions.

azjezz: Should be solved in the latest release

Issue #1147, opened 2026-02-15
`prefer-first-class-callable` false positive when callee accepts fewer parameters than caller provides


hirasso:
First of all: Wow, Mago is a revelation! A big thank you for everyone working on this.

### Bug Description

The [`prefer-first-class-callable` linter rule](https://mago.carthage.software/tools/linter/rules/best-practices#prefer-first-class-callable) suggests replacing arrow functions with first-class callable syntax. It even does that when the called function accepts fewer parameters than the caller will provide, leading to `ArgumentCountError` at runtime.                                                                                                           
                           
### Example

```php
declare(strict_types=1);

/**
 * Check if any element in the array matches the callback.
 *
 * @template TKey of array-key
 * @template TValue
 *
 * @param  array<TKey, TValue>  $array
 * @param  (callable(TValue, TKey): bool)  $callback
 */
function some(array $array, callable $callback): bool
{
    foreach ($array as $key => $value) {
        if ($callback($value, $key)) {
            return true;
        }
    }

    return false;
}

// flagged but correct:
some(['foo', 'bar', 2], static fn($value) => \is_int($value));
// recommended but will fail because is_int is a native function:
// some(['foo', 'bar', 2], \is_int(...));
```

### Why it fails

PHP _internal functions_ enforce strict argument counts. `\is_int(...)` expects exactly 1 argument, but the `some()` callback is called with 2 (`$value`, `$key`). This throws:

```
Fatal error: Uncaught ArgumentCountError: is_int() expects exactly 1 argument, 2 given
```

### Expected behavior

The rule should not suggest first-class callable syntax when the target function/method accepts fewer parameters than the calling context may provide. Ideally, it would analyze the callee's signature and compare it against the number of arguments the arrow function receives (or that the caller is known to pass).

### Playground Link / Reproduction

https://mago.carthage.software/playground#019c62fd-d9d4-be18-9647-1098ebcd42bb

### Environment

- Mago: 1.8.0
- macOS
- Installed via composer

PR #1160, opened 2026-02-18
fix(linter): add check-functions option to prefer-first-class-callable

kzmshx:
## 📌 What Does This PR Do?

Adds a `check-functions` config option (default `false`) to the `prefer-first-class-callable` rule. By default, the rule now only suggests FCC conversion for method and static method calls. Function calls can be opted in via `check-functions: true`.

## 🔍 Context & Motivation

Internal PHP functions throw `ArgumentCountError` on extra arguments, while user-defined functions silently ignore them. When a closure acts as an arity adapter (e.g., `fn($v) => is_int($v)` in a callback that receives extra arguments), converting to `is_int(...)` can cause runtime errors. The linter cannot statically distinguish internal from user-defined functions.

## 🛠️ Summary of Changes

- Added `check_functions: bool` (default `false`) to `PreferFirstClassCallableConfig`
- Added guard clause in `check()` for both arrow functions and closures to skip `Call::Function` when `check_functions` is disabled
- Updated rule description to document the option and false positive risk
- Changed examples from function call (`strtoupper`) to method call (`$formatter->format`)

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes carthage-software/mago#1147

### DIFF
--- a/crates/linter/src/rule/best_practices/prefer_first_class_callable.rs
+++ b/crates/linter/src/rule/best_practices/prefer_first_class_callable.rs
@@ -44,11 +44,12 @@ pub struct PreferFirstClassCallableRule {
 #[serde(default, rename_all = "kebab-case", deny_unknown_fields)]
 pub struct PreferFirstClassCallableConfig {
     pub level: Level,
+    pub check_functions: bool,
 }
 
 impl Default for PreferFirstClassCallableConfig {
     fn default() -> Self {
-        Self { level: Level::Warning }
+        Self { level: Level::Warning, check_functions: false }
     }
 }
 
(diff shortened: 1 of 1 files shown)


azjezz:
The problem is that this takes away a feature, but i guess gives more safety, wdyt about keeping the old behavior behind an option ( disabled by default ), and informing people in the rule description that it can be enabled, but might cause false positives? 


Also, if the PHP version is `8.6` or above, we can make the suggestion `foo(?)` which completely solves the problem. ( https://wiki.php.net/rfc/partial_function_application )

kzmshx:
@azjezz Thanks for the feedback!

https://github.com/carthage-software/mago/pull/1160/commits/cb030fff9ac9914c13e4f2b53e8e1022fe814071

I've made the rule opt-in, with the `ArgumentCountError` risk noted in the
description.

> if the PHP version is 8.6 or above, we can make the suggestion foo(?) which completely solves the problem.

Since `prefer-first-class-callable` specifically refers to `foo(...)` syntax, would it make sense to handle `foo(?)` in a separate `prefer-partial-application` rule (PHP 8.6+)? It could safely cover all call types and suppress `prefer-first-class-callable` when enabled.

azjezz:
> Since prefer-first-class-callable specifically refers to foo(...) syntax

Actually, starting with PHP 8.6, both `foo(?)` and `foo(...)` ( as well as `foo(?, 1, ...)` ) are considered partial applications, we already removed the naming "First Class Callable" from the ast, now both share same AST. so i don't see why we should split them in the linter, we will just rename this rule down the line probably.

crates/linter/src/rule/best_practices/prefer_first_class_callable.rs
```diff
@@ -56,6 +56,10 @@ impl Config for PreferFirstClassCallableConfig {
     fn level(&self) -> Level {
         self.level
     }
+
+    fn default_enabled() -> bool {
+        false
+    }
```
azjezz: Actually, this is not what i meant, this should stay as is.

My suggestion was to introduce an option, that says if we should trigger for `foo(...)` or not, if disabled, we trigger for methods / static methods, but not functions.

kzmshx: Got it. Replaced `default_enabled() -> false` with a `check-functions` option (default `false`) — methods/static methods only unless enabled. Also noted the FP risk in the description.

kzmshx:
> Actually, starting with PHP 8.6, both foo(?) and foo(...) ( as well as foo(?, 1, ...) ) are considered partial applications, we already removed the naming "First Class Callable" from the ast, now both share same AST

Makes sense, thanks for the context. I won't split it into a separate rule then.

PR #1177, opened 2026-02-20
fix(analyzer): clear safe_symbols before calling incremental analyzer to stop existing errors from being cleared on file update in watch mode

### LINKED ISSUE #1176: Analyzer: Saving a body-only change with watch mode clears existing errors in file
### Description

When using `mago watch --analyze`, making a "non-body" update on a file with existing errors will cause the existing errors to not be listed in the updated watcher output.

From what I can tell this is because in "non-body" updates, the `safe_symbols` is set to all `class_likes` from the `merged_codebase` here https://github.com/carthage-software/mago/blob/main/crates/orchestrator/src/service/incremental_analysis.rs#L502

I noticed that clearing the safe_symbols off of the `CodebaseMetadata` before passing it to `self.run_analyzer_selective`, but I'm not sure if that's the best solution

### Playground Link

_No response_

### Reproduction

1. Start `mago watch --analyze`
2. Observe errors in the output in `Child.php`
3. Add a non-existent method call `process` in `Child.php` and save
4. Observe the watcher output now says zero errors

Here's an example output using those steps.

Initial file

```php
<?php

class Child extends Base
{
    // BUG: incompatible return type (float instead of string).
    // The analyzer should report this on every watch cycle.
    public function process(int $id): float
    {
        return 1.0;
    }

    private function unused(): void {}
}

```

file update

```php
<?php

class Child extends Base
{
    // BUG: incompatible return type (float instead of string).
    // The analyzer should report this on every watch cycle.
    public function process(int $id): float
    {
        $this->something(); // Add a call to a function that doesn't exist.
        return 1.0;
    }

    private function unused(): void {}
}

```

```
2026-02-20T04:27:34.917718Z  INFO main mago::commands::analyze: Starting watch mode. Press Ctrl+C to stop.
2026-02-20T04:27:34.920327Z  INFO main mago_database::watcher: Database watcher started for workspace: /Users/chrisopperwall/projects/mago/bug1-safe-symbols-leak
2026-02-20T04:27:34.920350Z  INFO main mago::commands::analyze: Watching /Users/chrisopperwall/projects/mago/bug1-safe-symbols-leak for changes...
2026-02-20T04:27:34.920366Z  INFO main mago::commands::analyze: Running initial analysis...
error[incompatible-return-type]: Return type `float` of `Child::process()` is incompatible with parent return type `string` of `Base::process()`
  ┌─ Child.php:7:21
  │
3 │ class Child extends Base
  │       ----- In class `Child`
  ·
7 │     public function process(int $id): float
  │                     ^^^^^^^ Returns type `float` but parent expects `string`
  │
  ┌─ Base.php:3:16
  │
3 │ abstract class Base
  │                ---- Parent method `Base::process()` return type defined here
  │
  = Return types must be covariant: child must return equal or narrower types than parent.
  = Help: Change the return type to be compatible with the parent method.

help[unused-method]: Method `unused()` is never used.
   ┌─ Child.php:14:22
   │
 3 │ ╭ class Child extends Base
 4 │ │ {
 5 │ │     // BUG: incompatible return type (float instead of string).
 6 │ │     // The analyzer should report this on every watch cycle.
   · │
14 │ │     private function unused(): void {}
   │ │                      ^^^^^^ Method `unused()` is declared here.
15 │ │ }
   │ ╰─'
   │
   = This method is declared but never called within the class.
   = Help: Consider prefixing the method with an underscore (`_`) to indicate that it is intentionally unused, or remove it if it is not needed.

error: found 2 issues: 1 error(s), 1 help message(s)
 = 1 issues contain auto-fix suggestions

2026-02-20T04:27:34.981689Z  INFO main mago::commands::analyze: Initial analysis complete. Watching for changes...
2026-02-20T04:27:46.184161Z  INFO main mago::commands::analyze: Detected 1 file change(s), re-analyzing...
2026-02-20T04:27:46.192597Z  INFO main mago::service: No issues found.
2026-02-20T04:27:46.192657Z  INFO main mago::commands::analyze: Analysis complete. Watching for changes...

```

### Environment

- Mago: 1.9.1
-  OS: Linux
-  Built from source

chrisopperwall-qz:
## 📌 What Does This PR Do?

This is an attempt to fix the issue in https://github.com/carthage-software/mago/issues/1176

## 🔍 Context & Motivation

I think by having the class_like symbols in safe_symbols of the updated files might be stopping the classes from being analyzed. If this isn't the right solution I'm happy to close this PR but I thought I'd give it a shot, since this seemed to solve the issue for me.

This just affects "body_only" edits.

## 🛠️ Summary of Changes
- **Bug Fix:** Clear safe_symbols before calling incremental analyzer to stop existing errors from being cleared on file update in watch mode

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): Analyzer watch mode

### DIFF
--- a/crates/orchestrator/src/service/incremental_analysis.rs
+++ b/crates/orchestrator/src/service/incremental_analysis.rs
@@ -508,8 +508,8 @@ impl IncrementalAnalysisService {
                 changed_symbols,
             );
 
-            // Clear safe_symbols before analysis — they were needed for populate_codebase_targeted
-            // but must not leak into the analyzer (see non-body-only path for explanation).
+            // Clearing safe_symbols in just_body edits before calling the analyzer makes sure
+            // existing errors aren't cleared.
             merged_codebase.safe_symbols.clear();
             merged_codebase.safe_symbol_members.clear();
 


azjezz:
No, this is not the solution, safe_symbols must stay populated, as that tells the analyzer what symbols were definitly not modified, and should not be analyzed again. The problem is the modified class is somehow ending up there..

Issue #1178, opened 2026-02-20
Analyzer: Edits to one file clear existing issues in another


chrisopperwall-qz:
### Description

I can across some behavior where when saving an unrelated file while running the analyzer in watch mode caused real errors in other files to be removed as issues from the analyzer output, when they were still valid errors.

### Reproduction

1. Create a PHP file with a bad docblock (HasBadDocblock.php)
2. Create an unrelated PHP file with no errors (Unrelated.php)
3. Start `mago analyze --watch`
4. Observe one reported error for the bad docblock
5. Make a small update to Unrelated.php
6. Observe that the watcher output now shoes zero errors

HasBadDocblock.php
```php
<?php

class HasBadDocblock
{
    /**
     * @param string $nonExistent
     */
    public function doWork(int $x): int
    {
        return $x + 1;
    }
}
```

Unrelated.php
```php
<?php

class Unrelated
{
    public function greet(): string
    {
        return "hello"; // Changing this from "hello" to "hi" while the watcher is running will clear the docblock error in HasBadDocblock.php
    }
}

```

Analayze watcher output
```
2026-02-20T04:43:27.147214Z  INFO main mago::commands::analyze: Starting watch mode. Press Ctrl+C to stop.
2026-02-20T04:43:27.150242Z  INFO main mago_database::watcher: Database watcher started for workspace: /Users/chrisopperwall/projects/mago/bug2-codebase-issues-overwrite
2026-02-20T04:43:27.150266Z  INFO main mago::commands::analyze: Watching /Users/chrisopperwall/projects/mago/bug2-codebase-issues-overwrite for changes...
2026-02-20T04:43:27.150276Z  INFO main mago::commands::analyze: Running initial analysis...
error[invalid-param-tag]: The @param tag references an unknown parameter.
  ┌─ HasBadDocblock.php:6:15
  │
6 │      * @param string $nonExistent
  │               ^^^^^^^^^^^^^^^^^^^ Parameter `$nonExistent` is not defined in this function
  │
  = Each `@param` tag in a docblock must correspond to a parameter in the function's signature.
  = Help: Please check for typos or add the parameter to the function signature.

error: found 1 issues: 1 error(s)

2026-02-20T04:43:27.215982Z  INFO main mago::commands::analyze: Initial analysis complete. Watching for changes...
2026-02-20T04:43:38.140035Z  INFO main mago::commands::analyze: Detected 1 file change(s), re-analyzing...
[...]

PR #1179, opened 2026-02-20
fix(analyzer,watcher): preserve issues when unrelated files are saved

chrisopperwall-qz:
## 📌 What Does This PR Do?

This attempts to fix the issue described in #1177, where saving unrelated files while running `mago analyzer --watch` can cause valid errors to disappear in other files.

## 🔍 Context & Motivation

This is an attempt at fixing a bug, this might not be the best way so I'm happy for feedback or to close this if it's not a great way to fix this.

## 🛠️ Summary of Changes

- **Bug Fix:** preserve issues when unrelated files are saved while using watch mode

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Composer Plugin
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): Analyzer with watch mode

### DIFF
--- a/crates/orchestrator/src/service/incremental_analysis.rs
+++ b/crates/orchestrator/src/service/incremental_analysis.rs
@@ -183,6 +183,19 @@ impl IncrementalAnalysisService {
         self.file_states.len()
     }
 
+    /// Merges freshly generated codebase issues into the cached set.
+    ///
+    /// Retains cached issues for files in `files_to_skip` (all symbols safe, not re-analyzed)
+    /// and replaces with `new_issues` for all other files.
+    fn merge_codebase_issues(&mut self, new_issues: IssueCollection, files_to_skip: &HashSet<FileId>) {
+        let old_issues = std::mem::take(&mut self.codebase_issues);
+        self.codebase_issues = old_issues
+            .into_iter()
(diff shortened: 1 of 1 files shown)


Kenneth-Sills:
Ah, beat me to it. Not the maintainer and barely literate with Rust & this codebase, but a couple of notes:

First, we should add regression tests for #1176 and #1178. Now that we are aware of a bad behavior, best to put the replication into the test cases to make sure it never happens again on accident. 😉Wouldn't hurt to add some cases with `analyze_incremental(Some(...))`, either, though not _strictly_ relevant.

Second, I think this is only a partial fix. It will address #1178 *specifically*, but leaves another edge-case open.

When a file has not been changed, but one of its symbols shows up on the reference chain of modified symbol(s) in changed file, they're not "safe". Because of that, they won't be skipped and your new method won't preserve their codex issues. But since they're unchanged their codex won't get re-scanned, and because `take_issues(...)`, the new `merged_codebase.take_issues(true)` won't repeat those found in the last pass and now they'll be lost.

We probably want to delay this merge a couple of lines and, instead of passing `files_to_skip`, pass the IDs of files that still exist (in `current_file_ids`) and have not been changed (in `changed_file_ids`). AKA - all the files who didn't receive a new codex scan, even if they do need a new analysis scan.

PR #1210, opened 2026-02-24
fix(formatter): Fix attribute formatting for anonymous classes

### LINKED ISSUE #1115: formatter: PER-CS anonymous class attributes
### Feature

See https://github.com/php-fig/per-coding-style/pull/136

### Example

```php
https://mago.carthage.software/playground#019c4de5-16d5-0fb6-d9a1-00ef92333470


$example = new
    #[Attribute]
    class {
        ...
    };
```

RemcoSmitsDev:
## 📌 What Does This PR Do?

This PR fixes that attributes for anonymous classes are formatted on the next line right under the new keyword.
This aligns with the psr spec specified in https://github.com/php-fig/per-coding-style/pull/136.

## 🔍 Context & Motivation

This fixes a bug that attributes were not always formatted the same way as the should have. See the before test file for examples that attributes were formatted inconsistently.

## 🛠️ Summary of Changes

- **Bug Fix:** Fixed incorrect formatting of anonymous class attributes.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #1115

## 📝 Notes for Reviewers

First time contributing to the mago project, so the change might not be the right one :).

### DIFF
--- a/crates/formatter/src/internal/format/expression.rs
+++ b/crates/formatter/src/internal/format/expression.rs
@@ -1579,18 +1579,18 @@ impl<'arena> Format<'arena> for AnonymousClass<'arena> {
             let body = print_class_like_body(f, &self.left_brace, &self.members, &self.right_brace, Some(signature_id));
 
             if let Some(attributes) = misc::print_attribute_list_sequence(f, &self.attribute_lists) {
-                Document::Group(
-                    Group::new(vec![
+                Document::Group(Group::new(vec![
+                    in f.arena;
+                    self.new.format(f),
+                    Document::Indent(vec![
                         in f.arena;
-                        self.new.format(f),
-                        Document::Line(Line::default()),
+                        Document::Line(Line::hard()),
                         attributes,
-                        Document::Line(Line::default()),
+                        Document::Line(Line::hard()),
                         signature,
                         body,
-                    ])
-                    .with_break(true),
-                )
+                    ]),
+                ]))
             } else {
                 Document::Group(Group::new(vec![
                     in f.arena;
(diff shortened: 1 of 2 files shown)


crates/formatter/tests/cases/issue_1115/after.php
```diff
@@ -0,0 +1,68 @@
...
+};
+
+$withArgs = new
+#[Attribute]
+class($foo, $bar) {
+    public function __construct(
+        private readonly mixed $foo,
+        private readonly mixed $bar,
+    ) {}
+};
```
azjezz:
```suggestion
$simple = new
    #[Attribute]
    class {
        public function foo(): void {}
    };

$multiple = new
    #[Attribute1]
    #[Attribute2]
    class {
        public function bar(): void {}
    };

$alreadyFormatted = new
    #[Attribute]
    class {
        public function baz(): void {}
    };

$withArgs = new
    #[Attribute]
    class($foo, $bar) {
        public function __construct(
            private readonly mixed $foo,
            private readonly mixed $bar,
        ) {}
    };
```

crates/formatter/src/internal/format/expression.rs
```diff
@@ -1589,12 +1576,30 @@ impl<'arena> Format<'arena> for AnonymousClass<'arena> {
...
+                Document::Group(Group::new(vec![
+                    in f.arena;
+                    self.new.format(f),
+                    Document::space(),
+                    signature,
+                    body,
+                ]))
+            }
         })
     }
```
azjezz, line 1604: As per php-fig/per-coding-style#136 the body should be indented once.

Issue #869, opened 2026-01-09
`mago format` condition blocks formatting breaks with assignment


matthewnessworthy:
### Description

config in `mago.toml`
```
[formatter]
print-width = 120
tab-width = 4
use-tabs = false
method-chain-breaking-style = "next_line"
preserve-breaking-member-access-chain = true
preserve-breaking-argument-list = true
preserve-breaking-array-like = true
preserve-breaking-parameter-list = true
preserve-breaking-attribute-list = true
preserve-breaking-conditional-expression = true
break-promoted-properties-list = true
always-break-named-arguments-list = true
always-break-attribute-named-argument-lists = true
```

[php-fig.org/per/coding-style#51-if-elseif-else](https://www.php-fig.org/per/coding-style/#51-if-elseif-else)
> Expressions in parentheses MAY be split across multiple lines, where each subsequent line is indented at least once. When doing so, the first condition MUST be on the next line. The closing parenthesis and opening brace MUST be placed together on their own line with one space between them. Boolean operators between conditions MUST always be at the beginning.

Before:
```
if (
            $categoryAnySegment
            && (
                ($redirectPath = config('explicit-redirects.any-segment.' . $categoryAnySegment))
                || ($redirectPath = config('explicit-redirects.exclude-first-segment.' . $categoryExcludeFirst))
            )
        ) {
```

After:
```
if (
            $categoryAnySegment
            && (
                ($redirectPath = config('explicit-redirects.any-segment.' . $categoryAnySegment)) || ($redirectPath = config('explicit-redirects.exclude-first-segment.'
                . $categoryExcludeFirst))
            )
        ) {
```

Caveat: @azjezz I know I created https://github.com/carthage-software/mago/issues/753 before, but here is an example where creating a new feature like `preserve-breaking-condition-line` (or something like this), similar to your other `preserve-breaking-*` rules would prevent worse output. I understand your position on compressing statements based on line length, but would you please consider allowing conditions to be placed over multiple lines (still PER-CS compliant) to allow for (IMO) easier to read and understand condition blocks?

### Environment

- Mago: `mago 1.0.3`
- OS: `MacOS 26.1`
- Install: `composer`

PR #1222, opened 2026-02-26
feat(formatter): add an option to preserve breaking conditions

matthewnessworthy:
## 📌 What Does This PR Do?                                                                         
                                                                                                  
Adds a new `preserve_breaking_condition_statement` formatter setting that preserves multi-line      
condition layouts in control structures (`if`, `elseif`, `while`, `do-while`, `switch`, `match`)  
following PER Coding Style 3.0 rules.                                                               
                                                                                                  
## 🔍 Context & Motivation                   

The formatter already has `preserve_breaking_conditional_expression` for ternary expressions, but
there's no equivalent for control structure conditions. When a developer intentionally breaks a
complex condition across multiple lines for readability, the formatter should respect that layout
when this setting is enabled — placing each boolean operator on its own line and the closing
parenthesis on its own line.

[php-fig.org/per/coding-style#51-if-elseif-else](https://www.php-fig.org/per/coding-style/#51-if-elseif-else)
> Expressions in parentheses MAY be split across multiple lines, where each subsequent line is indented at least once. When doing so, the first condition MUST be on the next line. The closing parenthesis and opening brace MUST be placed together on their own line with one space between them. Boolean operators between conditions MUST always be at the beginning.

## 🛠️  Summary of Changes

- **Feature:** Added `preserve_breaking_condition_statement` bool field to `FormatSettings` and
`RawFormatSettings` via the `generate_formatter_settings!` macro, defaulting to `false`.
- **Feature:** Set preset values — Pint, Tempest, Drupal = `true`; Default, PSR-12, Hack = `false`.
- **Feature:** Implemented `must_break` detection in `print_condition()` using
`has_new_line_in_range` between the opening parenthesis and the condition.
- **Feature:** Added `must_break_condition` flag to `FormatterState`, threaded through to
`binaryish.rs` to force hard line breaks before binary operators when inside a must-break condition.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Related to #869

## 📝 Notes for Reviewers

When enabled, a multi-line condition like:
```php
if (
    $foo && $bar
) {
becomes:
if (
    $foo
    && $bar
) {

Single-line conditions are never forced to break. When disabled (the default), existing behavior is
unchanged — all 366 existing tests pass with zero regressions.

### DIFF
--- a/crates/formatter/src/internal/format/binaryish.rs
+++ b/crates/formatter/src/internal/format/binaryish.rs
@@ -315,11 +315,19 @@ fn print_binaryish_expression_parts<'arena>(
 
     let mut right_document = vec![in f.arena];
 
-    right_document.push(if force_break || operator_has_leading_comments || (line_before_operator && !should_inline_this_level) {
-        Document::Line(if force_break { Line::hard() } else if has_space_around { Line::default() } else { Line::soft() })
-    } else {
-        Document::String(if has_space_around { " " } else { "" })
-    });
+    right_document.push(
+        if force_break || operator_has_leading_comments || (line_before_operator && !should_inline_this_level) {
+            Document::Line(if force_break {
+                Line::hard()
+            } else if has_space_around {
+                Line::default()
+            } else {
+                Line::soft()
+            })
+        } else {
+            Document::String(if has_space_around { " " } else { "" })
+        },
+    );
 
     right_document.push(format_token(f, operator.span(), operator.as_str()));
 

### CODE REVIEW

File: .gitignore [hunk 1]
```diff
@@ -8,3 +8,4 @@
 .mago/
 /corpus/vendor/
 /corpus/composer.lock
+.planning/
```
azjezz: Please remove.

azjezz (changes requested):
LGTM. But before we can merge, please remove Claude from git history.

I don't mind using Claude, but it should never be in commit history for git blame :)

matthewnessworthy:
@azjezz sure thing, I'll make the changes soon. 
Thanks for the review and being accepting of the change

azjezz:
@matthewnessworthy you should use `just fix`, it will fix formatting and clippy issue

matthewnessworthy:
@azjezz i ran `just fix`, but it changed a couple of extra files, please let me know if that's fine, or if i should revert that commit in favour of running `cargo fix`

matthewnessworthy:
@azjezz rebased, formatted, and conflicts fixed

matthewnessworthy:
@azjezz rebased and conflicts fixed

docs/tools/formatter/configuration-reference.md
```diff
@@ -141,6 +141,7 @@ public function bar(
 | `preserve-breaking-parameter-list` | `boolean` | `false` | Preserve existing line breaks in parameter lists. |
 | `preserve-breaking-attribute-list` | `boolean` | `false` | Preserve existing line breaks in attribute lists. |
 | `preserve-breaking-conditional-expression` | `boolean` | `false` | Preserve existing line breaks in ternary expressions. |
+| `preserve-breaking-condition-statement` | `boolean` | `false` | Preserve existing line breaks in control structure conditions (`if`, `elseif`, `while`, `do-while`, `switch`, `match`). When enabled, each boolean operator is placed on its own line. |
```
azjezz: nitpick: the option should be named `preserve-breaking-condition-expression` - as it is not a statement, but an expression. or simply `preserve-breaking-condition`

azjezz:
Please run `just fix` locally to fix formatting/clippy issues :)

docs/tools/formatter/configuration-reference.md
matthewnessworthy: no problem, fixed up throughout the PR and commit pushed

PR #1232, opened 2026-02-27
fix(reporting): deterministic analyze output across identical runs

npo-mmenke:
## Summary
This PR fixes nondeterministic `mago analyze` output across identical runs.

The root cause was that some issues contain multiple primary annotations, while formatters selected the first primary annotation encountered. When annotation insertion order varied, the reported `line:column` changed between runs even though the issue set was the same.

Additionally, vendored/builtin invalid-UTF8 diagnostics were emitted at `INFO` level from parallel file loading, which could reorder log lines between runs.

## Changes
- Add deterministic primary annotation selection in `mago-reporting`:
  - `Issue::primary_annotation()`
  - `Issue::primary_span()`
- Use deterministic primary selection in formatters:
  - Emacs
  - GitHub
  - Checkstyle
  - Ariadne
- Use deterministic primary span in:
  - `IssueCollection::sorted()`
  - scoped ignore path resolution
- Reduce nondeterministic loader noise in normal output:
  - invalid UTF-8 diagnostics for vendored/builtin files changed from `info` to `debug`
  - host-file warning behavior remains unchanged
- Add unit test:
  - `test_primary_span_is_deterministic`

### Tests
- `cargo test -p mago-reporting`
- `cargo test -p mago-database`

### Determinism checks (`analyze --reporting-format=emacs --sort`)
Two identical runs per project (official benchmark projects):

- `psl`: identical SHA
- `wordpress`: identical SHA
- `magento`: identical SHA

### Performance check (A/B, interleaved, psl)
- before mean: `0.326s`
- after mean: `0.304s`
- delta mean: `-6.90%`

No performance regression observed.

### DIFF
--- a/crates/database/src/utils.rs
+++ b/crates/database/src/utils.rs
@@ -58,9 +58,7 @@ pub(crate) fn read_file(workspace: &Path, path: &Path, file_type: FileType) -> R
 
         match file_type {
             FileType::Host => tracing::warn!("{}", warning_message),
-            // Vendored/builtin file load order is parallelized; keep these diagnostics at debug
-            // to avoid nondeterministic INFO output ordering between identical runs.
-            FileType::Vendored | FileType::Builtin => tracing::debug!("{}", warning_message),
+            FileType::Vendored | FileType::Builtin => tracing::info!("{}", warning_message),
         }
 
         String::from_utf8_lossy(&bytes).into_owned()


crates/database/src/utils.rs
```diff
@@ -58,7 +58,9 @@ pub(crate) fn read_file(workspace: &Path, path: &Path, file_type: FileType) -> R
 
         match file_type {
             FileType::Host => tracing::warn!("{}", warning_message),
-            FileType::Vendored | FileType::Builtin => tracing::info!("{}", warning_message),
+            // Vendored/builtin file load order is parallelized; keep these diagnostics at debug
+            // to avoid nondeterministic INFO output ordering between identical runs.
+            FileType::Vendored | FileType::Builtin => tracing::debug!("{}", warning_message),
         }
 
         String::from_utf8_lossy(&bytes).into_owned()
```
azjezz: should stay `info`, if you want to hide the log, you can use `MAGO_LOG=warn` env var.

npo-mmenke:
Addressed: restored vendored/builtin invalid-UTF8 diagnostics to info in crates/database/src/utils.rs, pushed to fix/deterministic-analyze-output, and resolved the review thread.

PR #1234, opened 2026-02-27
fix(analyzer): report docblock-type-mismatch for array case

egabai:
## 📌 What Does This PR Do?

Fixes the analyzer to report `docblock-type-mismatch` when a `@var` tag incorrectly narrows an array type (e.g. `@var list<int>` with `list<string|int>`).

## 🔍 Context & Motivation

Previously, Mago allowed `@var` to narrow the inferred type even when the assigned value was not assignable to the declared type. For example:

```php
/** @return list<string|int> */
function foo(): array { return ['foo']; }

/** @var list<int> $val */
$val = foo();  // No error reported
bar($val);     // bar() expects list<int>, but $val is actually list<string|int>
```

The analyzer trusted the `@var` and used `list<int>` for `$val`, hiding the bug. This change makes Mago report `docblock-type-mismatch` in such cases, matching PHPStan's `ReportWrongPhpDocTypeInVarTag` behavior.

## 🛠️ Summary of Changes

- **Bug Fix:** Report `docblock-type-mismatch` when `@var` narrows an array type and the assigned value is not assignable to the declared type.
- **Bug Fix:** Add a clearer error message for this case.
- **Test:** Add `test_var_docblock_override_list_narrow` to cover this scenario.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [x] Analyzer
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

- The new check applies only to **array types** (e.g. `list`, `array`). Narrowing for generic objects (e.g. `@var Suspension<string>` with `Suspension::create()`) and scalars (e.g. `@var string` with `scalar`) is still allowed, as it is often intentional.
- No breaking changes.

### DIFF
--- a/crates/analyzer/src/utils/docblock.rs
+++ b/crates/analyzer/src/utils/docblock.rs
@@ -10,7 +10,10 @@ use mago_codex::ttype::comparator::ComparisonResult;
 use mago_codex::ttype::comparator::union_comparator::can_expression_types_be_identical;
 use mago_codex::ttype::comparator::union_comparator::is_contained_by;
 use mago_codex::ttype::expander;
+use mago_codex::metadata::CodebaseMetadata;
+use mago_codex::ttype::atomic::array::TArray;
 use mago_codex::ttype::expander::TypeExpansionOptions;
+use mago_codex::ttype::get_array_value_parameter;
 use mago_codex::ttype::union::TUnion;
 use mago_codex::ttype::union::populate_union_type;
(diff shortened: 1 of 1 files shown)


azjezz:
This is breaking downstream projects ( specifically Psl ), please check.

Also, please remove cursor from the commits. I don't mind the usage of AI, but it should not be present in git history.

Issue #177, opened 2025-04-04
feature: allow creating custom linting rules using PHP.


azjezz:

ruudk:
This would be epic indeed. Do you have the same in mind for analyzer? One of the things I love doing most is creating PHPStan rules to steer a project in a certain direction. Having the full type context is really good to have in order to not have a rule producing false positives.

azjezz:
for linter, you won't have type information, that's just how the linter is.

But i think if we managed to get this implemented, we would also be able to implement custom [heuristic](https://github.com/carthage-software/mago/tree/main/crates/analyzer/src/heuristic) checks for the analyzer using PHP, i.e ability to post process statements, expressions, class-like, and function-like nodes after they have been fully analyzed, which would give you access to the type information.

azjezz:
from discord:

> would be pretty nice for organizations to be able to create their own rules using php
> something along the lines of:
>
> ```php
> <?php
>
> use Mago\Ast\Node;
> use Mago\Ast\ExitConstruct;
> use Mago\Reporting\Issue;
> 
> Mago\Linter\rule('no-exit', function(Node $node): ?Issue {
>    if (!$node instanceof ExitConstruct) {
>      return null;
>    }
> 
>    return Issue::error("we said no exit!")->withAnnotation(
>      Annotation::primary($node->span())->withMessage("Here!")
>    );
> });
> ```
>
> ```toml
> [linter]
> custom-rules = ["rules/no-exit.php"]
> ```

peterjaap:
+1 for this. We're a Magento agency and we use PhpStan and PHP Codesniffer now, which has [various](https://github.com/bitExpert/phpstan-magento) [custom](https://github.com/mrm-commerce/magento-coding-standard-phpcs) [rulesets](https://github.com/extdn/extdn-phpcs). 

I love Mago due to it's speed and built-in baseline support (which for phpcs still isn't native and the phpstan one isn't flexible enough), so I would love to migrate these rulesets to a Mago ruleset so I can swap out phpcs and phpstan.

peterjaap:
@Bleksak I'd also be happy to PR the rules for Magento to Mago itself, if we can somehow place this behind a flag (or have Mago autodetect Magento?). You said [here](https://github.com/carthage-software/mago/discussions/333) that adding new rules is a fairly straightforward process, maybe you can give a small starter course on how to do it?

peterjaap:
This is what Claude has to say on the topic;

```
[...]

PR #1252, opened 2026-02-28
feat(linter): add Magento 2 lint rules

peterjaap:
## 📌 What Does This PR Do?

Add 11 new lint rules, 7 Magento-gated and 4 general-purpose:

Magento-gated (require integrations = ["magento"]):
- no-object-manager-singleton: flags ObjectManager::getInstance()
- no-object-manager-type-hint: flags ObjectManagerInterface injection
- no-proxy-interceptor-in-constructor: flags Proxy/Interceptor type hints
- no-this-in-template: flags $this in templates (use $block instead)
- no-test-namespace-import: flags test namespace imports in production code
- no-literal-namespace-string: flags hardcoded namespace strings
- no-registry: flags deprecated Magento\Framework\Registry usage

All rules include inline tests. Docs regenerated.

Inspired by phpstan-magento and magento-coding-standard.

## 🔍 Context & Motivation

We would like to move away from phpcs and PhpStan, and adding these rules to Mago allows us to do this.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [x] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Related to #177 

## 📝 Notes for Reviewers

New PHP file to test with that triggers all new linter rules

```php
<?php

namespace Vendor\Module\Model;

use Magento\Framework\ObjectManagerInterface;
use Magento\Framework\Registry;
use Magento\TestFramework\Helper\Bootstrap;
use Vendor\Module\Test\Unit\SomeHelper;

class Example
{
    public function __construct(
        private readonly ObjectManagerInterface $objectManager,
        private readonly \Vendor\Module\Model\Product\Proxy $proxy,
        private readonly \Vendor\Module\Model\Product\Interceptor $interceptor,
        private readonly Registry $registry,
    ) {}

    public function badPractices(): void
    {
        // no-object-manager-singleton
        $om = \Magento\Framework\App\ObjectManager::getInstance();

        // no-registry
        $this->registry->register('key', 'value');

        // no-literal-namespace-string
        $className = 'Magento\Catalog\Model\Product';
        $another = 'Magento\\Catalog\\Api\\ProductRepositoryInterface';

        // no-raw-sql-query
        $sql = "SELECT * FROM catalog_product_entity WHERE entity_id = 1";
        $insert = "INSERT INTO core_config_data (path, value) VALUES ('foo', 'bar')";
        $delete = 'DELETE FROM catalog_product_entity WHERE entity_id = 1';

        // no-direct-exception-throw
        throw new \Exception('Something went wrong');

        // no-array-merge-in-loop
        $result = [];
        foreach ($items as $item) {
            $result = array_merge($result, $item);
        }

        // no-executable-regex-modifier
        $output = preg_replace('/(\w+)/e', 'strtoupper("$1")', $input);
    }
}
```


azjezz:
Please keep PR focused, remove any non-Magento rules, and send those as a seperate PR. let's keep this one for Magento specific rules only :)

peterjaap:
@azjezz done!

peterjaap:
@azjezz no conflicts anymore

crates/linter/src/rule/best_practices/no_literal_namespace_string.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+impl LintRule for NoLiteralNamespaceStringRule {
+    type Config = NoLiteralNamespaceStringConfig;
+
+    fn meta() -> &'static RuleMeta {
+        const META: RuleMeta = RuleMeta {
+            name: "No Literal Namespace String",
+            code: "no-literal-namespace-string",
```
azjezz: this rule should be extracted into its own PR, it applies to all PHP, not just Magento. i find it useful.

However, since it won't require magento, it should be disabled by default.

crates/linter/src/rule/best_practices/no_test_namespace_import.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        ctx.collector.report(
+            Issue::new(self.cfg.level(), "Importing classes from test namespaces in production code is discouraged.")
+                .with_code(self.meta.code)
+                .with_annotation(
+                    Annotation::primary(use_item.name.span()).with_message("This import references a test namespace"),
+                )
+                .with_help("Remove this import and use proper dependency injection or mocking in tests instead."),
+        );
```
azjezz: shouldn't this be allowed if we are in the test namespace?

crates/linter/src/rule/best_practices/no_this_in_template.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        // Flag $this when used outside a class-like scope (common in .phtml templates)
+        // or when explicitly in a .phtml file
+        let in_class = ctx.scope.get_class_like_scope().is_some();
+        let is_template = ctx.source_file.name.ends_with(".phtml");
+
+        if in_class && !is_template {
+            return;
+        }
+
```
azjezz: shouldn't this be `||` not `&&`? what if `.phtml` file contains a class? - that's a different issue by itself.

crates/linter/src/rule/deprecation/no_registry.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        let fqcn = ctx.lookup_name(identifier);
+        if !fqcn.eq_ignore_ascii_case("Magento\\Framework\\Registry") {
+            return;
+        }
+
+        ctx.collector.report(
+            Issue::new(self.cfg.level(), "`Magento\\Framework\\Registry` is deprecated since Magento 2.3.")
+                .with_code(self.meta.code)
+                .with_annotation(Annotation::primary(identifier.span()).with_message("Deprecated Registry usage"))
```
azjezz, line 120: help messages, and notes should always be a single line. note you can have multiple `with_note` messages.

azjezz (changes requested):

crates/linter/src/rule/best_practices/no_test_namespace_import.rs
peterjaap: Good catch. I now check if the namespace contains a test and then skips the check.

crates/linter/src/rule/best_practices/no_this_in_template.rs
peterjaap: Fixed. Simplified the logic to just check if !ctx.source_file.name.ends_with(".phtml") - the rule is specifically about .phtml templates, so there's no need for the class scope check.

wpoortman:
Would be great to have this merged 🫶 .

azjezz:
Tbh, i forgot about this one 😅 

@peterjaap could you please rebase?

peterjaap:
Rebased onto `main` (`36c4cab4`). This turned out to be more than a replay — the branch was based on 1.14.1 and the linter API has moved since, so the rules needed porting:

- `mago_syntax::ast` → `mago_syntax::cst`
- `LintRule::check` now takes the arena type param: `check<'arena, A: Arena>(…, LintContext<'_, 'arena, A>)`
- `lookup_name()`, `Identifier::value()`, `DirectVariable::name` and `File::name` return bytes now, so the FQCN comparisons became `eq_ignore_ascii_case(b"…")`
- serde derives moved to `cfg_attr(feature = "serde", …)`

Three things worth flagging, since they change the diff beyond the rebase itself:

**Squashed to a single commit.** The old branch had intermediate commits adding rules that later commits removed, so replaying them meant resolving the same conflicts several times over. Happy to split it back up if you'd rather review it in pieces.

**Dropped `docs/tools/linter/rules/*.md`.** The rule reference is generated from `RuleMeta` by `docs/src/rules.rs` now, so the hand-written pages were obsolete.

**One change to `crates/linter/tests/rule_examples.rs`** — this is the bit I'd like your call on. The harness lints every rule's examples under a hardcoded `test.php`, but `no-this-in-template` only fires on `.phtml`, so its `bad_example` can't produce an issue and the test fails. I added a small `example_file_name(rule_code)` helper mapping that one rule to `test.phtml`. If you'd rather not have per-rule special-casing in shared test infra, the alternative is dropping `no-this-in-template` from this PR — let me know which you prefer.

Related: default `source.extensions` is `["php"]`, so `no-this-in-template` only does anything for users who add `phtml` to their `mago.toml`. That seemed correct to me, but flagging it in case you'd expect otherwise.

Checks: `cargo check` / `clippy -D warnings` / `fmt --check` all clean, `cargo test -p mago-linter --all-features` 1363 passed. The 6 failures in `mago --test external_analyzer_invocation` reproduce on a clean checkout of `main`, so they're unrelated to this branch.

azjezz:
@peterjaap due to #2192, we will no longer be adding framework or library specific features to Mago itself. such features would need to be their own extensions once 1.47.0 is released.

see https://mago.carthage.software/main/en/extensions/overview/

PR #1254, opened 2026-02-28
feat: add support for `only` and `only-in` filtering in analyzer and linter configuration

WalterWoshid:
## 📌 What Does This PR Do?

- Add the `only` option (global allowlist)
- Add the `only-in` option (scoped filtering)

## 🔍 Context & Motivation

We only use `strict-types` for the linter, would be nice if we wouldn't have to pass it to the CLI all the time.

## 📂 Affected Areas

- [X] Analyzer
- [X] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [X] Documentation

### DIFF
--- a/crates/reporting/src/lib.rs
+++ b/crates/reporting/src/lib.rs
@@ -43,6 +43,48 @@ pub enum IgnoreEntry {
     },
 }
 
+/// Represents an entry in the analyzer's or linter's `only` configuration.
+///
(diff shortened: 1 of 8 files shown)


WalterWoshid:
Draft because implementing `only-in` for scoped-only filtering as well to the `only` option.

Edit: Should be good now.

azjezz:
@WalterWoshid can you rebase please? :)

Issue #888, opened 2026-01-15
Formatter needs multiple runs


dotdash:
### Description

Another example where the formatter needs more than one run to reach the final form.

### Reproduction

Create a file with:
```php
<?php

declare(strict_types=1);

class X {
    public function foo(): void
    {
        $this->derp(
            'zzzzzzzzzzzzzz',
            $this->bar(['zzzzzzzzzzzzzzzzzzzzz' => 'zzzzzzzzzzzzzz'], 'zzzzzzzzzzzzzzzzzzz')
                ->baz(
                    'zzzzzzzzzzzzzzzzzzzz',
                    'zzzzzzzzzzzzzzzzzzzz',
                    'zzzzzzzzzzzzzzzzzzzz',
                    'zzzzzzzzzzzzzzzzzzzz',
                    'zzzzzzzzzzzzzzzzzzzz',
                )
        );
    }
}
```

`mago fmt` once:
```php
<?php

declare(strict_types=1);

class X
{
    public function foo(): void
    {
        $this->derp('zzzzzzzzzzzzzz', $this->bar([
            'zzzzzzzzzzzzzzzzzzzzz' => 'zzzzzzzzzzzzzz',
        ], 'zzzzzzzzzzzzzzzzzzz')->baz(
            'zzzzzzzzzzzzzzzzzzzz',
            'zzzzzzzzzzzzzzzzzzzz',
            'zzzzzzzzzzzzzzzzzzzz',
            'zzzzzzzzzzzzzzzzzzzz',
            'zzzzzzzzzzzzzzzzzzzz',
        ));
    }
}
```

`mago fmt` again:
```php
<?php

declare(strict_types=1);

class X
{
    public function foo(): void
    {
        $this->derp('zzzzzzzzzzzzzz', $this->bar([
            'zzzzzzzzzzzzzzzzzzzzz' => 'zzzzzzzzzzzzzz',
        ], 'zzzzzzzzzzzzzzzzzzz')->baz('zzzzzzzzzzzzzzzzzzzz', 'zzzzzzzzzzzzzzzzzzzz', 'zzzzzzzzzzzzzzzzzzzz', 'zzzzzzzzzzzzzzzzzzzz', 'zzzzzzzzzzzzzzzzzzzz'));
    }
}
```

### Environment

mago@main (Commit 52ec479867e1cc3072681c36c0a1b56537bd346c)

dotdash:
Note that the final result also violates the `print-width`, the line is 160 characters long.

[...]

PR #1260, opened 2026-03-01
fix(formatter): prevent preserve breaks from affecting parent groups

kzmshx:
## 📌 What Does This PR Do?

Separates preserve-sourced line breaks from structural breaks in the formatter's Group IR, so that `preserve_breaking_*` settings no longer affect parent layout decisions.

## 🔍 Context & Motivation

`preserve_breaking_*` settings set `should_break = true` based on source line breaks, which propagates to parent groups via `will_break()` and `propagate_breaks()`. Since formatting changes the source, the second pass produces different propagation and different output.

## 🛠️ Summary of Changes

- Replace `Group.should_break: bool` with `BreakMode` enum (`Auto` / `Force` / `Preserve`). `Preserve` breaks the group but is invisible to `will_break()` and `propagate_breaks()`.
- Replace `Line::hard()` with `Line::default()` for preserved conditionals — hard lines emit `BreakParent` which would re-introduce propagation.
- Split preserve logic from structural break logic in `array.rs`, `parameters.rs`, `expression.rs`, and `mod.rs`.
- Add regression and quality tests. Update `drupal_preset` expected output.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes carthage-software/mago#888

## 📝 Notes for Reviewers

`propagate_breaks()` includes a guard preventing `BreakParent` from promoting `Preserve` to `Force`, which would reintroduce propagation.

The `break_group` variable in `Conditional` is removed — `Line::hard()` already forced propagation via `BreakParent` regardless of the guard. `BreakMode::Preserve` with `Line::default()` replaces both.

The `drupal_preset` output changes because preserved arrays no longer trigger the `will_break()` bailout in `call_arguments`, allowing `expand_last` to be evaluated as a candidate.

### DIFF
--- a/crates/formatter/src/document/mod.rs
+++ b/crates/formatter/src/document/mod.rs
@@ -74,11 +74,22 @@ pub struct Space {
     pub soft: bool,
 }
 
+/// The break mode of a `Group`, determining when it breaks and whether
+/// the break propagates to parent groups via `will_break()`.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
+pub enum BreakMode {
+    /// Break only if the group doesn't fit on a single line.
+    Auto,
+    /// Always break. Propagates to parent groups.
+    Force,
+    /// Always break, but does not propagate to parent groups.
+    Preserve,
+}
+
 #[derive(Debug, Eq, PartialEq, PartialOrd, Ord)]
 pub struct Group<'arena> {
     pub contents: Vec<'arena, Document<'arena>>,
-    pub should_break: RefCell<bool>,
-    pub preserve_source_break: bool,
+    pub break_mode: RefCell<BreakMode>,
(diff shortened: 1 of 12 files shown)


azjezz:
Awesom! thanks @kzmshx

bendavies:
@kzmshx 

This change introduces a regression in formatting for PER coding standards. With 1.13.3, multi-line parameter lists now produce:

```php
public function foo(
  Bar $bar,
): void
{
```

[PER Coding Style 2.0 §4.4 ](https://www.php-fig.org/per/coding-style/#44-methods-and-functions)requires the closing parenthesis and opening brace to be on the same line:

```php
public function foo(
  Bar $bar,
): void {
```

kzmshx:
@bendavies 
Thanks for reporting! Fixed in #1292.

Issue #885, opened 2026-01-14
Feature Request: Enhanced Eloquent ORM Support (Scopes, Relationships, Magic Methods)


rjkooijman:
### Feature

Description

  When using Mago with Laravel projects, the analyzer produces many mixed-method-access and non-documented-method errors due to Eloquent's heavy use of magic methods. While the laravel integration exists for the linter, the analyzer lacks understanding of Eloquent's dynamic features.

  Problem

  Eloquent models use several patterns that Mago cannot currently analyze:

  1. Local Scopes

```php
  // Model definition
  class User extends Model
  {
      public function scopeActive(Builder $query): Builder
      {
          return $query->where('active', true);
      }
  }
```

```php
  // Usage - Mago reports: non-documented-method, mixed-method-access
  User::query()->active()->where('name', 'John')->get();
```

  Mago doesn't understand that scopeActive creates a callable active() method via __call().

  2. Relationship Queries

```php
  // Mago reports: non-documented-method on `comments()`
  $user->comments()->where('approved', true)->get();

  // After calling a scope on a relationship, everything becomes `mixed`
  $post->comments()->active()->latest()->get();
  //                 ^^^^^^^^ mixed-method-access
  //                          ^^^^^^^^ mixed-method-access
```

  3. Dynamic Wheres

```php
  // Mago reports: non-documented-method
  User::whereEmail('test@example.com')->first();
  User::firstOrCreate(['email' => $email]);
```

  4. Relationship Properties

```php
  // Mago reports: non-documented-property
  $user->posts; // Returns Collection, but Mago sees `mixed`
  $user->posts->first()->title; // Cascading mixed-method-access errors
```

  Current Workarounds

  Adding @method docblocks to every model helps partially:

  ```php
/**
   * @method static Builder<User> active()
   * @method static Builder<User> whereEmail(string $email)
   */
  class User extends Model
```

  However, this:
  - Requires manual maintenance
  - Doesn't work when scopes are called on relationships ($user->posts()->active())
  - Doesn't scale for large codebases

  Expected Behavior

  Mago's Laravel integration should:

  1. Recognize local scopes - Methods prefixed with scope should register as callable query builder methods
  2. Understand relationships return query builders - hasMany(), belongsTo(), etc. return typed builders that support chaining
  3. Support dynamic wheres - where{Column}() pattern should be recognized
  4. Infer relationship property types - $user->posts should return Collection<Post>
[...]

PR #1261, opened 2026-03-01
Laravel Eloquent plugin

AJenbo:
This is a port of my Laravel Eloquent module from https://github.com/AJenbo/phpantom_lsp (including 128 tests)

Hope it can serve as a starting point for solving https://github.com/carthage-software/mago/issues/885

Test file:

```php
<?php

namespace App\Models;

use Illuminate\Foundation\Auth\User as Authenticatable;

class User extends Authenticatable
{
    protected function demo(): void
    {
        User::where('id', 1)->firstOrFail();
    }
}
```

Before:
```
2026-03-01T21:34:18.634935Z DEBUG main mago::config: Configuration specifies 32 threads.
2026-03-01T21:34:18.634962Z DEBUG main mago::config: Configuration specifies a stack size of 12582912 bytes.
warning[non-documented-method]: Ambiguous method call to `where` on class `App\Models\User`.
   ┌─ app/Models/User.php:11:15
   │
11 │         User::where('id', 1)->firstOrFail();
   │         ----  ^^^^^ This method is not explicitly defined
   │         │      
   │         On an object of type `App\Models\User`
   │
   = While this call might be handled by `__call()` or `__callStatic()`, Mago cannot verify its arguments or return type without a corresponding `@method` docblock tag.
   = Help: To enable full analysis, add a `@method` tag to the docblock of the `App\Models\User` class. For example: `/** @method returnType where(argType $argName) */`

error[mixed-method-access]: Attempting to access a method on a non-object type (`mixed`).
   ┌─ app/Models/User.php:11:31
   │
11 │         User::where('id', 1)->firstOrFail();
   │         --------------------  ^^^^^^^^^^^ Cannot call method here
   │         │                      
   │         This expression has type `mixed`

error: found 2 issues: 1 error(s), 1 warning(s)

```

After:
```
2026-03-01T21:34:11.880885Z DEBUG main mago::config: Configuration specifies 32 threads.
2026-03-01T21:34:11.880914Z DEBUG main mago::config: Configuration specifies a stack size of 12582912 bytes.
2026-03-01T21:34:13.887421Z  INFO main mago::service: No issues found.

```

### DIFF
--- a/crates/analyzer/src/plugin/libraries/laravel/accessors.rs
+++ b/crates/analyzer/src/plugin/libraries/laravel/accessors.rs
@@ -1,0 +1,210 @@
+//! Accessor property type resolution for Laravel Eloquent models.
+//!
+//! Laravel supports two accessor patterns:
+//!
+//! - **Legacy accessors** (all versions): `getFullNameAttribute(): string` → `$full_name: string`
+//! - **Modern accessors** (Laravel 9+): `fullName(): Attribute<string>` → `$full_name: string`
+//!
+//! This module provides functions to detect accessor methods on a model class
+//! and resolve the corresponding virtual property type.
+//!
+//! Derived from phpantom_lsp's `LaravelModelProvider`:
+//! `is_legacy_accessor()`, `is_modern_accessor()`,
+//! `legacy_accessor_property_name()`, `extract_modern_accessor_type()`.
+
+use mago_codex::metadata::CodebaseMetadata;
+use mago_codex::metadata::class_like::ClassLikeMetadata;
+use mago_codex::metadata::function_like::FunctionLikeMetadata;
+use mago_codex::ttype::atomic::TAtomic;
+use mago_codex::ttype::atomic::object::TObject;
+use mago_codex::ttype::get_mixed;
+use mago_codex::ttype::union::TUnion;
+
+use super::utils::ATTRIBUTE_CLASS;
+use super::utils::accessor_method_candidates;
+use super::utils::extract_short_name;
+use super::utils::is_legacy_getter_accessor;
+
+/// Try to resolve the type for a virtual property produced by an accessor.
+///
+/// Given a property name like `full_name` on a model class, looks for:
+/// 1. A legacy accessor `getFullNameAttribute()` → uses its return type
+/// 2. A modern accessor `fullName(): Attribute<string>` → extracts the
+///    first generic argument from `Attribute`
+///
+/// Returns `Some(TUnion)` with the resolved type, or `None` if no
+/// accessor matches.
+pub fn resolve_accessor_property_type(
+    property_name: &str,
+    class_metadata: &ClassLikeMetadata,
+    codebase: &CodebaseMetadata,
+) -> Option<TUnion> {
+    let candidates = accessor_method_candidates(property_name);
+
+    for candidate in &candidates {
+        if let Some(method_meta) = codebase.get_method(&class_metadata.name, candidate) {
+            // Check if it's a legacy accessor (getXxxAttribute).
+            if is_legacy_getter_accessor(candidate) {
+                return Some(resolve_legacy_accessor_type(method_meta));
+            }
+
+            // Check if it's a modern accessor (returns Attribute).
+            if is_modern_accessor_method(method_meta) {
+                return Some(resolve_modern_accessor_type(method_meta));
+            }
+        }
+    }
+
+    None
+}
+
+/// Determine whether a method is a modern Eloquent accessor (Laravel 9+).
+///
+/// Modern accessors are methods that return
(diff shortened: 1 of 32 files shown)


crates/analyzer/src/plugin/libraries/laravel/accessors.rs
```diff
@@ -0,0 +1,265 @@
...
+        Some(rt) => rt,
+        None => return false,
+    };
+
+    for atomic in return_type.types.iter() {
+        if let TAtomic::Object(TObject::Named(named)) = atomic {
+            let name_str: &str = named.name.as_ref();
+            let clean = strip_leading_backslash(name_str);
+            // Strip generic parameters by taking the base name.
+            let base = clean.split('<').next().unwrap_or(clean).trim();
```
azjezz: class names will never contain generics, or backslash actually...

crates/analyzer/src/plugin/libraries/laravel/accessors.rs
```diff
@@ -0,0 +1,265 @@
...
+        Some(rt) => rt,
+        None => return get_mixed(),
+    };
+
+    for atomic in return_type.types.iter() {
+        if let TAtomic::Object(TObject::Named(named)) = atomic {
+            let name_str: &str = named.name.as_ref();
+            let clean = strip_leading_backslash(name_str);
+            let base = clean.split('<').next().unwrap_or(clean).trim();
+            if base.eq_ignore_ascii_case(ATTRIBUTE_CLASS)
```
azjezz: same here

azjezz (changes requested):

crates/analyzer/src/plugin/libraries/laravel/accessors.rs
AJenbo: arh, a bit to directly ported from PHPantom (it's type system is a bit stringy)

AJenbo:
Ok, clippy should be happy now

AJenbo:
I'm refactoring the PHPantom implementation, splitting it up in multiple section. If you prefer I can rework the porting to bring over one part at a time, that should make it easier to review and test.

https://github.com/AJenbo/phpantom_lsp/tree/main/src/virtual_members/laravel

azjezz:
@AJenbo small changes would be better indeed.

Otherwise, this PR will be open for weeks until i find the time to review the 5000 changes 😅 

And given that i don't use laravel, it would take even more time as i would have to verify every little thing.

AJenbo:
Alright I will do that, at least your not flagging fundamental issues with how I'm trying to bring this over so that's a good start.

azjezz:
I don't see any fundamental things i disagree with, but haven't really read it all. the resolved names passing to plugins and codebase metadata to filter hook are fine, could be useful to other plugins too.

jskorlol:
Does a plugin have to be merged into mago to be usable?
Can't I use an external plugin?

Laravel should support plugins like this.

azjezz:
currently no. we don't have external plugin support, but we will probably in the future, i think once the internal plugin API stabilizes, and we are sure it can cover everything we need/want.

jskorlol:
Thank you always. I'll wait. Until the day I can use it in Laravel.

mmodler:
Hi @AJenbo, what's the current status on this? Are you still planning to split the PR into smaller chunks as discussed?

We're running Mago (formatter + linter) across several Laravel and Statamic (a laravel based CMS) projects and would love to use the analyzer too, but without Eloquent support the false positive noise is too high. This plugin would unblock that for us.

Happy to help with testing against real-world projects if that's useful — we have a mix of Laravel and Statamic codebases.

AJenbo:
Yes it's still planned. My current plan is to finish the Laravel implementation in phpantom first and then bring it over so that I don't have to take to many double feature and bug report duplicates, and avoid refactoring. If you have time to test my LSP that would be helpful, even just running the analyze command on your project would give us an idea of where things stand.

gabrielrbarbosa:
@jskorlol I made a package to workaround these false positives for now maybe it works for you too: https://github.com/gabrielrbarbosa/laramago

dotdash:
Just a heads-up: We updated the `@mixin` handling, which should resolve a few issues with Eloquent where the EloquentQueryBuilder got degraded to a plain QueryBuilder when using certain methods. Will be in the next release.

AJenbo:
needed a fresh start any way

AJenbo:
@azjezz should the extension still be part of this repo or its own thing (and in that case where)?

I can start working on it in about a month.

azjezz:
@AJenbo extensions live outside this repo. see https://mago.carthage.software/main/en/extensions/overview/

AJenbo:
Would it be possible to have it under the carthage-software org? I'm happy to provide the fundamental implementation and work on it to get it to a point where Mago can process a major Laravel application, as well as help maintain it. But I would prefer not to become the project owner / manager.

PR #1277, opened 2026-03-02
fix(database): normalize dot-prefixed source paths in loader

npo-mmenke:
## Summary

This hotfix addresses a path-normalization bug in the database loader where dot-prefixed source globs (for example `./src/**/*.php`) could lead to inconsistent path handling and incorrect exclude behavior.

## Root cause

`load_paths` collected glob and directory results without normalizing them first.  
When configuration used dot-prefixed patterns, path matching could diverge from canonical excludes and produce incorrect file inclusion behavior.

## Fix

- Canonicalize matched glob file paths before collecting them.
- Canonicalize directory roots before recursive walking.
- Keep fallback behavior (`unwrap_or(path)`) to avoid hard-failing on non-canonicalizable paths.

## Regression test

Added a focused regression test:
- `test_dot_prefixed_paths_respect_excludes`

Scenario:
- `paths = ["./src/**/*.php"]`
- excludes cover `src/excluded/**`

Expected and verified:
- excluded file is not loaded
- non-excluded file is loaded
- resulting file count is exactly `1`

## Validation

- `cargo test -p mago-database test_dot_prefixed_paths_respect_excludes -- --nocapture`
- `cargo test -p mago-database`

### DIFF
--- a/crates/database/src/loader.rs
+++ b/crates/database/src/loader.rs
@@ -453,7 +453,8 @@ mod tests {
         create_test_file(&temp_dir, "src/excluded/skip.php", "<?php");
 
         let mut config = create_test_config(&temp_dir, vec!["./src/**/*.php"], vec![]);
-        config.excludes = vec![Exclusion::Path(Cow::Owned(temp_dir.path().join("src/excluded")))];
+        let excluded_dir = temp_dir.path().join("src/excluded").canonicalize().unwrap();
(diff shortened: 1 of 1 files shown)


azjezz:
Thank you for the PR.

However, it seems this issue has already been resolved a while back.

Sorry for missing this PR :)

PR #1296, opened 2026-03-04
lexer: fix octal panics in docblock

### LINKED ISSUE #1295: Analyzer: panic on invalid octal array keys in docblock types
### Description

`mago analyze` panics when array keys contain invalid octal numbers in PHP docblocks.

### Playground Link

https://mago.carthage.software/playground#019cba26-f140-e74f-a419-22077953627a

### Reproduction

Run `mago analyze` on this file:

```php
<?php

/**
 * @return array{08: 'Bar'}
*/
function foo()
{
    return ['08' => 'Bar'];
}
```

See panic:

```
thread '<unnamed>' (85689943) panicked at crates/type-syntax/src/parser/internal/array_like.rs:206:13:
internal error: entered unreachable code: lexer generated invalid integer `08`; this should never happen.
```

### Environment

- Mago: 1.13.3
- OS: MacOS
- Installed via Composer

Bleksak:
fixes #1295

This PR is generated purely by claude code. I have barely reviewed the code changes and I'm not 100% sure they are correct.

### DIFF
--- a/crates/syntax/tests/tokenizer.rs
+++ b/crates/syntax/tests/tokenizer.rs
@@ -1731,4 +1731,3 @@ pub const KEYWORD_TYPES: [(&[u8], TokenKind); 84] = [
     (b"__namespace__", TokenKind::NamespaceConstant),
     (b"__halt_compiler", TokenKind::HaltCompiler),
 ];
-


crates/type-syntax/src/lexer/mod.rs
```diff
@@ -181,6 +181,12 @@ impl<'input> TypeLexer<'input> {
 
         let is_float = matches!(self.input.peek(length, 3), float_separator!());
         if !is_float {
+            if kind == NumberKind::OctalOrFloat {
+                let has_invalid_octal_digit = (1..length).any(|i| matches!(self.input.peek(i, 1), [b'8' | b'9']));
+                if has_invalid_octal_digit {
+                    return (TypeTokenKind::LiteralString, length);
+                }
+            }
             return (TypeTokenKind::LiteralInteger, length);
```
azjezz: Not sure about this 🤔 we want to treat `08` as a literal string in `array{08: string}` but in a case like `@return 'foo'|08` maybe we should error?

azjezz: Actually, i think it would be better to follow PHPStan here, and treat this as a literal integer that is not octal, just 0 prefixed.

see https://phpstan.org/r/a0063a6e-2c19-4803-b978-a76bc616d609

PR #1297, opened 2026-03-05
feat(linter): add fixer for no-alias-function linter rule

edsrzf:
## 📌 What Does This PR Do?

Adds a fixer for the `no-alias-function` linter rule.

## 🔍 Context & Motivation

This rule seems like something that can be easily auto-fixed.

## 🛠️ Summary of Changes

- **Testing:** Added `test_lint_fix` macro to allow testing linter rules with fixers. I didn't see an existing way of doing this.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

(None)

## 📝 Notes for Reviewers

Feedback very welcome around the testing stuff I've added.

Claude Code was used to help author the tests and macro.

### DIFF
--- a/crates/linter/src/rule/consistency/no_alias_function.rs
+++ b/crates/linter/src/rule/consistency/no_alias_function.rs
@@ -1,4 +1,5 @@
 use indoc::indoc;
+use mago_text_edit::Safety;
 use mago_text_edit::TextEdit;
 use schemars::JsonSchema;
 use serde::Deserialize;
@@ -108,7 +109,7 @@ impl LintRule for NoAliasFunctionRule {
 
             ctx.collector.propose(issue, |edits| {
                 let function_span = function_call.function.span();
-                edits.push(TextEdit::replace(function_span, *original_name));
+                edits.push(TextEdit::replace(function_span, *original_name).with_safety(Safety::PotentiallyUnsafe));
             });
 
             break;


crates/linter/src/rule/consistency/no_alias_function.rs
```diff
@@ -105,9 +106,65 @@ impl LintRule for NoAliasFunctionRule {
                 .with_note(format!("The function `{function_name}` is an alias of `{original_name}`."))
                 .with_help(format!("Consider using the function `{original_name}` instead."));
 
-            ctx.collector.report(issue);
+            ctx.collector.propose(issue, |edits| {
+                let function_span = function_call.function.span();
+                edits.push(TextEdit::replace(function_span, *original_name));
```
azjezz: This fix should be marked unsafe.

If i'm using `join` and it gets replaced by `implode` it might call a different function!

```
<?php

namespace App;

function implode(): void {
  echo "IMPLODES";
}

$a = join([], ',');
```

in the example above, it breaks the code, as now `App\implode` gets called, not `implode`.

edsrzf: I was following other rules like `str-contains` and `str-starts-with`. Should those also be marked unsafe?

azjezz: probably.. please open another PR for those..

azjezz: actually, i think potentionally unsafe is the right level here.

PR #1298, opened 2026-03-05
feat(formatter): add parameter_attribute_on_new_line setting

### LINKED ISSUE #1279: formatter: New changes in the code are introduced with `preserve-breaking-parameter-list = true`
### Description

The Mago formatter introduces new changes when `preserve-breaking-parameter-list` is active. Adding a preserve-* option should not introduce new formatting changes.

### Playground Link

_No response_

### Reproduction

If we take this example code:
```php
<?php

function func($aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, #[SensitiveParameter] $b) {
    return true;
}
```
And format it with this config:
```
[formatter]
print-width = 120
tab-width = 4
use-tabs = false
```
It gives an expected result:
```php
<?php

function func(
    $aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
    #[SensitiveParameter] $b,
) {
    return true;
}
```

However, from there, adding this to the formatter config:
`preserve-breaking-parameter-list = true`
Reformats the file once again, giving this result:
```php
<?php

function func(
    $aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
    #[SensitiveParameter]
    $b,
) {
    return true;
}
```

### Environment

- Mago: 1.13.3 (also reproducible with 1.10.0, so not a new bug)
- OS: macOS Tahoe 26.3
- Installed via: Homebrew

kzmshx:
## 📌 What Does This PR Do?

When enabled (default, [PER-CS 12.2](https://www.php-fig.org/per/coding-style/#122-placement) compliant):

```php
function foo(
    #[SensitiveParameter]
    string $password,
) {}
```

When disabled:

```php
function foo(
    #[SensitiveParameter] string $password,
) {}
```

Attributes stay inline when the parameter list fits on one line, regardless of the setting.

## 🔍 Context & Motivation

Previously, attribute-to-parameter line breaks were unconditionally forced whenever the parameter list broke. Users had no way to keep attributes on the same line in a multi-line parameter list.

## 🛠️ Summary of Changes

- Added `parameter_attribute_on_new_line` setting (default: `true`) and enabled it in all presets
- Used `IfBreak` with group ID to conditionally break only when the parameter list group actually breaks
- Added 4 test suites covering enabled/disabled × preserve/no-preserve, each with functions, constructors, closures, and arrow functions

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #1279

## 📝 Notes for Reviewers

Issue #1279 is resolved by making the behavior configurable rather than always inlining attributes. The default keeps the line break for PER-CS compliance; users who prefer inline attributes can opt out.

### DIFF
--- a/crates/formatter/src/settings.rs
+++ b/crates/formatter/src/settings.rs
@@ -478,7 +478,7 @@ generate_formatter_settings! {
     /// ) {}
     /// ```
     ///
-    /// Default: true ([PER-CS 12.2](https://www.php-fig.org/per/coding-style/#122-method-and-function-parameters) compliant)
+    /// Default: true ([PER-CS 12.2](https://www.php-fig.org/per/coding-style/#122-placement) compliant)
     parameter_attribute_on_new_line: bool => "default_true",
 
     /// Whether to add a line before binary operators or after when breaking.
--- a/docs/tools/formatter/configuration-reference.md
+++ b/docs/tools/formatter/configuration-reference.md
@@ -142,6 +142,7 @@ public function bar(
 | `preserve-breaking-attribute-list` | `boolean` | `false` | Preserve existing line breaks in attribute lists. |
 | `preserve-breaking-conditional-expression` | `boolean` | `false` | Preserve existing line breaks in ternary expressions. |
 | `break-promoted-properties-list` | `boolean` | `true` | Always break parameter lists with promoted properties. |
+| `parameter-attribute-on-new-line` | `boolean` | `true` | Place parameter attributes on their own line when the parameter list breaks ([PER-CS 12.2](https://www.php-fig.org/per/coding-style/#122-placement) compliant). |
 | `line-before-binary-operator` | `boolean` | `true` | Place the binary operator on the next line when breaking. |
 | `always-break-named-arguments-list` | `boolean` | `false` | Always break named argument lists into multiple lines. |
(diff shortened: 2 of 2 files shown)


azjezz:
Please update documentation as well.

kzmshx:
@azjezz Done in 47c0ebc. Incorrect PER-CS anchor also fixed.

HeySora:
@kzmshx I understand that the issue I was experiencing in #1279 indeed won't happen after this MR (thank you!); but the underlying issue of #1279 (the fact that a preserve-* option introduces side effects) still exists, if I understand correctly? I do not see any change related to this matter in the formatter, so I believe the issue may still remain, albeit with different effects.

kzmshx:
@HeySora Thanks for the feedback! The cases we identified were fixed in #1260 and this PR. If you're still seeing unexpected changes when enabling the option, a new issue with a reproducing example would be really helpful.

PR #1302, opened 2026-03-06
Feat/symfony preset and linter rules

ostark:
## 📌 What Does This PR Do?

Adds a Symfony formatter preset and Symfony-aligned lint/format rules to close the gap between Mago and PHP-CS-Fixer's `@Symfony` ruleset. 

**All 16 relevant `@Symfony` rules are now covered.**

### Formatter

- **Symfony preset** via `[formatter] preset = "symfony"` — single quotes, trailing commas, brace/spacing conventions
- **`blank_line_before_statement`** — inserts blank line before `return` statements (matches `blank_line_before_statement: [return]`)
- **`ordered_imports`** — alphabetical import sorting with type grouping (classes, functions, constants)
- **PHPDoc reformatter** — trims blank lines, orders tags (@param → @return → @throws), vertical alignment, tag group separation, removes empty docblocks
- **PHPDoc type normalization** — `integer` → `int`, `boolean` → `bool`, `null|string` → `string|null`, `@type` → `@var`

Settings:
```toml
[formatter]
preset = "symfony"
# Symfony preset auto-enables:
#   sort-uses, blank-line-before-return, phpdoc-reformat,
#   phpdoc-align = "vertical", phpdoc-scalar-types, phpdoc-null-position = "last"
```

### Linter (11 new/extended rules)

| Rule | PHP-CS-Fixer equivalent | Auto-fix |
|------|------------------------|----------|
| `yoda-conditions` (extended `style = "non-yoda"`) | `yoda_style: false` | Yes |
| `global-namespace-import` | `global_namespace_import` | Yes |
| `line-length-limit` | Gordinskiy `line_length` plugin | No |
| `prefer-pre-increment` | `increment_style: pre` | Yes |
| `no-alternative-syntax` | `no_alternative_syntax` | No |
| `switch-continue-to-break` | `switch_continue_to_break` | Yes |
| `no-break-comment` | `no_break_comment` | No |
| `no-unused-closure-import` | `no_unused_imports` (closure part) | No |
| `no-null-property-init` | `no_null_property_initialization` | Yes |
| `no-unneeded-import-alias` | `no_unneeded_import_alias` | No |
| `no-short-bool-cast` | `no_short_bool_cast` | Yes |

### @Symfony coverage (16/16)

| Category | Rules |
|----------|-------|
[...]

PR #1306, opened 2026-03-06
feat(formatter): add PHPDoc reformatting support

ostark:
## 📌 What Does This PR Do?

Adds PHPDoc content reformatting to the formatter — parsing and reconstructing multi-line docblock comments with consistent formatting using the Document IR architecture.

## 🔍 Context & Motivation

PHP-CS-Fixer's `@Symfony` ruleset includes several PHPDoc-related fixers (`phpdoc_align`, `phpdoc_order`, etc.). This PR brings similar capabilities to mago's formatter, enabling docblocks to be normalized alongside the rest of the code.

## 🛠️ Summary of Changes

- **Rewrite:** Docblock reformatter now returns `Document<'arena>` IR directly instead of building strings, following the formatter's architecture
- **Feature:** Added `phpdoc_reformat` setting (default: `false`) — when enabled, multi-line docblocks are parsed and reconstructed
- **Feature:** Added `phpdoc_align` setting (`none`/`vertical`) — controls tag column alignment
- **Feature:** Added `phpdoc_null_position` setting (`keep`/`last`) — moves `null` to end of union types
- Uses arena allocation (`BumpVec`, `arena.alloc_str()`) instead of heap `Vec`
- Uses `mago_docblock::tag::split_tag_content` for type extraction instead of custom parsing
- No semantic changes: no tag renaming, no tag deletion, no scalar type normalization

## 📂 Affected Areas

- [x] Formatter
- [ ] Linter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other

## 🔗 Related Issues or PRs

Split from #1302

## 📝 Notes for Reviewers

This uses the existing `mago-docblock` crate for parsing. The reformatting handles tag ordering (param → return → throws → other), blank line trimming, alignment, and null position normalization. All settings are disabled by default — no behavioral change unless opted in.

### DIFF
--- a/crates/formatter/src/internal/comment/docblock.rs
+++ b/crates/formatter/src/internal/comment/docblock.rs
@@ -1,64 +1,56 @@
 use bumpalo::Bump;
+use bumpalo::collections::Vec as BumpVec;
 
 use mago_docblock::document::Element;
 use mago_docblock::document::TagKind;
+use mago_docblock::tag::split_tag_content;
 use mago_span::Span;
 
+use crate::document::BreakMode;
+use crate::document::Document;
+use crate::document::Group;
+use crate::document::Line;
 use crate::settings::FormatSettings;
 use crate::settings::PhpdocAlign;
 use crate::settings::PhpdocNullPosition;
 
 /// A parsed tag line ready for reconstruction.
 struct TagLine<'a> {
-    /// The tag name including `@`, e.g. `@param`.
     tag_name: &'a str,
-    /// The type portion, e.g. `string|null`. Already normalized.
     type_str: Option<String>,
-    /// The variable name, e.g. `$name`.
     variable: Option<String>,
-    /// The remainder description (single-line).
     description: String,
-    /// Sort priority: @param=0, @return=1, @throws=2, other=3.
     priority: u8,
-    /// Original index for stable sort.
     original_index: usize,
-    /// If the tag has multi-line content, store the raw reconstructed lines
-    /// (each line WITHOUT the ` * ` prefix). When set, type_str/variable/description are ignored.
-    raw_lines: Option<Vec<String>>,
+    raw_lines: Option<std::vec::Vec<String>>,
 }
 
-/// Attempt to reformat a docblock comment.
+/// Attempt to reformat a docblock comment, returning a `Document` IR.
 ///
-/// Returns `Some(reformatted)` allocated in the arena, or `None` if the docblock
-/// should be removed entirely (empty docblock → `no_empty_phpdoc`).
+/// Returns `Some(Document)` with the reformatted docblock, or `None` on parse
+/// failure so the caller falls through to default formatting.
 ///
-/// On parse failure, returns `None` so the caller falls through to default formatting.
+/// An empty `Document::String("")` signals an empty docblock.
 pub(crate) fn reformat_docblock<'arena>(
     arena: &'arena Bump,
     settings: &FormatSettings,
     content: &str,
     span: Span,
-) -> Option<&'arena str> {
-    // Only process multi-line docblocks
+) -> Option<Document<'arena>> {
     if !content.starts_with("/**") || !content.contains('\n') {
         return None;
     }
 
-    // Allocate content in the arena for the parser
     let arena_content = arena.alloc_str(content);
-
     let doc = mago_docblock::parse_phpdoc_with_span(arena, arena_content, span).ok()?;
 
-    // Skip docblocks containing code blocks — their formatting is too complex
-    // to reconstruct reliably and idempotently.
     let has_code_block = doc.elements.iter().any(|e| matches!(e, Element::Code(_)));
     if has_code_block {
         return None;
     }
 
-    // Separate description elements from tag elements
-    let mut description_lines: Vec<String> = Vec::new();
-    let mut tag_lines: Vec<TagLine<'_>> = Vec::new();
+    let mut description_lines: std::vec::Vec<String> = std::vec::Vec::new();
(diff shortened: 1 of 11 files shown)


crates/formatter/src/internal/comment/docblock.rs
```diff
@@ -0,0 +1,688 @@
...
+                                } else {
+                                    description_lines
+                                        .push(format!("{{@{} {}}}", tag.name, desc));
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            Element::Code(code) => {
```
azjezz: You already skip documents with code above, so this is unreachable

crates/formatter/src/internal/comment/docblock.rs
```diff
@@ -0,0 +1,688 @@
...
+
+    // Skip docblocks containing code blocks — their formatting is too complex
+    // to reconstruct reliably and idempotently.
+    let has_code_block = doc.elements.iter().any(|e| matches!(e, Element::Code(_)));
+    if has_code_block {
+        return None;
+    }
+
+    // Separate description elements from tag elements
+    let mut description_lines: Vec<String> = Vec::new();
```
azjezz: heap allocations could/would cause high memory usage here, we should use Vec<'arena, &'arena str>

crates/formatter/src/internal/comment/docblock.rs
```diff
@@ -0,0 +1,688 @@
...
+/// Returns `Some(reformatted)` allocated in the arena, or `None` if the docblock
+/// should be removed entirely (empty docblock → `no_empty_phpdoc`).
+///
+/// On parse failure, returns `None` so the caller falls through to default formatting.
+pub(crate) fn reformat_docblock<'arena>(
+    arena: &'arena Bump,
+    settings: &FormatSettings,
+    content: &str,
+    span: Span,
+) -> Option<&'arena str> {
```
azjezz: I don't agree with formatting approach here, of just concating string parts,this does not respect line width, new line settings, indent settings ..etc.

If we are going to introduce phpdoc formatting, it must be done properly, by also implementing `Format` trait to all phpdoc `Document` elements, and for the phpdoc itself, and construct `Document` IR, and let the printer adjust it based on the line width, and, use proper indent ( either spacing or tabs ), and the right new line char ( `\n`, `\r\n`, or `\r` )

crates/formatter/src/internal/comment/docblock.rs
```diff
@@ -0,0 +1,688 @@
...
+            Element::Tag(tag) => {
+                seen_tag = true;
+
+                let kind = tag.kind;
+                let tag_name_str = tag.name;
+
+                // Filter out @access and @package tags
+                if matches!(kind, TagKind::Access | TagKind::Package) {
+                    continue;
+                }
```
azjezz: why? this is deleting stuff from comments, we should never delete anything. if someone is using a documentation generator, this will change their documentation.

crates/formatter/src/internal/comment/docblock.rs
```diff
@@ -0,0 +1,688 @@
...
+
+                // Rename alias tags
+                let effective_name = match kind {
+                    TagKind::Type => "var",
+                    _ => {
+                        if tag_name_str == "link" {
+                            "see"
+                        } else {
```
azjezz: again, we should not change tags, `@link` and `@see` are not the same thing.

crates/formatter/src/internal/comment/docblock.rs
```diff
@@ -0,0 +1,688 @@
...
+                let tag_name_str = tag.name;
+
+                // Filter out @access and @package tags
+                if matches!(kind, TagKind::Access | TagKind::Package) {
+                    continue;
+                }
+
+                // Rename alias tags
+                let effective_name = match kind {
+                    TagKind::Type => "var",
```
azjezz: `@type` is used to define type aliases for static analysis, changing to `@var` breaks static analysis and causes type information loose.

crates/formatter/src/internal/comment/docblock.rs
```diff
@@ -0,0 +1,688 @@
...
+            } else {
+                (None, None, desc.to_string())
+            }
+        }
+        _ => (None, None, desc.to_string()),
+    }
+}
+
+/// Split the leading type annotation from the remaining description.
+fn split_type_from_desc(s: &str) -> (Option<String>, &str) {
```
azjezz: this logic is completely wrong, we have actually functions that do this in mago_phpdoc::tag module.

crates/formatter/src/internal/comment/docblock.rs
```diff
@@ -0,0 +1,688 @@
...
+}
+
+/// Check if a scalar type word starts at this position.
+fn check_scalar_word(s: &str) -> Option<(usize, &'static str)> {
+    let scalars: &[(&str, &str)] = &[
+        ("integer", "int"),
+        ("boolean", "bool"),
+        ("double", "float"),
+        ("real", "float"),
+    ];
```
azjezz: No, we should not change those, `@param Double $double` could refer to a class named that.

https://3v4l.org/Y77dk

azjezz (changes requested):

crates/formatter/src/internal/comment/docblock.rs
ostark: okay

crates/formatter/src/internal/comment/docblock.rs
ostark: okay

azjezz:
Closing this. 

We just added a new PHPDoc parser which would allow us to later implement a proper formatter for PHPDocs

PR #1307, opened 2026-03-06
feat(formatter): add opt-in alignment for parameters and named arguments

psihius:
## 📌 What Does This PR Do?

Adds two new formatter settings, both disabled by default:

- `align-parameters`
- `align-named-arguments`

When enabled, they align multiline parameter lists and multiline named arguments respectively, without changing the default formatter output.

Example for `align-parameters`:

```php
public function __construct(
    private readonly EntityManagerInterface $em,
    private readonly LoggerInterface $logger,
    string $auditLogsS3Endpoint,
    string $auditLogsS3Bucket,
    private readonly AuditLogRepository $auditLogRepository,
) {}
```

becomes:

```php
public function __construct(
    private readonly EntityManagerInterface $em,
    private readonly LoggerInterface        $logger,
    string                                  $auditLogsS3Endpoint,
    string                                  $auditLogsS3Bucket,
    private readonly AuditLogRepository     $auditLogRepository,
) {}
```

Example for `align-named-arguments`:

```php
return new self(
    status: LockCodeAvailabilityStatus::UNAUTHORIZED,
    message: 'For privacy and security reasons...',
    availableFrom: $sharingTime,
);
```

becomes:

```php
return new self(
    status       : LockCodeAvailabilityStatus::UNAUTHORIZED,
    message      : 'For privacy and security reasons...',
    availableFrom: $sharingTime,
);
```

This also applies to multiline attributes:

```php
#[ApiResource(
    normalizationContext: ['groups' => ['host:read']],
    denormalizationContext: ['groups' => ['host:write']],
    graphQlOperations: [
        new Query(),
        new QueryCollection(),
    ],
)]
```

becomes:

```php
#[ApiResource(
    normalizationContext  : ['groups' => ['host:read']],
    denormalizationContext: ['groups' => ['host:write']],
    graphQlOperations     : [
        new Query(),
        new QueryCollection(),
    ],
)]
```

## 🔍 Context & Motivation

There are existing teams and codebases that prefer aligned constructor signatures and aligned multiline named arguments, especially in DTOs, attributes, and service constructors with mixed promoted and non-promoted parameters.

The formatter did not have an opt-in way to express that preference. This change adds that support while keeping defaults and presets unchanged.

## 🛠️ Summary of Changes

- Added `align_parameters` setting (default: `false`)
- Added `align_named_arguments` setting (default: `false`)
- Kept both settings disabled in all presets
- Added formatter support for aligning multiline function and method parameter lists by the variable column
- Added formatter support for aligning multiline named arguments in calls and attributes
- Added regression coverage for enabled and default-off behavior
- Expanded fixtures with real-world mixed constructor and attribute shapes
- Documented both settings in the formatter configuration reference

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [x] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

Verification run:

- `cargo test -p mago-formatter align_parameters -- --exact`
- `cargo test -p mago-formatter align_parameters_default_off -- --exact`
- `cargo test -p mago-formatter align_named_arguments -- --exact`
- `cargo test -p mago-formatter align_named_arguments_default_off -- --exact`
- `cargo test -p mago-formatter breaking_named_arguments -- --exact`
- `cargo test -p mago-formatter attributes -- --exact`
- `cargo test -p mago-formatter preserve_breaking_parameter_list_promoted_properties -- --exact`

### DIFF
--- a/crates/formatter/src/internal/format/parameters.rs
+++ b/crates/formatter/src/internal/format/parameters.rs
@@ -4,6 +4,8 @@ use bumpalo::vec;
 use mago_span::HasSpan;
 use mago_syntax::ast::FunctionLikeParameter;
 use mago_syntax::ast::FunctionLikeParameterList;
+use mago_syntax::ast::Modifier;
+use mago_syntax::ast::Sequence;
 
 use crate::document::BreakMode;
 use crate::document::Document;
@@ -277,18 +279,22 @@ fn can_align_parameter<'arena>(
         parameter_list.parameters.as_slice()[index - 1].end_offset()
     };
 
-    let gap_before_parameter = &f.source_text[gap_start as usize..parameter.start_offset() as usize];
-    let prefix_before_variable =
-        &f.source_text[parameter.start_offset() as usize..parameter.variable.start_offset() as usize];
-
-    !contains_comment_marker(gap_before_parameter) && !contains_comment_marker(prefix_before_variable)
+    !has_comment_trivia_in_range(f, gap_start, parameter.start_offset())
+        && !has_comment_trivia_in_range(f, parameter.start_offset(), parameter.variable.start_offset())
 }
 
-fn contains_comment_marker(text: &str) -> bool {
-    text.contains("/*") || text.contains("//") || text.contains('#')
+fn has_comment_trivia_in_range(f: &FormatterState<'_, '_>, start: u32, end: u32) -> bool {
+    if start >= end {
+        return false;
+    }
+
+    f.all_comments().iter().any(|comment| {
+        let comment_span = comment.span;
+        comment_span.start.offset < end && comment_span.end.offset > start
+    })
 }
 
-fn get_modifier_sequence_width(modifiers: &mago_syntax::ast::Sequence<'_, mago_syntax::ast::Modifier<'_>>) -> usize {
+fn get_modifier_sequence_width(modifiers: &Sequence<'_, Modifier<'_>>) -> usize {
     let ordered_modifiers = [
         modifiers.get_final(),
         modifiers.get_abstract(),
--- a/crates/formatter/tests/cases/align_parameters/after.php
+++ b/crates/formatter/tests/cases/align_parameters/after.php
@@ -36,7 +36,7 @@ final readonly class PlaygroundConstructorExamples
         #[Autowire(service: 'api_platform.graphql.schema_builder')]
         private readonly object $schemaBuilder,
         private readonly string $projectDir,
-        ?ClockInterface $clock = null,
+        ?ClockInterface         $clock = null,
     ) {}
 
     public function auditLogUpload(
--- a/crates/formatter/tests/cases/align_parameters_comment_marker_false_positive/after.php
+++ b/crates/formatter/tests/cases/align_parameters_comment_marker_false_positive/after.php
@@ -1,0 +1,13 @@
+<?php
+
+final class ParameterAttributeMarkerStrings
+{
+    public function __construct(
+        #[Example(pattern: '//')]
+        private readonly FooBarBaz         $first,
+        #[Example(pattern: '/*')]
+        private readonly string            $second,
+        #[Example(pattern: '#')]
+        private readonly ?VeryLongTypeName $third = null,
+    ) {}
+}
--- a/crates/formatter/tests/cases/align_parameters_comment_marker_false_positive/before.php
+++ b/crates/formatter/tests/cases/align_parameters_comment_marker_false_positive/before.php
@@ -1,0 +1,13 @@
+<?php
+
+final class ParameterAttributeMarkerStrings
+{
+    public function __construct(
+        #[Example(pattern: '//')]
+        private readonly FooBarBaz $first,
+        #[Example(pattern: '/*')]
+        private readonly string $second,
+        #[Example(pattern: '#')]
+        private readonly ?VeryLongTypeName $third = null,
+    ) {}
+}
--- a/crates/formatter/tests/cases/align_parameters_comment_marker_false_positive/settings.inc
+++ b/crates/formatter/tests/cases/align_parameters_comment_marker_false_positive/settings.inc
@@ -1,0 +1,4 @@
+FormatSettings {
+    align_parameters: true,
+    ..Default::default()
+}
--- a/crates/formatter/tests/mod.rs
+++ b/crates/formatter/tests/mod.rs
@@ -134,6 +134,7 @@ test_case!(arrow_return);
 test_case!(match_breaking);
 test_case!(array_alignment);
 test_case!(align_parameters);
+test_case!(align_parameters_comment_marker_false_positive);
 test_case!(align_parameters_default_off);
 test_case!(align_named_arguments);
 test_case!(align_named_arguments_default_off);


crates/formatter/src/internal/format/parameters.rs
```diff
@@ -167,3 +178,52 @@ pub(super) fn should_hug_the_only_parameter<'arena>(
...
+}
+
+fn get_parameter_prefix_width<'arena>(
+    f: &mut FormatterState<'_, 'arena>,
+    parameter: &'arena FunctionLikeParameter<'arena>,
+) -> usize {
+    let mut width = 0;
+
+    if !parameter.modifiers.is_empty() {
+        width += get_document_width(&Document::Array(print_modifiers(f, &parameter.modifiers)));
```
azjezz: This is buggy.

If we have comments between modifiers, something like `public /*static*/ readonly`, the call to `print_modifiers` would consume those comments from the formatter state to calculate width, but later, when we are actually printing the modifiers, we would lose the comment, as it would have been consumed...

I'm not 100% sure here, need tests.

crates/formatter/src/internal/format/parameters.rs
```diff
@@ -167,3 +178,52 @@ pub(super) fn should_hug_the_only_parameter<'arena>(
...
+
+    if parameter.ellipsis.is_some() {
+        width += string_width("...");
+    }
+
+    width
+}
+
+fn get_document_width(doc: &Document<'_>) -> usize {
+    match doc {
```
azjezz: we have `get_document_width` function in `crate::internal::format::array` module, i think it makes more more sense to move that to `crate::internal::format::misc` and use it here and there, instead of having duplicate logic.

psihius:
I've addressed the comments with some trully cursed test cases, but seeing the impact on the codebase, i'm not sure this should be addressed. I think that the better way is to throw an error and ask user to fix the comments. Removing last commit and force-pushing is an option :)

psihius:
Followed up on the review comments and simplified the handling here.

The earlier attempt tried to keep alignment in commented parameter lists, but that pushed parameter-specific behavior into the generic comment printer and made the implementation too invasive for a pathological case.

The current rule is simpler: if a parameter list has comments in the alignment span, alignment bails out for that list and normal comment formatting is preserved.

Validation rerun after the change:
- focused formatter tests for parameter/named-argument alignment
- clean playground protocol on `mago-default`

On the playground sweep with `align-parameters = true` and `align-named-arguments = true`, the opt-in diff count dropped from 748 to 725 files after this simplification.

crates/formatter/src/internal/format/parameters.rs
psihius: Moved the shared document-width helper to `format::misc` and reused it from both call sites.

crates/formatter/src/internal/format/parameters.rs
psihius: Changed this to a bailout instead of trying to align through commented parameter prefixes. If there are comments in the parameter gap or prefix, that parameter list no longer participates in alignment. The fixture keeps the messy cases, but they now verify the bailout behavior instead of forcing alignment through them.

crates/formatter/src/internal/format/parameters.rs
psihius: After discussion on Discord, it was decided that if things get too hairy, we just give up trying to align the code

psihius:
One clarification on the `get_document_width` change in `array.rs`.

That helper was not removed as behavior. It was moved from `format::array` into `format::misc` and `array.rs` still calls the shared version. I checked the current call sites before leaving it that way: the live use is still in `array.rs`, with the shared definition in `misc.rs`.

The reason for that move was the review note about duplicated helper logic. It was a local helper extraction, not a formatter behavior change.

That said, the general rule is fair: feature PRs should avoid incidental refactors unless they are directly required to address the review or the implementation would otherwise duplicate logic.

psihius:
Squashed it all into single commit

crates/formatter/src/internal/format/mod.rs
```diff
@@ -1753,6 +1753,17 @@ impl<'arena> Format<'arena> for FunctionLikeParameter<'arena> {
                 contents.push(Document::String("..."));
             }
 
+            if let (Some(padding), Some(list_id)) =
+                (f.parameter_state.variable_padding, f.parameter_state.list_group_id)
+                && padding > 0
+            {
+                let spaces = " ".repeat(padding);
```
azjezz:
```suggestion
            let mut spaces = Vec::with_capacity_in(align.name_padding, f.arena);
            spaces.resize(align.name_padding, b' ');
            let spaces = Document::String(unsafe { std::str::from_utf8_unchecked(spaces.into_bump_slice()) });
```

Avoid heap allocations.

crates/formatter/src/internal/format/parameters.rs
```diff
@@ -167,3 +186,126 @@ pub(super) fn should_hug_the_only_parameter<'arena>(
...
+
+fn parameter_list_exceeds_print_width<'arena>(
+    f: &FormatterState<'_, 'arena>,
+    parameter_list: &'arena FunctionLikeParameterList<'arena>,
+) -> bool {
+    let start = parameter_list.left_parenthesis.start.offset;
+    let end = parameter_list.right_parenthesis.end.offset;
+
+    let source = &f.source_text[start as usize..end as usize];
+    let mut flattened = String::with_capacity(source.len());
```
azjezz: heap allocation should be avoided, we should allocate on the arena.

crates/formatter/src/internal/format/parameters.rs
```diff
@@ -167,3 +186,126 @@ pub(super) fn should_hug_the_only_parameter<'arena>(
...
+    string_width(flattened.trim()) > f.settings.print_width
+}
+
+fn get_max_parameter_prefix_width<'arena>(parameter_list: &'arena FunctionLikeParameterList<'arena>) -> usize {
+    parameter_list.parameters.iter().map(get_parameter_prefix_width).max().unwrap_or(0)
+}
+
+fn needs_promoted_parameter_type_gutter<'arena>(parameter_list: &'arena FunctionLikeParameterList<'arena>) -> bool {
+    // Best-effort alignment for mixed promoted constructors that include an untyped promoted parameter.
+    // See tests/cases/align_parameters/after.php:105.
```
azjezz: remove this comment please

crates/formatter/src/internal/format/parameters.rs
```diff
@@ -167,3 +186,126 @@ pub(super) fn should_hug_the_only_parameter<'arena>(
...
+    } else {
+        parameter_list.parameters.as_slice()[index - 1].end_offset()
+    };
+
+    let gap_before_parameter = &f.source_text[gap_start as usize..parameter.start_offset() as usize];
+    let prefix_before_variable =
+        &f.source_text[parameter.start_offset() as usize..parameter.variable.start_offset() as usize];
+
+    // Best-effort bailout for pathological comment placement in parameter prefixes.
+    // See tests/cases/align_parameters/after.php:81-102.
```
azjezz: same

crates/formatter/src/internal/format/parameters.rs
psihius: Applied in `cdcafab6`. `parameter_list_exceeds_print_width` now builds the flattened buffer with `Vec::with_capacity_in(..., f.arena)` and converts the bump slice back to `&str`, so this path no longer uses a heap-backed `String`.

crates/formatter/src/internal/format/mod.rs
psihius: Applied in `cdcafab6`. This now uses an arena-backed buffer for the padding spaces in `FunctionLikeParameter::format`, following the same pattern as the named-argument alignment path, so it no longer allocates on the heap.

crates/formatter/src/internal/format/parameters.rs
```diff
@@ -167,3 +187,125 @@ pub(super) fn should_hug_the_only_parameter<'arena>(
...
+    }
+
+    if parameter.ellipsis.is_some() {
+        width += string_width("...");
+    }
+
+    width
+}
+
+fn can_align_parameter<'arena>(
```
azjezz, line 270: This is wrong and error prone. if a paramete default value is `"//"` this logic will break because it thinks it contains a comment.

To actually check if a comment exists within a specific range, we must iterate over trivias we have, and check if their positions fall within a given range. not look at the raw source code.

crates/formatter/src/internal/format/parameters.rs
```diff
@@ -167,3 +187,125 @@ pub(super) fn should_hug_the_only_parameter<'arena>(
...
+        &f.source_text[parameter.start_offset() as usize..parameter.variable.start_offset() as usize];
+
+    !contains_comment_marker(gap_before_parameter) && !contains_comment_marker(prefix_before_variable)
+}
+
+fn contains_comment_marker(text: &str) -> bool {
+    text.contains("/*") || text.contains("//") || text.contains('#')
+}
+
+fn get_modifier_sequence_width(modifiers: &mago_syntax::ast::Sequence<'_, mago_syntax::ast::Modifier<'_>>) -> usize {
```
azjezz:
```suggestion
fn get_modifier_sequence_width(modifiers: &Sequence<'_, Modifier<'_>>) -> usize {
```

crates/formatter/src/internal/format/parameters.rs
psihius: Applied in d23bb20a.

`can_align_parameter(...)` now checks actual comment trivia spans from `f.all_comments()` for the two bailout ranges instead of scanning raw source for comment markers. That fixes the false-positive case for comment-like text such as `"//"`, `"/*"`, and `"#"`, and it also stops ordinary `#[...]` attributes from bailing out alignment just because they start with `#`.

I added `align_parameters_comment_marker_false_positive` for the marker-text regression and updated `align_parameters` to cover the attribute case that now aligns correctly.

Validation rerun after the change:
- `just test`
- `just check`
- `cargo fmt --all --check`
- clean playground default sweep
- opt-in `align-parameters = true` playground sweep

crates/formatter/src/internal/format/parameters.rs
psihius: Applied in d23bb20a.

`parameters.rs` now imports `Sequence` and `Modifier`, and `get_modifier_sequence_width(...)` uses the suggested `&Sequence<'_, Modifier<'_>>` signature directly.

psihius:
Addressed the remaining review feedback in `d23bb20a`.

Changes:
- replaced the raw source comment-marker scan in parameter alignment with trivia-span checks over `f.all_comments()`
- applied the suggested `Sequence` / `Modifier` helper signature cleanup
- added a regression fixture for comment-marker false positives and updated the existing alignment fixture for the attribute case that now aligns correctly

Validation rerun after the change:
- `just test`
- `just check`
- `cargo fmt --all --check`
- clean playground default sweep
- opt-in `align-parameters = true` playground sweep

PR #1305, opened 2026-03-06
feat(linter): add new Symfony-related linter rules

ostark:
## 📌 What Does This PR Do?

Adds several new linter rules commonly enforced by PHP-CS-Fixer's `@Symfony` ruleset, plus enhancements to the existing `yoda-conditions` rule.

## 🔍 Context & Motivation

To support a future Symfony linter preset, these rules implement checks that PHP-CS-Fixer enforces under its `@Symfony` configuration. Each rule is disabled by default and can be opted into individually.

## 🛠️ Summary of Changes

- **Feature:** `non-yoda` style for `yoda-conditions` rule — configurable `style` option (`yoda`/`non-yoda`)
- **Feature:** `line-length-limit` rule — reports lines exceeding configurable max length (default: 120), skips comment/string-only lines
- **Feature:** `global-namespace-import` rule — detects fully-qualified references like `\DateTime` and suggests `use` imports
- **Bug Fix:** Exclude function call results from yoda-conditions constant check (matching PHP-CS-Fixer behavior)
- **Feature:** `prefer-pre-increment` rule
- **Feature:** `no-short-bool-cast` rule
- **Feature:** `no-alternative-syntax` rule
- **Feature:** `switch-continue-to-break` rule
- **Feature:** `no-null-property-init` rule
- **Feature:** `no-unneeded-import-alias` rule
- **Feature:** `no-unused-closure-import` rule
- **Feature:** `no-break-comment` rule

## 📂 Affected Areas

- [ ] Formatter
- [x] Linter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other

## 🔗 Related Issues or PRs

Split from #1302

## 📝 Notes for Reviewers

All new rules are disabled by default. The `yoda-conditions` rule now accepts a `style` setting to support both Yoda and non-Yoda enforcement, with correct operator flipping for comparison operators.

PR #1309, opened 2026-03-06
feat(linter): add line-length-limit rule

ostark:
## 📌 What Does This PR Do?

Adds a `line-length-limit` rule that enforces a maximum line length.

## 🔍 Context & Motivation

Long lines reduce readability and cause horizontal scrolling. This rule helps enforce consistent line lengths across a codebase. Common in Symfony and PSR coding standards.

## 🛠️ Summary of Changes

- **Feature:** Added `line-length-limit` rule with configurable limit (default 120 chars)
- Options to exclude comments and URLs from checking

## 📂 Affected Areas

- [x] Linter

## 🔗 Related Issues or PRs

Split from #1305

## 📝 Notes for Reviewers

This rule reports diagnostics only — no auto-fix, as line wrapping is context-dependent.


azjezz:
I'm against this rule.

This should be decided by the formatter, not the linter, the formatter has a line-width setting to control how width lines should be, there is no reason to enforce this by the linter.

While people might use `mago lint` but not `mago fmt`, this is not a reason for us to have two of our tools cover the same thing.

PR #1313, opened 2026-03-06
feat(linter): add no-alternative-syntax rule

ostark:
## 📌 What Does This PR Do?

Adds a `no-alternative-syntax` rule that forbids PHP alternative control structure syntax (`endif`, `endwhile`, `endfor`, `endforeach`, `endswitch`).

## 🔍 Context & Motivation

Alternative syntax reduces code clarity and is discouraged in most modern PHP coding standards. It's mainly used in templates but should be avoided in application code.

## 🛠️ Summary of Changes

- **Feature:** Added `no-alternative-syntax` rule

## 📂 Affected Areas

- [x] Linter

## 🔗 Related Issues or PRs

Split from #1305


yannicka:
It would be good to have this enabled by default in the next major version. I think it would be interesting to have a meta-issue to track these kinds of changes to be made in V2.

azjezz:
@yannicka yes, we will enable this by default starting v2.0.0, we will revisit all rules when releasing that to decide what should be enabled/disabled by default. but we can't enable new ones in minor releases.

PR #1314, opened 2026-03-06
feat(linter): add `switch-continue-to-break` rule

ostark:
## 📌 What Does This PR Do?

Adds a `switch-continue-to-break` rule that detects `continue` inside switch cases and suggests using `break` instead.

## 🔍 Context & Motivation

In PHP, `continue` inside a switch case behaves identically to `break`. Using `continue` is misleading — it suggests loop continuation semantics when it actually just exits the switch case. PHP itself emits a warning for this since PHP 7.3.

## 🛠️ Summary of Changes

- **Feature:** Added `switch-continue-to-break` rule with auto-fix
- Correctly handles `continue 2` (targeting outer loops) as valid

## 📂 Affected Areas

- [x] Linter

## 🔗 Related Issues or PRs

Split from #1305


ostark:
Done — ran `php scripts/linter-rules-docs.php` and pushed the docs update.

PR #1318, opened 2026-03-06
feat(linter): add no-break-comment rule

ostark:
## 📌 What Does This PR Do?

Adds a `no-break-comment` rule that requires a `// no break` comment when a non-empty switch case falls through to the next case.

## 🔍 Context & Motivation

Intentional fall-through in switch cases is often indistinguishable from a missing `break`. Requiring a `// no break` comment documents the intent explicitly. This is a PSR-2 and Symfony coding standard requirement.

## 🛠️ Summary of Changes

- **Feature:** Added `no-break-comment` rule
- Correctly identifies non-empty cases that fall through
- Recognizes existing `// no break` comments (case-insensitive)

## 📂 Affected Areas

- [x] Linter

## 🔗 Related Issues or PRs

Split from #1305

### DIFF
--- a/crates/linter/src/rule/safety/no_break_comment.rs
+++ b/crates/linter/src/rule/safety/no_break_comment.rs
@@ -1,0 +1,524 @@
+use indoc::indoc;
+use schemars::JsonSchema;
+use serde::Deserialize;
+use serde::Serialize;
+
+use mago_reporting::Annotation;
+use mago_reporting::Issue;
+use mago_reporting::Level;
+use mago_span::HasSpan;
+use mago_syntax::ast::Construct;
+use mago_syntax::ast::DeclareBody;
+use mago_syntax::ast::Expression;
+use mago_syntax::ast::IfBody;
+use mago_syntax::ast::Node;
+use mago_syntax::ast::NodeKind;
+use mago_syntax::ast::Statement;
+use mago_syntax::ast::SwitchCase;
+
+use crate::category::Category;
+use crate::context::LintContext;
+use crate::requirements::RuleRequirements;
+use crate::rule::Config;
+use crate::rule::LintRule;
+use crate::rule_meta::RuleMeta;
+use crate::settings::RuleSettings;
+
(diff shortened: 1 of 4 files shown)


crates/linter/src/rule/safety/no_break_comment.rs
```diff
@@ -0,0 +1,524 @@
...
+                || case_lower.contains("// fall-through")
+                || case_lower.contains("// fallthrough")
+                || case_lower.contains("/* fall through")
+                || case_lower.contains("/* fall-through")
+                || case_lower.contains("/* fallthrough")
+                || case_lower.contains("# no break")
+                || case_lower.contains("# no-break")
+                || case_lower.contains("# fall through")
+                || case_lower.contains("# fall-through")
+                || case_lower.contains("# fallthrough");
```
azjezz: we have `program.trivia` which should be used for checking comments, we should never re-parse source code like this, this is slow, and inefficient.

This also would miss `//  no-break` or tabs instead of spaces, we should instead look at trivia, trim the start, and look there.

azjezz (changes requested):

ostark:
Rewritten to use `program.trivia` instead of source text scanning. Comments are now trimmed before matching, so `//  no break` (with extra whitespace) is handled correctly.

azjezz:
rebase needed.

crates/linter/src/rule/safety/no_break_comment.rs
```diff
@@ -0,0 +1,674 @@
...
+    }
+}
+
+fn check_statement<'arena>(
+    rule: &NoBreakCommentRule,
+    ctx: &mut LintContext<'_, 'arena>,
+    statement: &Statement<'arena>,
+    trivia: &Sequence<'arena, Trivia<'arena>>,
+) {
+    match statement {
```
azjezz, line 139: This shoud use a `Walker` implementation, otherwise it is repeating the walker and would need to be updated every time the AST changes.

PR #1319, opened 2026-03-07
feat(formatter): add `preserve-breaking-member-access-chain-first-method-on-same-line` option

psihius:
## 📌 What Does This PR Do?

Improves multiline member-access chain formatting so existing object-chain breaks are preserved more consistently, without forcing static access onto its own line.

This affects cases like:

```php
$context->buildViolation('Problem')
    ->atPath('field')
    ->addViolation()
;
```

which now format to:

```php
$context
    ->buildViolation('Problem')
    ->atPath('field')
    ->addViolation()
;
```

while static access stays attached:

```php
Network\SocketOptions::create()
    ->withAddressReuse(false)
    ->withPortReuse(false)
;
```

## 🔍 Context & Motivation

The previous chain behavior required a tradeoff between preserving useful multiline object chains and avoiding regressions in cases where static access or compact chains should remain attached.

This change narrows the rule so it behaves structurally:
- object `->` chains can break before the first access when the chain is already multiline or too wide
- static `::` stays attached
- semicolon-on-next-line remains controlled by the existing `method-chain-semicolon-on-next-line` setting

## 🛠️ Summary of Changes

- refined first-link handling in member-access chains
- preserve already-broken multiline object chains
- expand long single-line object chains when they exceed `print-width`
- keep static access attached at the front of the chain
- scope semicolon-on-next-line behavior to multiline member-access-chain statements
- added regression coverage for:
  - preserved first-break object chains
  - static factory + object chain combinations
  - assignment / return / nested-scope semicolon handling
- updated formatter config docs for the chain-related settings

## 📂 Affected Areas

- [x] Formatter
- [x] Tests
- [x] Documentation
- [ ] Linter
- [ ] Analyzer
- [ ] CLI behavior unrelated to formatting

## 🔗 Related Issues or PRs

- formatter parity follow-up for fluent member-access chains

## 📝 Notes for Reviewers

Focused formatter tests:
- `cargo test -p mago-formatter preserve_breaking_member_access_chain -- --exact`
- `cargo test -p mago-formatter preserve_breaking_member_access_chain_same_line_first_break -- --exact`
- `cargo test -p mago-formatter method_chain_semicolon_group_scope -- --exact`
- `cargo test -p mago-formatter issue_1105 -- --exact`

Full verification:

### DIFF
--- a/crates/formatter/src/internal/format/member_access.rs
+++ b/crates/formatter/src/internal/format/member_access.rs
@@ -329,6 +329,32 @@ impl<'arena> MemberAccessChain<'arena> {
         false
     }
 
+    fn has_line_break_before_statement_terminator(&self, f: &FormatterState) -> bool {
+        let Some(last_access) = self.accesses.last() else {
+            return false;
+        };
+
+        let chain_end = match last_access.get_arguments_list() {
+            Some(argument_list) => argument_list.span().end.offset,
+            None => last_access.get_selector().span().end.offset,
+        };
+
+        let terminator_start = match f.parent_node() {
+            Node::ExpressionStatement(statement) => statement.terminator.span().start.offset,
+            Node::Return(r#return) => r#return.terminator.span().start.offset,
+            Node::Assignment(_) => match (f.grandparent_node(), f.great_grandparent_node()) {
+                (Some(Node::Expression(_)), Some(Node::ExpressionStatement(statement))) => {
+                    statement.terminator.span().start.offset
+                }
+                (Some(Node::Expression(_)), Some(Node::Return(r#return))) => r#return.terminator.span().start.offset,
+                _ => return false,
+            },
+            _ => return false,
+        };
+
+        misc::has_new_line_in_range(f.source_text, chain_end, terminator_start)
+    }
+
     fn get_flat_width(&self) -> Option<usize> {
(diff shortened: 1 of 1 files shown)


crates/formatter/src/internal/format/member_access.rs
```diff
@@ -265,11 +265,64 @@ impl<'arena> MemberAccessChain<'arena> {
...
+            Some(args) => args.span().end.offset,
+            None => last_access.get_selector().span().end.offset,
+        };
+        let start = self.base.span().start.offset;
+
+        if misc::has_new_line_in_range(f.source_text, start, end) {
+            return false;
+        }
+
+        string_width(&f.source_text[start as usize..end as usize]) > f.settings.print_width
```
azjezz: we should not calculate width this way, as it depends on non formatted content that might have a different width post formatting.

psihius: Replaced the raw source-span width heuristic with a structural flat-width estimate built from the AST, so this no longer depends on pre-formatted source text. The estimate now bails out for unsupported shapes instead of reading source slices, and the original formatter behavior is preserved for the covered chain cases. Verified with `cargo fmt --all --check`, the targeted formatter tests from this PR, `just check`, and the full `just test` workspace suite.

PR #1321, opened 2026-03-07
fix(formatter): avoid aligning compact inline arrays

psihius:
## 📌 What Does This PR Do?

Stops `align-assignment-like` from padding compact inline arrays into columns.

Before this change, enabling `align-assignment-like` could turn:

```php
['status' => 'error', 'data' => $data]
```

into:

```php
['status' => 'error', 'data'   => $data]
```

This PR keeps that compact inline form unchanged, while still allowing alignment for multiline or width-broken mappings.

## 🔍 Context & Motivation

`align-assignment-like` already aligns useful grouped constructs, but it also affected compact inline arrays in ways that added noise without improving readability.

This change narrows the behavior so compact inline arrays stay compact, while meaningful multiline mappings still align.

## 🛠️ Summary of Changes

- limit array key/value alignment to arrays that are already multiline or exceed `print-width`
- keep compact inline arrays unaligned
- add regression coverage for compact inline arrays vs multiline mappings
- update the setting docs to describe the narrower behavior

## 📂 Affected Areas

- [x] Formatter
- [x] Tests
- [x] Documentation
- [ ] Linter
- [ ] Analyzer
- [ ] CLI behavior unrelated to formatting

## 🔗 Related Issues or PRs

- bugfix for `align-assignment-like` array behavior

## 📝 Notes for Reviewers

Focused formatter tests:
- `cargo test -p mago-formatter align_array_like -- --exact`
- `cargo test -p mago-formatter align_array_like_compact_inline -- --exact`
- `cargo test -p mago-formatter issue_1082 -- --exact`

Full verification:
- `cargo fmt --all --check`
- `just test`

### DIFF
--- a/crates/formatter/src/internal/format/array.rs
+++ b/crates/formatter/src/internal/format/array.rs
@@ -16,6 +16,7 @@ use crate::document::Group;
 use crate::document::IfBreak;
 use crate::document::Line;
 use crate::document::clone_in_arena;
+use crate::document::group::GroupIdentifier;
 use crate::internal::FormatterState;
 use crate::internal::format::Format;
 use crate::internal::format::alignment::AlignmentRun;


crates/formatter/src/internal/format/array.rs
```diff
@@ -580,9 +582,14 @@ fn get_document_width(doc: &Document<'_>) -> usize {
...
     f: &FormatterState<'_, 'arena>,
+    array_like: &ArrayLike<'arena>,
     elements: &[&'arena ArrayElement<'arena>],
+    preserve_break: bool,
 ) -> std::vec::Vec<AlignmentRun> {
-    if !f.settings.align_assignment_like || elements.is_empty() {
+    if !f.settings.align_assignment_like
+        || elements.is_empty()
+        || (!preserve_break && !array_like_exceeds_print_width(f, array_like))
+    {
```
azjezz: Shouldn't we just use `Document::IfBreak` to determine if we should indent normal or indent using specific width? this means we dont have to change anything here 🤔

psihius: Addressed this one

crates/formatter/src/internal/format/array.rs
```diff
@@ -665,13 +665,17 @@ fn get_array_element_alignment(runs: &[AlignmentRun], index: usize) -> Option<Al
     runs.iter().find(|run| run.contains(index)).map(|run| run.widths)
 }
 
-fn calculate_array_element_alignment(element: &ArrayElement<'_>, widths: &AlignmentWidths) -> AssignmentAlignment {
+fn calculate_array_element_alignment(
+    element: &ArrayElement<'_>,
+    widths: &AlignmentWidths,
+    break_group_id: crate::document::group::GroupIdentifier,
```
azjezz:
```suggestion
    break_group_id: GroupIdentifier,
```

missing import

PR #1400, opened 2026-03-16
fix(codex): bounds-check shift amounts in const inference

xHeaven:
## 📌 What Does This PR Do?

Bounds-check bit-shift amounts in constant expression inference to prevent panics on out-of-range or negative shifts.

## 🔍 Context & Motivation

`lhs << rhs` and `lhs >> rhs` on `i64` panic in debug mode when `rhs >= 64` or `rhs < 0`, and silently wrap in release mode (producing incorrect values).

## 🛠️ Summary of Changes

- **Bug Fix:** Replaced unchecked `lhs << rhs` / `lhs >> rhs` with `u32::try_from(rhs)` + `checked_shl` / `checked_shr`, returning a generic `int` type for invalid shift amounts.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): `codex`

## 🔗 Related Issues or PRs

I'm unaware of related issues.

## 📝 Notes for Reviewers

Reproduction:
```php
const A = 1 << 65;
```

Run analysis in debug mode, it will panic. Run it in release mode, it will wrap it into `1 << (65 % 64)` which is `2`, PHP produces `0` for `1 << 65`.

### DIFF
--- a/crates/analyzer/tests/cases/bitwise_shift_bounds.php
+++ b/crates/analyzer/tests/cases/bitwise_shift_bounds.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+/** @param int<8, 8> $v */
+function take_8(int $v): void {}
+
+/** @param int<0, 0> $v */
+function take_0(int $v): void {}
+
+const SHIFT_OK = 1 << 3;
+const SHIFT_OVER = 1 << 65;
+const SHIFT_NEG = 1 >> -1;
+
+take_8(SHIFT_OK);
+take_0(SHIFT_OVER);
+take_0(SHIFT_NEG); // @mago-expect analysis:possibly-invalid-argument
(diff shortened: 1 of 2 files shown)


azjezz:
@xHeaven can you please add an analyzer test case reproducing this? just to make sure i don't break it in the future lol

xHeaven:
> @xHeaven can you please add an analyzer test case reproducing this? just to make sure i don't break it in the future lol

Totally, one second!

xHeaven:
Thank god you asked. Actually, `1 << 65` is valid PHP code, so we need to handle it like so, it just returns `0`. `1 << -1` throws `ArithmeticError`, though, so that needs to be a different case. All should be good now.

PR #1414, opened 2026-03-17
fix(analyzer): resolve parent class template params via @extends

### LINKED ISSUE #1412: False positive: template-typed closure parameter not resolved through @extends — shaped array key access yields mixed
### Description

When a generic class has a closure parameter typed with a template (TContext), and a child class binds that template to a shaped array via @extends Base<array{foo: Foo}>, Mago does not propagate the concrete type into the closure. Accessing a key on the closure parameter resolves to mixed instead of the expected shaped array value type, producing a false mixed-argument diagnostic.

```php
<?php

class Foo
{
    public string $value = 'hello';
}

/**
 * @template TContext of array
 */
class Base
{
    /** @var \Closure(TContext, mixed ...): mixed */
    private \Closure $evaluator;

    /**
     * @param \Closure(TContext, mixed ...): mixed $evaluator
     */
    public function __construct(\Closure $evaluator)
    {
        $this->evaluator = $evaluator;
    }
}

/**
 * @extends Base<array{foo: Foo}>
 */
class Child extends Base
{
    public function __construct()
    {
        parent::__construct(
            fn (array $context) => $this->handle($context['foo']),
        );
    }

    private function handle(Foo $foo): string
    {
        return $foo->value;
    }
}
```

https://phpstan.org/r/4dd664a6-2148-4161-835d-1e1414568584

### Playground Link

https://mago.carthage.software/playground#019cfb47-1d65-b0c0-7564-5179c2c514ff

### Reproduction

_No response_

### Environment

latest

xHeaven:
## 📌 What Does This PR Do?

Fixes false positive `mixed-argument` / `mixed-property-access` when calling `parent::` methods from a child class that specializes parent template parameters via `@extends`.

## 🔍 Context & Motivation

When a child class like `@extends Base<array{foo: Foo}>` calls `parent::__construct()`, the analyzer failed to resolve `TContext` to `array{foo: Foo}`. This caused shaped array access like `$context['foo']` to resolve as `mixed`, producing false positive errors.

## 🛠️ Summary of Changes

- **Bug Fix:** Look up template specializations through `@extends` in `populate_template_result_from_invocation` when the calling class differs from the method's declaring class.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): `analyzer`

## 🔗 Related Issues or PRs

Fixes #1412

### DIFF
--- a/crates/analyzer/src/invocation/template_result.rs
+++ b/crates/analyzer/src/invocation/template_result.rs
@@ -103,7 +103,7 @@ pub fn populate_template_result_from_invocation<'ctx, 'arena>(
         }
 
         if let StaticClassType::Object(TObject::Named(instance_type)) = &method_context.class_type
-            && instance_type.name != declaring_class_metadata.name
+            && !instance_type.name.eq_ignore_ascii_case(&declaring_class_metadata.original_name)
             && let Some(calling_class_metadata) = context.codebase.get_class_like(&instance_type.name)
         {
             for (template_name, _) in &declaring_class_metadata.template_types {
@@ -162,7 +162,7 @@ pub fn populate_template_result_from_invocation<'ctx, 'arena>(
         }
     }
 
-    if instance_type.name != method_context.class_like_metadata.name
+    if !instance_type.name.eq_ignore_ascii_case(&method_context.class_like_metadata.original_name)
         && let Some(calling_class_metadata) = context.codebase.get_class_like(&instance_type.name)
     {
         for (template_name, _) in &method_context.class_like_metadata.template_types {


crates/analyzer/src/invocation/template_result.rs
```diff
@@ -135,6 +162,32 @@ pub fn populate_template_result_from_invocation<'ctx, 'arena>(
         }
     }
 
+    if instance_type.name != method_context.class_like_metadata.name
```
azjezz: I think we should compare again `original_name` here? and compare case insensative.

crates/analyzer/src/invocation/template_result.rs
```diff
@@ -102,6 +102,33 @@ pub fn populate_template_result_from_invocation<'ctx, 'arena>(
             }
         }
 
+        if let StaticClassType::Object(TObject::Named(instance_type)) = &method_context.class_type
+            && instance_type.name != declaring_class_metadata.name
```
azjezz: same here

PR #1430, opened 2026-03-18
fix(formatter): mixed-precedence binary chain grouping

xHeaven:
## 📌 What Does This PR Do?

Fixes formatting of mixed-precedence binary chains (e.g. multiplication groups inside addition chains) that regressed after #1407.

## 🔍 Context & Motivation

#1407 correctly removed redundant parens around higher-precedence sub-expressions (`($a * $b) + $c` → `$a * $b + $c`). However those parens also served as layout boundaries. Without them, the `is_wrapped_in_parens` flag from outer contexts like `return (...)` leaked into inner sub-expressions, causing the formatter to flatten all operators onto the same indentation level. Reported downstream in [tempestphp/tempest-framework#2051](https://github.com/tempestphp/tempest-framework/pull/2051).

## 🛠️ Summary of Changes

- **Bug Fix:** In `print_binaryish_expression_parts`, reset `is_wrapped_in_parens` before formatting non-flattenable binary sub-expressions so they produce self-contained groups instead of leaking raw document arrays into the parent chain.
- **Tests:** Added `mixed_precedence_binary_chain` fixture. Updated `issue_301` expectation — `&&` groups now correctly stay compact on one line.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Follow-up to #1407

## 📝 Notes for Reviewers

Other option would be reverting #1407. I'm open to that, too, because it causes this:

```diff
-$this->assertSame((7 * 24 * 60 * 60) + (24 * 60 * 60) + (60 * 60) + 60 + 1, $interval->inSeconds());
+$this->assertSame(7 * 24 * 60 * 60 + 24 * 60 * 60 + 60 * 60 + 60 + 1, $interval->inSeconds());
```

Though the diff is correct, it's probably visually unwanted. Or perhaps we can say that the example is stinky code anyway. What do you think?

### DIFF
--- a/crates/formatter/tests/cases/mixed_precedence_binary_chain/after.php
+++ b/crates/formatter/tests/cases/mixed_precedence_binary_chain/after.php
@@ -1,0 +1,36 @@
+<?php
+
+class Interval
+{
+    public function inSeconds(): int
+    {
+        return (
+            $this->years * 365 * 24 * 60 * 60
+            + $this->months * 30 * 24 * 60 * 60
+            + $this->weeks * 7 * 24 * 60 * 60
+            + $this->days * 60 * 60 * 24
+            + $this->hours * 60 * 60
+            + $this->minutes * 60
+            + $this->seconds
+        );
+    }
+
+    public function inMilliseconds(): int
+    {
+        return (
+            $this->years * 365 * 24 * 60 * 60 * 1000
+            + $this->months * 30 * 24 * 60 * 60 * 1000
+            + $this->weeks * 7 * 24 * 60 * 60 * 1000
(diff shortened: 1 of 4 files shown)


azjezz:
@xHeaven i think reverting #1407 is the right call here..

xHeaven:
> @xHeaven i think reverting #1407 is the right call here..

Alright, sounds good to me. Would you prefer a PR or you will do it by yourself?

azjezz:
@xHeaven PR, i can't push to main anyway, so will have to open a PR myself otherwise lol

PR #1454, opened 2026-03-20
fix(composer): don't spinlock waiting for proc to close

chrisopperwall-qz:
## 📌 What Does This PR Do?

When running `analyze --watch` with a composer installation of mago, the PHP wrapper process will run at 100% on a single core seemingly polling until the process is done. This adds a 500ms delay that will stop the php wrapper from looping so hard.

This also fixes a bug where currently when the mago process gets killed by a signal, it exits with the signal code (so like 2 for SIGINT), instead of the exit code 130 (128 + 2). The $status variable has a boolean termsig which is true when the process was killed by a signal, and a termsig field which is the termsig is. Updating this causes the PHP wrapper to exit with the same exit code that mago did.

## 🔍 Context & Motivation

I noticed that when `mago analyze --watch` was running from a composer installation, a PHP process was using up 100% of a CPU core even when the watcher said it was waiting for changes to be made.

This second fix is less important, but I noticed that when I killed the mago process (the rust binary, not the PHP script), the exit code returned from proc_get_status was -1 instead of the exit code of the mago process. The $status return value has a field to use when the process was sigterm'd, which propagates the right exit code when the php wrapper exits.

This image shows the exit codes being set correctly
Screenshot 2026-03-19 at 6 54 43 PM

This image shows the cpu at 100% while it's waiting for changes
Screenshot 2026-03-19 at 6 59 20 PM

## 🛠️ Summary of Changes

- **Bug Fix:** Stopped spinlocking when running watch mode from a composer install. Also fixed signal-based error code.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): Composer wrapper

### DIFF
--- a/composer/internal.php
+++ b/composer/internal.php
@@ -492,7 +492,7 @@ function execute(string $executablePath, array $args): never
     }
 
     do {
-        usleep(STATUS_CHECK_INTERVAL);
+        \usleep(STATUS_CHECK_INTERVAL);
         $status = proc_get_status($process);
     } while ($status['running']);
 


composer/internal.php
```diff
@@ -489,11 +492,15 @@ function execute(string $executablePath, array $args): never
...
         $status = proc_get_status($process);
     } while ($status['running']);
 
     $exitCode = $status['exitcode'];
-    proc_close($process);
+    if ($status['signaled']) {
+        $exitCode = $status['termsig'] + 128;
+    }
 
+    proc_close($process);
```
azjezz, line 504: we should probably install `php-standard-library/process` here and use it install..

composer/internal.php
```diff
@@ -489,11 +492,15 @@ function execute(string $executablePath, array $args): never
     }
 
     do {
+        \usleep(STATUS_CHECK_INTERVAL);
```
azjezz, line 495: Actually, this is wrong, because now the minimum execution time of Mago is 0.5 seconds. Mago takes 0.3 to analyze Psl. now, it will take 0.5

chrisopperwall-qz: Makes sense to me, thanks for updating it!

PR #1494, opened 2026-03-26
feat(linter): add `no-fully-qualified-global-{function,constant,class}` rules

MidnightDesign:
## Summary

Adds three pedantic linting rules that flag fully-qualified references to global symbols within a namespace:

- **`no-fully-qualified-global-function`**: flags calls like `\strlen()`, suggesting `use function strlen;`
- **`no-fully-qualified-global-constant`**: flags access like `\PHP_VERSION`, suggesting `use const PHP_VERSION;`
- **`no-fully-qualified-global-class`**: flags references like `new \DateTime()`, `\DateTime::method()`, and `\DateTime` in type hints, suggesting `use DateTime;`

All rules are disabled by default (pedantic), with `Level::Help`.

## Test plan

- [x] 14 unit tests covering success and failure cases

### DIFF
--- a/crates/linter/src/rule/consistency/no_fully_qualified_global_class_like.rs
+++ b/crates/linter/src/rule/consistency/no_fully_qualified_global_class_like.rs
@@ -60,13 +60,13 @@ impl NoFullyQualifiedGlobalClassLikeRule {
         let short_name = class_name.split('\\').next_back().unwrap_or(class_name);
 
         ctx.collector.report(
-            Issue::new(self.cfg.level, "Fully-qualified class reference detected.")
+            Issue::new(self.cfg.level, "Fully-qualified class-like reference detected.")
                 .with_code(self.meta.code)
                 .with_annotation(
                     Annotation::primary(identifier.span())
                         .with_message(format!("The reference to `\\{class_name}` uses a fully-qualified name")),
                 )
-                .with_note("Fully-qualified class references bypass the import system, making it harder to see which classes, interfaces, traits, and enums a file depends on.")
+                .with_note("Fully-qualified class-like references bypass the import system, making it harder to see which classes, interfaces, traits, and enums a file depends on.")
                 .with_help(format!("Add `use {class_name};` and reference `{short_name}` directly.")),
         );
     }
@@ -77,15 +77,14 @@ impl LintRule for NoFullyQualifiedGlobalClassLikeRule {
 
     fn meta() -> &'static RuleMeta {
         const META: RuleMeta = RuleMeta {
-            name: "No Fully Qualified Global Class",
+            name: "No Fully Qualified Global Class-Like",
             code: "no-fully-qualified-global-class-like",
             description: indoc! {"
-                Disallows fully-qualified class references within a namespace.
+                Disallows fully-qualified class-like references within a namespace.
 
                 Instead of using the backslash prefix (e.g., `new \\DateTime()` or `\\Exception`
-                in a type hint), prefer an explicit `use` import statement. This applies to all
-                class-like types — classes, interfaces, traits, and enums — and improves
-                readability by keeping imports centralized at the top of the file.
+                in a type hint), prefer an explicit `use` import statement. This improves
+                readability and keeps imports centralized at the top of the file.
             "},
             good_example: indoc! {r#"
                 <?php
--- a/docs/tools/linter/rules/consistency.md
+++ b/docs/tools/linter/rules/consistency.md
@@ -24,7 +24,7 @@ This document details the rules available in the `Consistency` category.
 | Lowercase Type Hint | [`lowercase-type-hint`](#lowercase-type-hint) |
 | No Alias Function | [`no-alias-function`](#no-alias-function) |
 | No Alternative Syntax | [`no-alternative-syntax`](#no-alternative-syntax) |
-| No Fully Qualified Global Class | [`no-fully-qualified-global-class-like`](#no-fully-qualified-global-class-like) |
+| No Fully Qualified Global Class-Like | [`no-fully-qualified-global-class-like`](#no-fully-qualified-global-class-like) |
 | No Fully Qualified Global Constant | [`no-fully-qualified-global-constant`](#no-fully-qualified-global-constant) |
 | No Fully Qualified Global Function | [`no-fully-qualified-global-function`](#no-fully-qualified-global-function) |
 | No Hash Comment | [`no-hash-comment`](#no-hash-comment) |
@@ -665,12 +665,11 @@ endif;
 
 ## <a id="no-fully-qualified-global-class-like"></a>`no-fully-qualified-global-class-like`
 
-Disallows fully-qualified class references within a namespace.
+Disallows fully-qualified class-like references within a namespace.
 
 Instead of using the backslash prefix (e.g., `new \DateTime()` or `\Exception`
-in a type hint), prefer an explicit `use` import statement. This applies to all
-class-like types — classes, interfaces, traits, and enums — and improves
-readability by keeping imports centralized at the top of the file.
+in a type hint), prefer an explicit `use` import statement. This improves
+readability and keeps imports centralized at the top of the file.
 
 
 


MidnightDesign:

crates/linter/src/rule/consistency/no_fully_qualified_global.rs
```diff
@@ -0,0 +1,269 @@
...
+            "#},
+            category: Category::Consistency,
+            requirements: RuleRequirements::None,
+        };
+
+        &META
+    }
+
+    fn targets() -> &'static [NodeKind] {
+        const TARGETS: &[NodeKind] = &[NodeKind::FunctionCall, NodeKind::ConstantAccess];
```
azjezz: wouldn't be better to target `NodeKind::Identifier`? this will catch function call, constant access, but also, class names.

MidnightDesign: But class names don't have the same lookup mechanism, do they? I can't just reference `DateTime` (unqualified, without an import) inside of a namespace. Or am I misunderstanding you?

azjezz: doesn't matter, the rule is just meant to trigger for `\foo` telling user to add `use`. for global namespace, you can already chekc the context to see if we are inside a namespace, if not, tell user to remove `\` because it is redudnant.

MidnightDesign: Right, sorry, my head was in the ambiguous-constant-access rule.

Still: I don't think it's that easy for classes. What if the class isn't imported because there's a conflicting class in the current namespace? (There's an `App\DateTime` and I still want to use the global `\DateTime` in `App\Foo`.)

azjezz: you can still import it, but in case you want to use both, you can import it with an alias, e.g `use DateTime as StdlibDateTime`.

MidnightDesign: So you think those should still be flagged and the user required to import them aliased? I'm not sure I would want that as a user of Mago. I'm aware that it would just be consistent with functions and constants. Maybe I'm just too used to the way other tools handle this.

But if we go that route, what's stopping the rule from being a general "any used symbol must be imported" rule?

azjezz: then i think we split this rule into 3, 1 for functions, 1 for consts, 1 for class-likes.

because someone might also not want to import funcs, or consts..etc. personally, i would want all 3

MidnightDesign: Yeah you're probably right. Keep it all in this PR or spread them across individual PRs?

azjezz: keep them in this PR, but take `ambiguous-constant-access` to another PR, that one is fine and can be merged now.

crates/linter/src/rule/consistency/no_fully_qualified_global_class.rs
```diff
@@ -0,0 +1,231 @@
...
+                let Expression::Identifier(identifier) = access.class else {
+                    return;
+                };
+                identifier
+            }
+            Node::Hint(Hint::Identifier(identifier)) => identifier,
+            _ => return,
+        };
+
+        if !identifier.is_fully_qualified() {
```
azjezz: uncaught: 

- `$x = \Foo::bar(...);`
- `$x = \Foo::$bar;`
- `$x instanceof \Foo`

MidnightDesign: Good catch, didn't think of those.

MidnightDesign: This will be relatively hard to maintain though. I didn't think about this too hard, so this might be a stupid question, but shouldn't the class reference itself just be a node itself so we can just use that as a target?

azjezz: No, a class name is just an `Identifier`. there is no special node for it.

You just have to target `Binary` ( for instanceof ), `StaticPropertyAccess`, and `PartialFunctionApplication` for `Foo::bar(...)` and upcoming `Foo::bar(?)`

azjezz: Actually, also:

`class X implements \Bar {}`, `class X extends \Baz {}`, `class X { use \Qux; }` :D 

all should be caught too..

MidnightDesign:
There are only two success cases for classes - `new` with an import and `new` without a namespace. Is that adequate or should I add success cases for all node kinds? (I'd say it's sufficient as-is.)

crates/linter/src/rule/consistency/no_fully_qualified_global_class.rs
```diff
@@ -0,0 +1,377 @@
...
+use crate::category::Category;
+use crate::context::LintContext;
+use crate::requirements::RuleRequirements;
+use crate::rule::Config;
+use crate::rule::LintRule;
+use crate::rule_meta::RuleMeta;
+use crate::settings::RuleSettings;
+
+#[derive(Debug, Clone)]
+pub struct NoFullyQualifiedGlobalClassRule {
```
azjezz:
```suggestion
pub struct NoFullyQualifiedGlobalClassLikeRule {
```

( and rename rest + file ).

Since this does not catch only classes, but all class-likes.

crates/linter/src/rule/consistency/no_fully_qualified_global_function.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    fn check<'arena>(&self, ctx: &mut LintContext<'_, 'arena>, node: Node<'_, 'arena>) {
+        if ctx.scope.get_namespace().is_empty() {
+            return;
+        }
+
+        let Node::FunctionCall(call) = node else {
+            return;
+        };
+
```
azjezz: Should also target `FunctionPartialApplication` to catch `$x = \foo(...);` :)

azjezz:
Just two things remaining, everything else looks good! :D

crates/linter/src/rule/consistency/no_fully_qualified_global_class.rs
MidnightDesign: DX strings like "Fully-qualified class reference detected" as well? (Just "class" is a tiny bit more readable, but "class-like" is precise and correct.)

MidnightDesign: I feel like "class-like" is a term language nerds like me understand, but a normal developer might just be a bit confused by that term for a second. I'll go with "class" in short strings like titles and the full "classes, interfaces, traits and enums" in longer prose, avoiding "class-like" entirely.

Just give me a heads-up if you disagree with that direction and I'll change it to the short and precise "class-like".

@azjezz

crates/linter/src/rule/consistency/no_fully_qualified_global_function.rs
MidnightDesign: Fixed in https://github.com/carthage-software/mago/pull/1494/commits/0407f27231e3e05a3a674c037872d501725d5fb9

crates/linter/src/rule/consistency/no_fully_qualified_global_class.rs
azjezz: No, `class-like` is fine, this is the term with use in Mago to refer to stuff that's like a class.

MidnightDesign: Got it. Fixed in https://github.com/carthage-software/mago/pull/1494/commits/9dc6fc017766f867e63b5366a45ce6d2c9914af8.

azjezz:
Thanks @MidnightDesign !

MidnightDesign:
Thanks again for guiding me through this @azjezz .

PR #1495, opened 2026-03-26
feat(linter): add `no-unnecessary-docblock-tag` rule

MidnightDesign:
> [!NOTE]
> This PR was written by AI (Claude Code). The full session transcript will be posted as a GitHub Gist.
> I ([@MidnightDesign](https://github.com/MidnightDesign)) don't have too much experience with Rust, so I confirmed manually that:
> 1. The rule descriptions are accurate and readable
> 2. The tests cover all edge cases I could think of
> 3. The code doesn't contain any Rust red flags like `.unwrap()`

## Summary

- Detects `@param` and `@return` docblock tags that merely repeat the type already present in the function/method signature without adding any description.
- Tags that provide more specific types (e.g., `list<string>` for an `array` parameter) or include descriptions are **not** flagged.
- Walks functions, methods in classes/interfaces/traits/enums, and namespaced code.

Disabled by default (pedantic), with `Level::Help`.

## Test plan

- [x] 8 unit tests covering success and failure cases

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/linter/src/rule/mod.rs
+++ b/crates/linter/src/rule/mod.rs
@@ -312,4 +312,5 @@ define_rules! {
     NoDbSchemaChange(no_db_schema_change @ NoDbSchemaChangeRule),
     NoUnescapedOutput(no_unescaped_output @ NoUnescapedOutputRule),
     NoRolesAsCapabilities(no_roles_as_capabilities @ NoRolesAsCapabilitiesRule),
+    NoUnnecessaryDocblockTag(no_unnecessary_docblock_tag @ NoUnnecessaryDocblockTagRule),
 }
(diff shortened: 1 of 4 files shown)


MidnightDesign:

crates/linter/src/rule/consistency/ambiguous_constant_access.rs
```diff
@@ -0,0 +1,231 @@
+use indoc::indoc;
```
azjezz: I think you pushed the wrong code here.. this is same as #1494

MidnightDesign: Sorry about that. Should be the right code now.

MidnightDesign:
We basically have the same issue as in https://github.com/carthage-software/mago/pull/1496#discussion_r2994289292. To get to the docblock, we need the `Program`, so currently the only option seems to be to have `Program` be the target and iterate through the entire codebase.

The only way to solve this is probably to reference doc blocks from their corresponding nodes, so you can just target those nodes in the rule and grab the doc block from there.

azjezz:
Same as #1496 - needs to wait for new docblock parser.

PR #1496, opened 2026-03-26
feat(linter): add `no-shorthand-array-type` rule

### LINKED ISSUE #1016: New Rule: Disallow legacy shorthand array syntax (`T[]`)
### Feature

(Written by hand, rephrased to be more readable by AI.)

### Description

Introduce a new lint rule to disallow the `T[]` shorthand syntax. Although this is a valid shorthand for `array<array-key, T>`, it is often used out of habit rather than intent. This rule would encourage the use of explicit `list<T>` or `array<K, V>` types.

### Motivation

* **Intent over Syntax:** The `T[]` syntax is functionally correct but semantically vague. It doesn't communicate whether the developer expects a sequential list or a specific key-value map.
* **Encouraging Precision:** By disallowing the shorthand, Mago can nudge developers toward more precise type definitions (`list` vs `array`), which leads to better-documented and safer codebases.
* **Modernizing Codebases:** Standardizing on explicit array and list types helps eliminate "legacy-style" documentation, ensuring the codebase adheres to modern PHP typing standards.

### Proposed Behavior

* **Description:** Flags usage of the `T[]` shorthand in docblocks and types.
* **Fixer:** Ideally, a fixer could convert `T[]` to the equivalent `array<array-key, T>`.

### Example

```php
**❌ Prohibited:**


/**
 * @param User[] $users
 */
public function process(array $users): void;



**✅ Preferred:**


/**
 * @param list<User> $users
 */
public function process(array $users): void;

// OR

/**
 * @param array<int, User> $users
 */
public function process(array $users): void;
```

MidnightDesign:
> [!NOTE]
> This PR was written by AI (Claude Code). The full session transcript will be posted as a GitHub Gist.
> I ([@MidnightDesign](https://github.com/MidnightDesign)) don't have too much experience with Rust, so I confirmed manually that:
> 1. The rule descriptions are accurate and readable
> 2. The tests cover all edge cases I could think of
> 3. The code doesn't contain any Rust red flags like `.unwrap()`

## Summary

Disallows imprecise array type syntax in docblock type annotations:

- **`T[]` shorthand** (e.g., `string[]`, `User[]`) — flags each `[]` occurrence, including nested `string[][]`
- **Single-argument generic arrays** (e.g., `array<string>`, `non-empty-array<Foo>`) — should always have an explicit key type

Both forms omit the key type, making them semantically vague. The rule encourages `list<T>` or `array<K, V>` instead.

Disabled by default (pedantic), with `Level::Help`.

Closes #1016

## Test plan

- [ ] `cargo test --workspace --locked` passes
- [x] 13 unit tests covering success and failure cases

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MidnightDesign:

crates/linter/src/rule/consistency/no_shorthand_array_type.rs
```diff
@@ -0,0 +1,467 @@
...
+
+        for trivia in &program.trivia {
+            if trivia.kind != TriviaKind::DocBlockComment {
+                continue;
+            }
+
+            let Ok(document) = mago_docblock::parse_trivia(ctx.arena, trivia) else {
+                continue;
+            };
+
```
azjezz, line 227: I'm not sure we want to this. this will parse every docblock, in every file in the project, and our docblock parser is not really "fast", so this will be a big performance regression for big projects ( I'm assuming - haven't tested it ).

Not sure what the right approach might be..

I'll look into it.

MidnightDesign: Narrowing to targets like `NodeKind::FunctionLikeParameter`, `NodeKind::FunctionLikeReturnTypeHint` and `NodeKind::Assignment` probably won't work because that list is somewhat neverending because types can appear in all kinds of places. Also, I'm not sure if we can even get from such nodes to their doc blocks easily.

A cheap win would be to just pre-filter each docblock by "contains "[]" or "array<"..?

MidnightDesign: @azjezz Did you get a chance to look into this?

azjezz: I did, and i think its not worth adding currently.

We are working on a new docblock parser that will be *much* faster, as we will need it anyway for the LSP, adding rules and new code that uses the only one is not ideal.

I think we should close this and wait until a the new parser has landed.

PR #1508, opened 2026-03-26
feat(linter): add `ambiguous-constant-access` rule

MidnightDesign:
## Summary

Adds `ambiguous-constant-access`: flags unqualified constant access in namespaces without a `use const` import (e.g., `PHP_VERSION` without import). Mirrors the existing `ambiguous-function-call` rule but for constants.

The rule is disabled by default (pedantic), with `Level::Help`.

Extracted from #1494 as requested.

## Test plan

- [x] 6 unit tests covering success and failure cases

### DIFF
--- a/docs/tools/linter/rules/consistency.md
+++ b/docs/tools/linter/rules/consistency.md
@@ -9,6 +9,7 @@ This document details the rules available in the `Consistency` category.
 
 | Rule | Code |
 | :--- | :---------- |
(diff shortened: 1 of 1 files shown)


azjezz:
@MidnightDesign please run `php scripts/linter-rules-docs.php` to update linter documentation.

MidnightDesign:
@azjezz By the way, thank you for helping me out through these PRs. This has been a delightful OSS experience so far.

PR #1544, opened 2026-03-30
fix(composer): prevent race condition when downloading binary concurrently

veewee:
## Summary

- When two `mago` processes start simultaneously (e.g. parallel CI jobs, editor + CLI), both detect the binary is missing and start downloading to the same path. The second process fails because it conflicts with the first's in-progress download or extraction.
- Adds a `locked()` helper that uses `flock(LOCK_EX)` for cross-process mutual exclusion. `ensure_binary()` now acquires a per-triple lock file before downloading, so the second process blocks until the first finishes, then sees the binary already exists and skips the download entirely (double-checked locking pattern).
- Lock files are scoped per version+triple (`.mago-{$triple}.lock`), so downloads for different platforms never block each other.

## How it works

1. Fast path: if the binary exists, return immediately (no lock overhead)
2. Acquire exclusive lock on `{$releaseDir}/.mago-{$triple}.lock` via `flock(LOCK_EX)`. This is a blocking call: if another process holds the lock, the kernel suspends the caller until the lock is released (no polling or busy-waiting).
3. Re-check if binary exists (another process may have completed the download while we waited)
4. If still missing, download and extract as before
5. Lock is released in a `finally` block, even on failure

### DIFF
--- a/composer/internal.php
+++ b/composer/internal.php
@@ -20,6 +20,7 @@ use function escapeshellarg;
 use function escapeshellcmd;
 use function extension_loaded;
 use function fclose;
+use function flock;
 use function file_exists;
 use function file_get_contents;
 use function file_put_contents;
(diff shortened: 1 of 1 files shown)


azjezz:
Internal is growing, i'm wondering what would be the breaking point where i just pull Psl :D

PR #1557, opened 2026-04-01
feat(linter): add no-is-null rule

edsrzf:
## 📌 What Does This PR Do?

Adds a linter rule disallowing the use of `is_null` in favor of strict equality checks (`=== null`), with fixer.

## 🔍 Context & Motivation

Strict comparison is clearer than `is_null`, has no function call overhead, and is the same amount of code.

## 🛠️ Summary of Changes

- **Feature:** Added `redundancy/no-is-null` rule.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

No issue, some discussion on Discord

## 📝 Notes for Reviewers

Some limitations with the fixer:
- The fixer fixes with a Yoda-style comparison so it doesn't cause `yoda` issues, but this could become a problem if #1308 is merged.
- It fixes `!is_null($v)` to `!(null === $v)`. I couldn't find a clean way of handling the inverted case due to the way the linters descend through the AST. It would end up flagging the issue twice: once for the unary expression, and once for the function call. We could add another rule to fix `!($x === $y)`, but it might be nice to have a more general way to handle this type of situation within linters.

Claude Code was used to help author parts of this PR, but I have a complete understanding of the code and have tested it myself.


azjezz:
I'm wondering if we should have a way to look at parent node somehow 🤔

Issue #342, opened 2025-08-31
Add support for Wordpress Rules


rlorenzo:
### 🚀 Describe the Feature

I would love to be able to use this project in my Wordpress setup. Can you support the Wordpress phpcs ruleset: https://github.com/WordPress/WordPress-Coding-Standards

### ✅ Use Case / Example

Wordpress is one of the largest frameworks/projects written in PHP.

### 💡 Proposed Solution

Add the Wordpress ruleset or have the ability import or use phpcs rules.

### 📌 Priority

Must have

azjezz:
Thanks for the suggestion and for pointing us to the `WordPress/WordPress-Coding-Standards` repository! It's a fantastic resource.

I've done a deep dive into the sniffs it contains. Many of them are excellent candidates for new linter rules in Mago. Some, like flagging deprecated WordPress functions or classes, are already handled by our existing static analyzer, but I've extracted a great list of new rules that would add immense value.

This list will form the foundation for our official WordPress integration. We can add more rules in the future, but this is a powerful start.

The rules are broken down into two categories:

  * **PHP General Best Practices:** These rules are beneficial for any PHP project and will be enabled by default in the linter.
  * **WordPress-Specific Rules:** These rules are tailored for WordPress development and will be activated when you enable the WordPress integration in your `mago.toml`.

Here is the proposed list of new linter rules:

-----

## PHP General Best Practices

### No `ini_set` - #422 

  * **code**: `no-ini-set`
  * **description**: This rule discourages dynamically changing PHP configuration at runtime with `ini_set()` or its alias `ini_alter()`. Modifying INI settings can lead to unpredictable behavior, mask server configuration issues, and cause conflicts with other code or the hosting environment.
  * **correct code example**:
    ```php
    // In framework config files (e.g., wp-config.php), use constants.
    define( 'WP_DEBUG', true );

    // Use framework-provided functions where available.
    wp_raise_memory_limit( 'admin' );
    ```
  * **incorrect code example**:
    ```php
    // This can override server settings in an unpredictable way.
    ini_set( 'display_errors', 1 );
    ini_set( 'memory_limit', '256M' );
    ```
  * **what to tell user**:
    > `ini_set()` is discouraged. Use framework-level constants or dedicated functions where available to avoid runtime configuration conflicts.

### Require `preg_quote` Delimiter

  * **code**: `require-preg-quote-delimiter`
[...]

PR #1572, opened 2026-04-04
feat(linter): add WordPress security rules for nonce, SQL preparation and input sanitization

rlorenzo:
## 📌 What Does This PR Do?

I love the project and was among the first to ask for [WordPress support](https://github.com/carthage-software/mago/issues/342). This PR adds three new WordPress security linter rules that detect common CSRF, SQL injection, and unsanitized input vulnerabilities. These rules activate only when `integrations = ["wordpress"]` is configured.

Hopefully, adding these security-related rules will allow more people to transition from phpcs to mago. If this PR is okay, I can contribute other WordPress rules like:

- WordPress.WP.DeprecatedFunctionsSniff
- WordPress.WP.GlobalVariablesOverrideSniff
- WordPress.DB.SlowDBQuerySniff
- WordPress.PHP.DontExtractSniff
- ValidFunctionNameSniff + ValidVariableNameSniff (would require adding support for snake_casing)
- WordPress.WP.EnqueuedResourcesSniff

## 🔍 Context & Motivation

WordPress plugins and themes are frequent targets for security vulnerabilities. This PR brings mago closer to PHPCS parity for WordPress security linting, covering the three most impactful PHPCS sniffs: nonce verification (CSRF), prepared SQL (SQLi), and input sanitization (XSS/injection).

These rules were validated against two production WordPress codebases and compared side-by-side with PHPCS output to ensure parity.

## 🛠️ Summary of Changes

- **Feature:** Added `nonce-verification` rule that flags `$_POST`, `$_GET`, `$_REQUEST`, `$_FILES` access inside functions/methods/closures that lack a call to `wp_verify_nonce()`, `check_admin_referer()`, or `check_ajax_referer()`. Supports custom nonce functions via config. Exempts `isset()`, `empty()`, `unset()`, type-test functions, comparisons, array comparison functions, and type casts.
- **Feature:** Added `prepared-sql` rule that flags `$wpdb` query method calls (`query`, `get_var`, `get_col`, `get_row`, `get_results`, `prepare`) where the SQL string contains variables not passed through `$wpdb->prepare()`. Handles interpolated strings, concatenation, formatting functions (`sprintf`, `vsprintf`), named arguments, and nullsafe property access. Recognizes `absint()`, `intval()`, `floatval()` and numeric casts as safe.
- **Feature:** Added `validated-sanitized-input` rule that flags superglobal access (`$_GET`, `$_POST`, `$_REQUEST`, `$_COOKIE`, `$_SERVER`) without proper sanitization. Enforces `wp_unslash()` wrapping for string sanitizers while recognizing 10 unslashing sanitizers (e.g., `absint`, `intval`, `sanitize_key`) that don't require it. Supports 45+ sanitization functions, partial sanitizer argument tracking (`sanitize_meta`, `sanitize_option`, `wp_kses`), and PHP 8+ named arguments. Exempts `isset()`, `empty()`, `unset()`, type-test functions, comparisons, and array comparison functions.
- **Docs:** Added documentation for all three rules to `docs/tools/linter/rules/security.md` and `docs/tools/linter/rules-and-categories.md`.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [x] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Related to the existing WordPress integration rules: `no-db-schema-change`, `no-direct-db-query`, `no-unescaped-output`, `no-roles-as-capabilities`, `use-wp-functions`.

PHPCS equivalents:
- `WordPress.Security.NonceVerification`
- `WordPress.DB.PreparedSQL`
- `WordPress.Security.ValidatedSanitizedInput`

## 📝 Notes for Reviewers

- All three rules require `integrations = ["wordpress"]`. They will not activate for non-WordPress projects.
- The `prepared-sql` rule deliberately rejects `esc_sql()` as insufficient, preferring `$wpdb->prepare()`. This is stricter than PHPCS, which accepts `esc_sql()`. In the [WordPress docs](https://developer.wordpress.org/reference/functions/esc_sql/), they recommend against relying on `esc_sql()` alone, since it only escapes quotes and does not protect against SQL injection.
- The `validated-sanitized-input` rule tracks partial sanitizers. For example, `sanitize_meta($_POST['key'], $value, 'post')` correctly flags the first argument (key) as unsanitized while accepting the second (value).
- 76 unit tests cover success cases, failure cases, edge cases, and configuration (custom functions). The `rule_examples` integration test also validates good/bad examples.


crates/linter/src/rule/security/nonce_verification.rs
```diff
@@ -0,0 +1,555 @@
...
+
+        // Walk statements in order so accesses before verification are still reported.
+        let mut superglobal_spans: Vec<Span> = Vec::new();
+        let mut nonce_verified = false;
+        for stmt in statements {
+            if !nonce_verified && node_tree_contains_nonce_call(ctx, Node::Statement(stmt), &all_nonce_functions) {
+                nonce_verified = true;
+            }
+            if !nonce_verified {
+                collect_from_node(ctx, Node::Statement(stmt), &mut superglobal_spans, false);
```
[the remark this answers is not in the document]
rlorenzo: Fixed — collect_from_node is now called before checking for nonce calls in the same statement. Also added nonce verification functions to the exempt list so the superglobal inside the nonce call itself is not flagged.

crates/linter/src/rule/security/validated_sanitized_input.rs
```diff
@@ -0,0 +1,884 @@
...
+                    (None, _) => true,
+                };
+
+                if arg_sanitized && needs_unslash {
+                    // String sanitizer: only mark as sanitized if the argument is wrapped in wp_unslash()
+                    let arg_value = match arg {
+                        Argument::Positional(pos) => pos.value,
+                        Argument::Named(named) => named.value,
+                    };
+                    let effectively_sanitized = sanitized || is_wp_unslash_call(ctx, arg_value);
```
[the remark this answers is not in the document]
rlorenzo: Fixed — string sanitizers now check is_wp_unslash_call on the current argument directly, ignoring the inherited sanitized state. This prevents the bypass when nesting string sanitizers under numeric sanitizers.

docs/tools/linter/rules/security.md
```diff
@@ -348,6 +351,94 @@ echo $_GET['user_comment'];
...
+#### Correct code
+
+```php
+<?php
+
+function process_form() {
+    if (!wp_verify_nonce($_POST['_wpnonce'], 'my_action')) {
+        wp_die('Security check failed');
+    }
+    $name = sanitize_text_field($_POST['name']);
```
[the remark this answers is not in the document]
rlorenzo: Fixed — both doc examples now use sanitize_text_field(wp_unslash(...)).

crates/linter/src/rule/security/nonce_verification.rs
```diff
@@ -0,0 +1,555 @@
...
+                (CSRF) attacks. All form and request data processing should verify the nonce first.
+            "},
+            good_example: indoc! {r"
+                <?php
+
+                function process_form() {
+                    if (!wp_verify_nonce($_POST['_wpnonce'], 'my_action')) {
+                        wp_die('Security check failed');
+                    }
+                    $name = sanitize_text_field($_POST['name']);
```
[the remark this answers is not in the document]
rlorenzo: Fixed — good_example now uses sanitize_text_field(wp_unslash($_POST['name'])).

azjezz:
@rlorenzo ignore copilot, i don't know how it was even enabled/triggered...

crates/linter/src/rule/security/nonce_verification.rs
```diff
@@ -0,0 +1,562 @@
...
+    "is_callable",
+    // Array comparison functions — read-only checks, no data processing
+    "in_array",
+    "array_search",
+    "array_keys",
+    // Nonce verification functions themselves — the superglobal passed to these IS the check
+    "wp_verify_nonce",
+    "check_admin_referer",
+    "check_ajax_referer",
+];
```
azjezz, line 53:
```suggestion
const EXEMPT_FUNCTIONS: &[&str] = &[
    "is_array",
    "is_string",
    "is_int",
    "is_integer",
    "is_long",
    "is_float",
    "is_double",
    "is_real",
    "is_numeric",
    "is_bool",
    "is_null",
    "is_object",
    "is_resource",
    "is_callable",
    "in_array",
    "array_search",
    "array_keys",
    "wp_verify_nonce",
    "check_admin_referer",
    "check_ajax_referer",
];
```

crates/linter/src/rule/security/nonce_verification.rs
```diff
@@ -0,0 +1,562 @@
...
+    fn default() -> Self {
+        Self { level: Level::Warning, custom_nonce_functions: Vec::new() }
+    }
+}
+
+impl Config for NonceVerificationConfig {
+    fn level(&self) -> Level {
+        self.level
+    }
+}
```
azjezz, line 82: new linter rules are always disabled by default, this applies to all rules.

```suggestion
    fn default_enabled() -> bool {
        false
    }

    fn level(&self) -> Level {
        self.level
    }
```

crates/linter/src/rule/security/nonce_verification.rs
```diff
@@ -0,0 +1,562 @@
...
+) -> bool {
+    // Check if this node is a function call to a nonce function
+    if let Node::FunctionCall(function_call) = node
+        && function_call_matches_any(ctx, function_call, nonce_functions).is_some()
+    {
+        return true;
+    }
+
+    // Don't descend into nested function/method/closure definitions — they are separate scopes
+    match node.kind() {
```
azjezz: Please don't include redundant comments. the functions are clear as is. this applies to all files.

```suggestion
fn is_nonce_superglobal(name: &str) -> bool {
    NONCE_SUPERGLOBALS.iter().any(|sg| name.eq_ignore_ascii_case(sg))
}

fn is_superglobal_expr(expr: &Expression) -> bool {
    matches!(expr, Expression::Variable(Variable::Direct(var)) if is_nonce_superglobal(var.name))
}

fn node_tree_contains_nonce_call<'arena>(
    ctx: &LintContext<'_, 'arena>,
    node: Node<'_, 'arena>,
    nonce_functions: &[&str],
) -> bool {
    if let Node::FunctionCall(function_call) = node
        && function_call_matches_any(ctx, function_call, nonce_functions).is_some()
    {
        return true;
    }

    match node.kind() {
```

crates/linter/src/rule/security/prepared_sql.rs
```diff
@@ -0,0 +1,508 @@
...
+            .arguments
+            .iter()
+            .find_map(|arg| match arg {
+                Argument::Named(named) if named.name.value.eq_ignore_ascii_case("query") => Some(named.value),
+                _ => None,
+            })
+            .or_else(|| {
+                method_call.argument_list.arguments.first().map(|arg| match arg {
+                    Argument::Positional(arg) => arg.value,
+                    Argument::Named(arg) => arg.value,
```
azjezz: shouldn't this be skipped? because it could be another argument. if this was `query` it would have been found in `find_map`.

crates/linter/src/rule/security/prepared_sql.rs
```diff
@@ -0,0 +1,508 @@
...
+fn is_wpdb_prepare_call(expr: &Expression) -> bool {
+    if let Expression::Call(Call::Method(method_call)) = expr
+        && is_wpdb_variable(method_call.object)
+        && let ClassLikeMemberSelector::Identifier(ident) = &method_call.method
+    {
+        return ident.value.eq_ignore_ascii_case("prepare");
+    }
+
+    false
+}
```
azjezz, line 175:
```suggestion
    if let Expression::Call(Call::Method(method_call)) = expr
        && is_wpdb_variable(method_call.object)
        && let ClassLikeMemberSelector::Identifier(ident) = &method_call.method
    {
        ident.value.eq_ignore_ascii_case("prepare")
    } else {
        false
    }
}
```

crates/linter/src/rule/security/prepared_sql.rs
```diff
@@ -0,0 +1,508 @@
...
+
+/// Check if an expression contains unsafe (non-$wpdb, non-escaped) variables.
+fn contains_unsafe_variables(ctx: &LintContext, expr: &Expression) -> bool {
+    match expr {
+        // Single-quoted string literals have no interpolation — always safe
+        Expression::Literal(Literal::String(_)) => false,
+        // Numeric literals are safe
+        Expression::Literal(Literal::Integer(_) | Literal::Float(_)) => false,
```
azjezz:
```suggestion
        Expression::Literal(_) => false,
```

crates/linter/src/rule/security/prepared_sql.rs
```diff
@@ -0,0 +1,508 @@
...
+                    | UnaryPrefixOperator::FloatCast(..)
+                    | UnaryPrefixOperator::DoubleCast(..)
+                    | UnaryPrefixOperator::RealCast(..)
+                    | UnaryPrefixOperator::BoolCast(..)
+                    | UnaryPrefixOperator::BooleanCast(..)
+            ) =>
+        {
+            false
+        }
+        // Parenthesized expressions
```
azjezz: here, we should probably check for `@` operator prefix, if the oprand is safe, currently, it would report the expression as unsafe below ( in the `_` fallback ).

for `@` simply check the operand again using `contains_unsafe_variables`

azjezz (changes requested):

crates/linter/src/rule/security/prepared_sql.rs
azjezz: same for `&`.

rlorenzo:
@azjezz Changes made

rlorenzo:
@azjezz I rebased this branch off the latest main and squashed the commits into one for cleaner merging on approval. Let me know if any more changes are needed.

rlorenzo:
@azjezz Rebased off the latest main and resolved conflicts

crates/linter/src/rule/security/nonce_verification.rs
```diff
@@ -0,0 +1,553 @@
...
+            "},
+            category: Category::Security,
+            requirements: RuleRequirements::Integration(Integration::WordPress),
+        };
+
+        &META
+    }
+
+    fn targets() -> &'static [NodeKind] {
+        const TARGETS: &[NodeKind] = &[NodeKind::Function, NodeKind::Method, NodeKind::Closure];
```
azjezz: Should target `ArrowFunction` too, the follow code escapes:

```php
fn() => process_post($_POST['x'])
```

crates/linter/src/rule/security/validated_sanitized_input.rs
```diff
@@ -0,0 +1,863 @@
...
+
+#[derive(Debug, Clone)]
+pub struct ValidatedSanitizedInputRule {
+    meta: &'static RuleMeta,
+    cfg: ValidatedSanitizedInputConfig,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(default, rename_all = "kebab-case", deny_unknown_fields)]
+pub struct ValidatedSanitizedInputConfig {
```
azjezz, line 127: Wont 'this rule also trigger for `wp_verify_nonce($_POST['_wpnonce'], ...)`? since `wp_verify_nonce` is not in `SANITIZATION_FUNCTIONS`..

crates/linter/src/rule/security/nonce_verification.rs
```diff
@@ -0,0 +1,553 @@
...
+                MethodBody::Abstract(_) => return,
+            },
+            Node::Closure(closure) => closure.body.statements.as_slice(),
+            _ => return,
+        };
+
+        let custom_refs: Vec<&str> = self.cfg.custom_nonce_functions.iter().map(|s| s.as_str()).collect();
+        let all_nonce_functions: Vec<&str> =
+            DEFAULT_NONCE_FUNCTIONS.iter().copied().chain(custom_refs.iter().copied()).collect();
+
```
azjezz: We should avoid allocating a vector here on every call.

crates/linter/src/rule/security/validated_sanitized_input.rs
```diff
@@ -0,0 +1,863 @@
...
+
+        let all_sanitizers: Vec<&str> = SANITIZATION_FUNCTIONS
+            .iter()
+            .copied()
+            .chain(UNSLASHING_SANITIZERS.iter().copied())
+            .chain(custom_refs.iter().copied())
+            .chain(custom_unslash_refs.iter().copied())
+            .collect();
+
+        let mut issues: Vec<Span> = Vec::new();
```
azjezz: Same, we should avoid allocations here.

For issues, we can emit the issue directly instead of allocating a Vec of N spans.

azjezz (changes requested):
Hey @rlorenzo, sorry for the late review.

I have left some comments below :)

rlorenzo:
@azjezz Thanks for the review!

1. Added ArrowFunction to nonce-verification targets with tests
2. Added a VERIFICATION_FUNCTIONS exempt list so wp_verify_nonce no longer gets a false positive; added tests
3. Avoiding allocating a Vec; emitting instead
4. Avoiding allocating a Vec; emitting instead

crates/linter/src/rule/security/validated_sanitized_input.rs
```diff
@@ -0,0 +1,854 @@
...
+        ctx: &mut LintContext<'_, 'arena>,
+        node: Node<'_, 'arena>,
+        sanitized: bool,
+    ) {
+        match node.kind() {
+            NodeKind::Function | NodeKind::Method | NodeKind::Closure | NodeKind::ArrowFunction => return,
+            _ => {}
+        }
+
+        // `isset`/`empty`/`unset` are validation/cleanup — exempt all superglobal accesses inside.
```
azjezz: Please remove all the inline comments, they are redundant.

azjezz (changes requested):

crates/linter/src/rule/security/validated_sanitized_input.rs
rlorenzo: @azjezz Removed redundant inline comments

nickchomey:
@rlorenzo hey, thanks very much for your efforts to get WP coding standards support into mago. I'm just discovering mago now and would love to use it, but am worried about what standards remain to be supported. Do you have any idea of what's missing still? Perhaps a tracking issue could be created that lists them all?

PR #1582, opened 2026-04-05
feat(linter): add no-service-state-mutation rule for worker-mode safety

### LINKED ISSUE #1579: [Rule Request] no-service-state-mutation — detect stateful services incompatible with worker mode (FrankenPHP/Swoole/RoadRunner)
### Problem

PHP's Shared Memory Model in worker-mode runtimes (FrankenPHP, RoadRunner, Swoole/OpenSwoole) means the application kernel and all singleton services **persist across HTTP requests**. Unlike traditional PHP-FPM where every request starts with a clean process, worker mode keeps services alive indefinitely.

This means any service that mutates `$this->property` outside of `__construct()` will **leak state between requests**. This is the #1 source of worker-mode bugs and causes:

- **Data leakage** between users (User A sees User B's data)
- **Stale state** carried over from previous requests
- **Memory leaks** from properties that accumulate data
- **Non-deterministic behavior** that is extremely difficult to reproduce and debug

With FrankenPHP now backed by the PHP Foundation, Symfony's native FrankenPHP runtime support (available since Symfony 6.4), and worker mode becoming the mainstream deployment model, this class of bug needs first-class static analysis coverage.

### Prior Art

[Phanalist](https://denzyldick.github.io/phanalist/) has an **E0012** rule that does exactly this — it detects `$this->property` assignments inside service methods (outside `__construct()`). We currently use Phanalist as a stopgap for this check, but it has limitations:

- No inline suppression mechanism (only namespace-level exclusions)
- No configurability for allowed methods (e.g., `reset()`)
- Leaks a debug file (`e12_debug.txt`) when the rule fires
- Not actively maintained at the same pace as Mago

Mago's existing Symfony integration, configurable rule system, and inline suppression annotations (`@mago-expect`, `@mago-ignore`) make it a natural fit for this rule.

### Proposed Rule

| Property | Value |
|---|---|
| **Name** | `no-service-state-mutation` |
| **Category** | Safety |
| **Integration** | Symfony |
| **Level** | `warning` (not error) |
| **Auto-fix** | No (requires architectural decision) |
| **Fixable** | No |

#### Detection

The rule should detect any assignment to `$this->property` inside methods other than `__construct()` and `reset()`:

- Direct assignment: `$this->foo = $value;`
- Compound assignment: `$this->foo .= $value;`, `$this->foo += $value;`
- Increment/decrement: `$this->foo++`, `--$this->count`
- Array push: `$this->items[] = $value;`
- Unset: `unset($this->foo);`

#### Configuration

```toml
[rules.no-service-state-mutation]
level = "warning"

# Namespaces to check (services live here)
include-namespaces = ["App\\"]

# Namespaces to exclude (entities, DTOs, value objects hold state legitimately)
exclude-namespaces = ["App\\Entity\\", "App\\DTO\\", "App\\ValueObject\\"]

# Methods where $this->property assignment is allowed
allowed-methods = ["__construct", "reset"]
```

#### Violating code

```php
namespace App\MessageHandler;

use Symfony\Component\Messenger\Attribute\AsMessageHandler;

#[AsMessageHandler(bus: 'command.bus')]
final class ProcessOrderHandler
{
    private array $processedIds = [];
    private int $count = 0;

    public function __construct(
        private readonly OrderRepository $orderRepository,
        private readonly CommandBus $commandBus,
    ) {
    }

    public function __invoke(ProcessOrderCommand $command): void
    {
        $order = $this->orderRepository->find($command->orderId);

        // WARNING: no-service-state-mutation
        // $this->processedIds will accumulate across ALL requests in worker mode
        $this->processedIds[] = $order->getId();

        // WARNING: no-service-state-mutation
        // $this->count will keep incrementing across requests
        $this->count++;

        $this->commandBus->dispatch(new FulfillOrderCommand($order->getId()));
    }
}
```

#### Fixed code (option 1: remove the state)

```php
namespace App\MessageHandler;

use Symfony\Component\Messenger\Attribute\AsMessageHandler;

#[AsMessageHandler(bus: 'command.bus')]
final class ProcessOrderHandler
{
    public function __construct(
        private readonly OrderRepository $orderRepository,
        private readonly CommandBus $commandBus,
    ) {
    }

    public function __invoke(ProcessOrderCommand $command): void
    {
        $order = $this->orderRepository->find($command->orderId);

        // If tracking is needed, use local variables or inject a dedicated service:
        // $this->logger->info('Processing order', ['id' => $order->getId()]);
        $this->commandBus->dispatch(new FulfillOrderCommand($order->getId()));
    }
}
```

#### Fixed code (option 2: implement ResetInterface)

```php
namespace App\MessageHandler;

use Symfony\Component\Messenger\Attribute\AsMessageHandler;
use Symfony\Contracts\Service\ResetInterface;

#[AsMessageHandler(bus: 'command.bus')]
final class ProcessOrderHandler implements ResetInterface
{
    private array $processedIds = [];
    private int $count = 0;

    public function __construct(
        private readonly OrderRepository $orderRepository,
        private readonly CommandBus $commandBus,
    ) {
    }

    public function __invoke(ProcessOrderCommand $command): void
    {
        $order = $this->orderRepository->find($command->orderId);

        $this->processedIds[] = $order->getId();
        $this->count++;

        $this->commandBus->dispatch(new FulfillOrderCommand($order->getId()));
    }

    public function reset(): void
    {
        $this->processedIds = [];
        $this->count = 0;
    }
}
```

### Why Mago

1. **Existing Symfony integration** — Mago already ships framework-specific rules for Symfony. This rule fits naturally alongside them.
2. **Complements `disallowed-functions`** — The `disallowed-functions` rule covers the function-level half of worker-mode safety (banning `ini_set`, `setlocale`, `header`, etc.). This rule would complete the picture by covering property-level state mutations.
3. **Configurable rule system** — TOML-based configuration with namespace inclusion/exclusion maps well to this use case.
4. **Inline suppression** — `@mago-expect` allows per-case exceptions with justification, which existing tools like Phanalist lack.
5. **Growing ecosystem need** — With worker mode becoming the default PHP deployment model, this rule would serve the broader PHP community.

### Willingness to Contribute

We are willing to contribute a PR for this rule or help with testing, specification, and real-world validation. We have a large Symfony 7.3 codebase (77+ bundles) that would serve as an excellent test bed.

Swahjak:
## Summary

Adds a new `no-service-state-mutation` lint rule that detects `$this->property` mutations inside service methods, which cause state leakage between requests in worker-mode PHP runtimes (FrankenPHP, RoadRunner, Swoole).

Closes #1579

## Rule Details

- **Name:** No Service State Mutation
- **Code:** `no-service-state-mutation`
- **Category:** Safety
- **Integration:** Symfony (only active when Symfony integration is enabled)
- **Level:** Warning (default)
- **Fixable:** No (requires architectural decision)

### Detection

Catches all forms of `$this->property` mutation outside allowed methods:

- Direct assignment: `$this->foo = $value`
- Compound assignment: `$this->foo .= $value`, `$this->foo += $value`
- Increment/decrement: `$this->foo++`, `--$this->count`
- Array push: `$this->items[] = $value`
- Array key assignment: `$this->items[$key] = $value`
- Unset: `unset($this->foo)` (reports all targets in multi-unset)

### Configuration

```toml
[rules.no-service-state-mutation]
level = "warning"
include-namespaces = ["App\\"]
exclude-namespaces = ["App\\Entity\\", "App\\DTO\\", "App\\ValueObject\\"]
allowed-methods = ["__construct", "reset"]
```

## Test plan

- [x] 18 unit tests (8 success cases, 10 failure cases)
- [x] Tests cover all mutation types, namespace filtering, allowed methods
- [x] `cargo test --workspace --locked --all-targets` passes
- [x] `cargo clippy -p mago-linter -- -D warnings` passes
- [x] `just check` passes (except pre-existing `mago analyze` issue on `composer/internal.php`)

### DIFF
--- a/crates/linter/src/rule/safety/no_service_state_mutation.rs
+++ b/crates/linter/src/rule/safety/no_service_state_mutation.rs
@@ -178,7 +178,7 @@ impl LintRule for NoServiceStateMutationRule {
                 let is_reset_class = class.implements.as_ref().is_some_and(|implements| {
                     implements.types.iter().any(|iface| {
                         let name = ctx.lookup_name(iface);
-                        self.cfg.reset_interfaces.iter().any(|ri| name == ri.as_str())
+                        self.cfg.reset_interfaces.iter().any(|ri| name.eq_ignore_ascii_case(ri.as_str()))
                     })
                 });
 


crates/linter/src/rule/safety/no_service_state_mutation.rs
```diff
@@ -0,0 +1,615 @@
...
+            ],
+            allowed_methods: vec!["__construct".to_string(), "reset".to_string()],
+        }
+    }
+}
+
+impl Config for NoServiceStateMutationConfig {
+    fn level(&self) -> Level {
+        self.level
+    }
```
azjezz, line 62:
```suggestion
impl Config for NoServiceStateMutationConfig {
    fn default_enabled() -> bool {
        false
    }

    fn level(&self) -> Level {
        self.level
    }
```

crates/linter/src/rule/safety/no_service_state_mutation.rs
```diff
@@ -0,0 +1,615 @@
...
+        // Check namespace filters.
+        // Patterns may end with a backslash (e.g. `App\Entity\`). A namespace like `App\Entity`
+        // should match `App\Entity\` because it is that namespace itself, not just a sub-namespace.
+        // We compare against both the raw namespace and namespace + `\` to cover both cases.
+        let namespace = ctx.scope.get_namespace();
+        if !namespace.is_empty() {
+            let namespace_with_sep = format!("{namespace}\\");
+            let matches_pattern =
+                |ns: &str, pattern: &str| ns.starts_with(pattern) || namespace_with_sep.starts_with(pattern);
+
```
azjezz: `namespace_with_sep` allocation is redundant. ( N files x M assignments = NxM allocations per `mago lint` ). we can just check if ns starts with `namespace` and that its length is at least `namespace.len() + 2`  and that the following char is `\\`.

crates/linter/src/rule/safety/no_service_state_mutation.rs
```diff
@@ -0,0 +1,615 @@
...
+                        .with_note(
+                            "In worker-mode runtimes (FrankenPHP, RoadRunner, Swoole), services persist across requests.",
+                        )
+                        .with_note("Mutating `$this->` properties causes shared state that leaks between requests.")
+                        .with_help("Use a local variable, a DTO, or a request-scoped service instead.");
+
+                    ctx.collector.report(issue);
+                }
+            }
+            return;
```
azjezz: nitpick

```suggestion
            }

            return;
```

crates/linter/src/rule/safety/no_service_state_mutation.rs
```diff
@@ -0,0 +1,615 @@
...
+    use crate::rule::safety::no_service_state_mutation::NoServiceStateMutationRule;
+    use crate::settings::Settings;
+    use crate::test_lint_failure;
+    use crate::test_lint_success;
+
+    fn symfony_settings(s: &mut Settings) {
+        s.integrations.insert(Integration::Symfony);
+    }
+
+    // ── Success cases ──────────────────────────────────────────────────
```
azjezz: please remove those comments.

azjezz (changes requested):
Overall, LGTM. 

Just few chnages. and you should run `just doc-linter-rules` to re-generate documentation :)

azjezz:
Please run `just fix` too to fix all clippy and formatting issues.

crates/linter/src/rule/safety/no_service_state_mutation.rs
```diff
@@ -154,10 +190,28 @@ impl LintRule for NoServiceStateMutationRule {
...
+
+            self.in_reset_class.store(is_reset_class, Ordering::Relaxed);
+            return;
+        }
+
+        // Skip mutations in classes implementing a reset interface.
+        if self.in_reset_class.load(Ordering::Relaxed) {
+            return;
+        }
+
```
azjezz: this is wrong. `Rule` *MUST* be stateless.

the correct approach for this rule:

1. target `Class` ( no assignment ).
2. create a `Walker` or `MutWalker` ( see other rules that do this )
3. walk down the class, until you walk into assignment, and do reporting there
4. add walk_ guards to not walk into nested classes/functions.

Swahjak: Done — rewrote the rule to target only `Class`/`Trait` and use a `MutationCollector` implementing `MutWalker` to walk method bodies. All mutable state (`AtomicBool`, `in_reset_class`) is removed. Walk guards prevent descending into anonymous classes, closures, and arrow functions.

crates/linter/src/rule/safety/no_service_state_mutation.rs
Swahjak: Done — added `default_enabled() -> false`.

crates/linter/src/rule/safety/no_service_state_mutation.rs
Swahjak: Done — replaced the `format!()` allocation with a length + char check.

crates/linter/src/rule/safety/no_service_state_mutation.rs
Swahjak: Fixed (the unset block was rewritten as part of the Walker refactor).

crates/linter/src/rule/safety/no_service_state_mutation.rs
Swahjak: Done — removed all section-header comments.

Swahjak:
Done — ran `just fix` and `just doc-linter-rules`.

Swahjak:
## Changes in this update

All review feedback has been addressed:

**Stateless rewrite (critical):** The rule no longer stores any mutable state. It targets only `Class`/`Trait` nodes and uses a `MutationCollector` struct implementing `MutWalker` to walk method bodies and detect mutations. Walk guards prevent descending into anonymous classes, closures, and arrow functions.

**Other fixes:**
- Added `default_enabled() -> false`
- Removed `format!()` allocation in namespace matching — replaced with length + char check
- Removed section-header comments from tests
- Ran `just fix` (clippy + formatting)
- Ran `just doc-linter-rules` to regenerate documentation

**New test:** Added a regression test verifying that property reads in array literal values (`'id' => $this->id`) are not flagged as mutations.

All 29 tests pass, clippy clean.

Swahjak:
One thing I noticed while testing: the `reset_interfaces` check currently only inspects the direct `implements` clause in the class AST. If a class extends a parent that implements `ResetInterface`, the rule will still flag it.

I see that the codex layer (`ClassLikeMetadata`) already tracks `all_parent_interfaces` (transitive), but `LintContext` doesn't have access to it.

Would you be open to wiring `CodebaseMetadata` (or a subset of it) into `LintContext` so linter rules can do transitive interface/hierarchy checks? Or would you prefer to keep the linter strictly per-file and accept this as a known limitation for now?

Happy to help with either approach.

azjezz:
>Would you be open to wiring CodebaseMetadata (or a subset of it) into LintContext

No. `codex` crate is only used by the analyzer. and no intentions in adding it to the linter.

The linter works by only looking at the syntax tree of 1 file at a time, it has no idea about other files.

This is different in the analyzer ( and codex ), where before analysis, we have a compilation phase, where we first scan every file to extract metadata from it, then merge all files into a single codebase metadata, then do analysis. 

In order to support `all_parent_interfaces`, we would need to do the same, i.e scan all files, build codebase metadata, then lint. and it would also require the linter to load vendor, all of this would simply kill the performance of the linter, probably making it 10-50 times slower, so no.

If a rule does require type information, or codebase metadata, we tend to move them to the analyzer.

However, i don't think that's needed here for no-service-state-mutation, while some false-positives might come from it, those are fine to suppress, without having to use full `mago analyze` which is slower than `mago lint`.

crates/linter/src/rule/safety/no_service_state_mutation.rs
```diff
@@ -0,0 +1,889 @@
...
+            if let Some(span) = is_property_mutation(value) {
+                self.findings.push(span);
+            }
+        }
+    }
+
+    // Don't descend into nested classes/functions — they have their own scope.
+    fn walk_anonymous_class(&mut self, _: &AnonymousClass<'arena>, _: &mut LintContext<'ctx, 'arena>) {}
+    fn walk_closure(&mut self, _: &Closure<'arena>, _: &mut LintContext<'ctx, 'arena>) {}
+    fn walk_arrow_function(&mut self, _: &ArrowFunction<'arena>, _: &mut LintContext<'ctx, 'arena>) {}
```
azjezz, line 281: must also not descend into inner classes. e.g:

```php
class A {
   public function x(): void {
      class B {
        private string $x = '';

        public function doSomething(): void {
           $this->x .= 'a';
        }
      }
   }
}
```

`$this->x` mutation currently would be reported twice.

Must skip it here ( also applies to nested traits )

Swahjak: Good catch — fixed in 4398a12f by also overriding `walk_class`/`walk_trait`/`walk_interface`/`walk_enum` in `MutationCollector`, plus a regression test (`mutation_in_nested_class_reported_once`) asserting the inner mutation is reported exactly once.

crates/linter/src/rule/safety/no_service_state_mutation.rs
```diff
@@ -0,0 +1,921 @@
...
+        Self { meta: Self::meta(), cfg: settings.config.clone() }
+    }
+
+    fn check<'arena>(&self, ctx: &mut LintContext<'_, 'arena>, node: Node<'_, 'arena>) {
+        let members = match node {
+            Node::Class(class) => {
+                let is_reset_class = class.implements.as_ref().is_some_and(|implements| {
+                    implements.types.iter().any(|iface| {
+                        let name = ctx.lookup_name(iface);
+                        self.cfg.reset_interfaces.iter().any(|ri| name == ri.as_str())
```
azjezz: interface names are case insensative, we should do `eq_ignore_ascii_case` here.

Swahjak:
Good catch on the case-insensitive interface comparison — you've had a sharp eye throughout this review and caught several important improvements. Fixed in b1e9ba09.

PR #1606, opened 2026-04-08
perf(syntax): avoid per-node `Vec` allocation when traversing the AST

dotdash:
## 📌 What Does This PR Do?

Node::children() allocates a Vec<Node> on every call. The linter walk and filter_map_internal call it for every node in the tree, so a file with N AST nodes triggers N heap allocations just for the traversal scaffolding, with matching frees and occasional memmove when vecs grow.

Add visit_children, which drives the same match logic but delivers each child to a caller-supplied FnMut rather than collecting into a Vec. children() becomes a thin wrapper that collects from visit_children, preserving all existing callers unchanged.

Switch mago_linter::walk, filter_map_internal, and the recursive metric rules (cyclomatic complexity, Halstead, Kan defect) to visit_children so none of those traversal paths allocate a Vec per node.

## 🛠️ Summary of Changes

- **Bug Fix:** Improved linter performance by eliminating unnecessary allocations.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

noppidy

## 📝 Notes for Reviewers

In tempest, with cyclo, halstead and kan enabled:

```
Summary
  ~/mago-new --threads 32 lint --retain-code none ran
    1.17 ± 0.04 times faster than ~/mago-base --threads 32 lint --retain-code none
```

With the default config that comes with tempest:
```
Summary
  ~/mago-new --threads 32 lint --retain-code none ran
    1.14 ± 0.02 times faster than ~/mago-base --threads 32 lint --retain-code none
```

### DIFF
--- a/crates/linter/src/lib.rs
+++ b/crates/linter/src/lib.rs
@@ -129,40 +129,54 @@ fn is_constant_expression_context(kind: NodeKind) -> bool {
     )
 }
 
-fn walk<'arena>(node: Node<'_, 'arena>, ctx: &mut LintContext<'_, 'arena>, excluded_rules: &HashSet<usize>) {
-    let mut in_scope = false;
-    if let Some(scope) = Scope::for_node(ctx, node) {
-        ctx.scope.push(scope);
-
-        in_scope = true;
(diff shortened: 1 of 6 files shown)


dotdash:
Mhm... when did I break that...?

PR #1620, opened 2026-04-12
feat(formatter): add opt-in arithmetic and bitwise binary paren omission

psihius:
## 📌 What Does This PR Do?

Adds an opt-in formatter setting for omitting redundant parentheses around arithmetic binary expressions under comparison and null coalesce expressions.

With `omit-redundant-arithmetic-binary-expression-parentheses = true`:

```php
if ($i === ($retries - 1)) {
}
$value = $maybe ?? ($fallback + 1);
```

formats as:

```php
if ($i === $retries - 1) {
}
$value = $maybe ?? $fallback + 1;
```

The default remains unchanged, so Mago continues preserving its current parenthesized output unless the setting is explicitly enabled.

## 🔍 Context & Motivation

Mago currently has no formatter option for projects that prefer arithmetic operands to remain ungrouped when PHP precedence already preserves the same meaning. This adds a targeted compatibility mode without changing the default formatter behavior or existing presets.

## 🛠️ Summary of Changes

- **Feature:** Added `omit-redundant-arithmetic-binary-expression-parentheses`.
- **Formatter:** Updated binary-expression parenthesis handling for arithmetic children under comparison and null coalesce parents when the setting is enabled.
- **Presets:** Added the setting to existing presets with `false` to preserve current behavior.
- **Tests:** Added additive formatter fixtures for enabled and disabled modes, including mixed already-formatted and cleanup-needed cases.
- **Docs:** Documented the new formatter configuration option.
- **Config:** Added TOML deserialization coverage for the new setting.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [x] Documentation
- [x] Other (please specify): formatter configuration parsing

## 🔗 Related Issues or PRs

N/A

## 📝 Notes for Reviewers

Default formatter behavior is unchanged. The new behavior only applies when `omit-redundant-arithmetic-binary-expression-parentheses` is enabled.

Verification run:

- `cargo test -p mago-formatter binary_precedence_parens_noise -- --exact`
- `cargo test -p mago-formatter binary_precedence_parens_noise_disabled -- --exact`
- `cargo test -p mago omit_redundant`
- `just test`
- `just check`
- `cargo fmt --all --check`

Also validated against a larger local corpus for side effects.


crates/formatter/src/internal/parens.rs
```diff
@@ -290,8 +290,16 @@ impl<'arena> FormatterState<'_, 'arena> {
...
+
+                if (operator.is_arithmetic()
+                    && !e.operator.is_arithmetic()
+                    && !parent_omits_redundant_arithmetic_parentheses)
+                    || (operator.is_arithmetic()
+                        && e.operator.is_arithmetic()
+                        && (operator.is_multiplicative() || e.operator.is_multiplicative()))
                     || (operator.is_bit_shift() && !e.operator.is_bit_shift())
                     || (operator.is_bitwise() && e.operator.is_bitwise() && !e.operator.is_same_as(operator))
                 {
```
azjezz: this seems more complex than it should be..

Also, we should probably add same option for `bitwise` - as this path is taken for 2 reason, either arithmetic, or bitwise, if we are adding an option for omitting arithmetic parens, we should do the same for bitwise.

psihius: Included a toggle and tests for bitwise parentes handling,

Refactored the code to make it more readable

PR #1642, opened 2026-04-15
feat(formatter): align assignment-like match arms

psihius:
## 📌 What Does This PR Do?

Adds match arm support to the existing `align-assignment-like` formatter option and fixes a few assignment-alignment edge cases for consecutive statements.

## 🔍 Context & Motivation

`align-assignment-like` already aligns variable assignments, array key-value pairs, class properties, constants, and enum cases. Match arms are another assignment-like construct, but were not covered by the option.

Statement alignment also missed array append assignments and could apply outer alignment padding inside nested assignment chains or continue a run past an assignment whose right-hand side already spans multiple lines.

## 🛠️ Summary of Changes

- **Feature:** Aligns consecutive match arms when `align-assignment-like` is enabled.
- **Bug Fix:** Treats array append assignments as alignable statement assignments.
- **Bug Fix:** Prevents statement alignment padding from leaking into nested right-hand-side assignments.
- **Bug Fix:** Ends statement alignment runs after assignments with multiline right-hand sides.
- **Docs:** Updates the formatter configuration reference for `align-assignment-like`.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [x] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

None.

## 📝 Notes for Reviewers

Verification:

- `cargo test -p mago-formatter`
- `cargo fmt --all --check`


azjezz (changes requested):
1. array append change should be its own PR, that is a bug fix.
2. match alignment should be a new option, disabled by default, reducing friction when updating.

PR #1673, opened 2026-04-19
fix(formatter): avoid inlining nested single-element arrays

### LINKED ISSUE #1672: Formatter incorrectly inlines single-element array-like arguments with nested arrays
### Description

I'm not sure exactly how to put it to words, but basically the formatter doesn't expand correctly arrays and tries to keep the closing `]` on the same line, even though it should most likely be placed on it's own new line. This would match the behavior of other formatters.

### Playground Link

_No response_

### Reproduction

```php
<?php
class TestCase
{
    public function test()
    {
        $fixtureFiles = [];
        $this->logger->info('Loading a total of {count} fixture files', ['count' => count($fixtureFiles)]);
    }
}
```

Using the following configuration:

```toml
[formatter]
print-width = 100
```

Will produce the following diff:

```diff
-        $this->logger->info('Loading a total of {count} fixture files', ['count' => count($fixtureFiles)]);
+        $this->logger->info('Loading a total of {count} fixture files', ['count' =>
+            count($fixtureFiles)]);
```

I think it should instead produce something like this:
```diff
-        $this->logger->info('Loading a total of {count} fixture files', ['count' => count($fixtureFiles)]);
+        $this->logger->info('Loading a total of {count} fixture files', [
+            'count' => count($fixtureFiles)
+        ]);
```

### Environment

- Mago 1.23.1

Zuruuh:
## 📌 What Does This PR Do?

Fix #1672, make sure we correctly expand arrays on multiple lines.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

I've tried to make sure it wouldn't change any of the existing test cases, but I've had to change the `drupal_preset` and the `preserve_breaking_array_like_expand_last` test cases because they were impacted by this issue exactly. I don't know if it's fine for the second one to change because from what I understand it should maybe "preserve" the existing format? But from what I see the formatting settings of the test cases are the default so no `preserve_*` option so idk really

### DIFF
--- a/crates/formatter/src/internal/format/array.rs
+++ b/crates/formatter/src/internal/format/array.rs
@@ -29,6 +29,7 @@ use crate::internal::format::misc::get_document_width;
 use crate::internal::format::misc::is_expandable_expression;
 use crate::internal::format::misc::is_string_word_type;
 use crate::internal::utils::get_expression_width;
+use crate::internal::utils::string_width;
 
 #[derive(Debug, Clone, Copy)]
 pub enum ArrayLike<'arena> {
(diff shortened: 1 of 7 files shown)


azjezz:
Thanks @Zuruuh !

PR #1685, opened 2026-04-21
feat(guard): allow Guard structural autofix

yoannblot:
## 📌 What Does This PR Do?

Add Guard structural autofix capacity.

## 🔍 Context & Motivation

When configuring a project with structural rules, it feels weird not to be able to fix them instead of just checking invalid ones.
Adding `final` or `readonly` keywords on a class is something simple, I'd rather my mago tool do it instead of manually adding it.

This PR introduces a new `guard --fix` command which will do it for us (not the default behavior).

## 🛠️ Summary of Changes

- **Feature:** Introduced `--fix` flag to the `guard` command to automatically resolve structural violations.
- **Feature:** Implemented autofix support for several structural rules:
    - `must-be-final` (true/false)
    - `must-be-readonly` (true/false)
    - `must-implement` (Single and AllOf constraints)
    - `must-extend` (when the class doesn't already have a parent)
    - `must-use-trait` (Single and AllOf constraints)
- **Feature:** Added `--dry-run` and `--format-after-fix` options to the `guard` command.
- **Internal:** Updated `StructuralFlaw` to include suggested `TextEdit`s.
- **Documentation:** Added comprehensive usage guide and command reference for the new fix capabilities.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [x] CLI
- [ ] Dependencies
- [x] Documentation
- [x] Other (please specify): Guard Core (Structural Analysis)

## 📝 Notes for Reviewers

- Fixes are applied using fully qualified names (FQN) with a leading backslash to ensure correctness regardless of the namespace context.
- Some rules like `must-be-abstract` or `must-be-named` are intentionally not auto-fixed as they require manual architectural decisions.
- The `--format-after-fix` flag is recommended to ensure the inserted code matches the project's coding style.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1529,6 +1529,7 @@ dependencies = [
  "mago-span",
  "mago-syntax",
  "mago-syntax-core",
+ "mago-text-edit",
  "schemars",
  "serde",
  "toml",
--- a/crates/guard/Cargo.toml
+++ b/crates/guard/Cargo.toml
@@ -22,6 +22,7 @@ mago-collector = { workspace = true }
 mago-syntax-core = { workspace = true }
 mago-database = { workspace = true }
 mago-codex = { workspace = true }
+mago-text-edit = { workspace = true }
 foldhash = { workspace = true }
 serde = { workspace = true }
 bumpalo = { workspace = true }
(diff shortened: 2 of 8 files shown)


yoannblot:
Hi @azjezz! Wdyt about this suggestion?

PR #1692, opened 2026-04-22
feat(prelude): add the memcached extension

UweOhse:
## 📌 What Does This PR Do?

This change adds a stub for the Memcached class.

## 🔍 Context & Motivation

This is the one extension i use for which no stub exists, and i want to be able to check the code using that.

## 🛠️ Summary of Changes

It only adds the crates/prelude/assets/extensions/memcached.php file.

## 📂 Affected Areas

- [x ] Other (please specify): crates/prelude

## 🔗 Related Issues or PRs

none.

## 📝 Notes for Reviewers

The PHP memcached (https://github.com/php-memcached-dev/php-memcached) extension is, well, kind of messy.

It renamed some libmemcached options, and i had problems hunting down some of the constants.

Its own stub file (memcached-api.php) is a bit incomplete (it doesn't include the COMPRESSION_ZSTD and OPT_SERIALIZER options). `grep REGISTER_MEMC_  memcached-api.php` for a look at the inconsistent renaming.

I wouldn't be surprised if some constants are wrong, but i did check them twice.
I wouldn't be surprised at all if i missed some option.

An especially messy area are the bool options, which are defined in this way:
```
   /** @var bool */
    public const bool HAVE_JSON = false;
    /** @var bool */
    public const bool HAVE_IGBINARY = false;
    /** @var bool */
    public const bool HAVE_MSGPACK = false;
    /** @var bool */
    public const bool HAVE_ENCODING = false;

    /**
     * Feature support
     */
    /** @var bool */
    public const bool HAVE_SESSION = false;
    /** @var bool */
    public const bool HAVE_SASL = false;
```
The value of these consts depends on the situation at the time the php-memcached extension is compiled, so these are constants with inconstant values.
I opted for the quite dirty `@var bool` solution to enable code like `if (Memcache::HAVE_IGBINARY) { use the igbinary serializer }`  but maybe that's not the best solution?

The `GET_ERROR_RETURN_VALUE` bool const indeed is a const.

`just test` ran successfully.

`just check` failed at the composer stage.
```
error[non-existent-use-import]: Imported class, interface, trait, or enum `Composer\InstalledVersions` does not exist.
  ┌─ composer/internal.php:7:5
  │
7 │ use Composer\InstalledVersions;
  │     ^^^^^^^^^^^^^^^^^^^^^^^^^^ `Composer\InstalledVersions` not found
  │
  = Help: Ensure that the class, interface, trait, or enum `Composer\InstalledVersions` is defined in the codebase.
```
I don't think that this can be related to my change. Maybe it's the debian composer (2.8.8)?

I understand if you decline this change.

### DIFF
--- a/crates/prelude/assets/extensions/memcached.php
+++ b/crates/prelude/assets/extensions/memcached.php
@@ -2,9 +2,7 @@
 
 class Memcached
 {
-    public const int OPT_SERIALIZER = -1003; // taken from the PHP website, because memcached-api.php doesn't have it. WTF?
-
-    // taken from php-memcached / memcached-api.php
+    public const int OPT_SERIALIZER = -1003;
 
     public const int OPT_HASH = 2;
     public const int HASH_DEFAULT = 0;
@@ -48,9 +46,6 @@ class Memcached
     public const int OPT_RANDOMIZE_REPLICA_READS = 30;
     public const int OPT_REMOVE_FAILED_SERVERS = 35;
 
-    /**
-     * Supported serializers
-     */
     /** @var bool */
     public const bool HAVE_JSON = false;
     /** @var bool */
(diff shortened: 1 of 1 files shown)


azjezz (changes requested):
Please remove all comments in the stub that are not providing type information, as those are not benifitial.

PR #1724, opened 2026-04-26
refactor(codex): use Vec instead of HashMap for template_variance

dotdash:
## 📌 What Does This PR Do?

Replace `HashMap<usize, Variance>` with `Vec<Variance>` for the
`template_variance` field in `ClassLikeMetadata`.

## 🔍 Context & Motivation
Template parameter indices are always 0..N (positions in the
`template_types` IndexMap), so the keys are never sparse. Using a
`HashMap` allocates more than necessary and requires hashing for what
are effectively sequential slot lookups.

## 🛠️ Summary of Changes

- **Refactor:** Replaced `HashMap<usize, Variance>` with `Vec<Variance>`
  in `ClassLikeMetadata` and updated all affected call sites.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [x] Other: codex / type system internals

## 🔗 Related Issues or PRs

noppidy

## 📝 Notes for Reviewers

:train:

### DIFF
--- a/crates/codex/src/metadata/class_like.rs
+++ b/crates/codex/src/metadata/class_like.rs
@@ -1,4 +1,3 @@
-use foldhash::HashMap;
 use foldhash::fast::RandomState;
 use indexmap::IndexMap;
 use serde::Deserialize;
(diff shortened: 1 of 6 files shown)


crates/codex/src/metadata/class_like.rs
```diff
@@ -307,7 +306,13 @@ impl ClassLikeMetadata {
     /// Adds or updates the variance for a specific parameter index. Returns the previous variance if one existed.
     #[inline]
     pub fn add_template_variance_parameter(&mut self, index: usize, variance: Variance) -> Option<Variance> {
-        self.template_variance.insert(index, variance)
+        if index < self.template_variance.len() {
+            Some(std::mem::replace(&mut self.template_variance[index], variance))
+        } else {
+            self.template_variance.resize(index, Variance::Invariant);
```
azjezz: the default should be `Covariant`, no? at least in type comparison that is true.

PR #1746, opened 2026-04-30
feat(linter/safety): add glob pattern support to no-service-state-mutation namespace filters

### LINKED ISSUE #1745: feat(linter/safety): support glob patterns in `no-service-state-mutation` namespace filters
## Use case

In a typical Symfony project, classes like Entities, DTOs, and Value Objects live under namespaces such as `App\Entity`, `App\Dto`, or `App\ValueObject`. These classes are **data containers**, not services — they are designed to carry and mutate state, so property assignments in their methods are completely intentional and should not be flagged by the `no-service-state-mutation` rule.

The current `exclude-namespaces` option supports simple **prefix matching** only. This works for straightforward cases, but breaks down when a project organises these classes more granularly — for example:

```
App\Entity\User\          ← should be excluded
App\Entity\Order\         ← should be excluded
App\Dto\Request\          ← should be excluded
App\Dto\Response\         ← should be excluded
App\Dto\Internal\         ← should be excluded
```

With prefix matching you either exclude too broadly (`App\`) or have to enumerate every leaf namespace explicitly, which is fragile and tedious to maintain.

## Proposed solution

Allow **glob patterns** in `exclude-namespaces` (and symmetrically in `include-namespaces`) so users can write expressive, concise exclusion rules:

```toml
[linter.rules.no-service-state-mutation]
exclude-namespaces = [
  "App\\Entity\\**",
  "App\\Dto\\**",
  "App\\ValueObject\\**",
]
```

The `**` wildcard should match any number of namespace segments (like the guard crate already does for perimeter rules), and `*` should match exactly one segment. Brace expansion (`{Entity,Dto}`) would also be welcome for power users.

## Context

The `guard` crate already ships a full-featured PHP FQN pattern matcher (`crates/guard/src/matcher.rs`) that handles `*`, `**`, partial wildcards, and brace expansion against `\`-separated namespace strings. Reusing or sharing that logic for the linter's namespace filters would keep semantics consistent across the tool.

Related: #1582

Swahjak:
## Summary

Adds glob pattern support to `include-namespaces` and `exclude-namespaces` in the `no-service-state-mutation` linter rule.

**Motivation:** In Symfony projects, data-container classes (Entities, DTOs, Value Objects) live under namespaces like `App\Entity` or `App\Dto` and naturally mutate state — they should be excluded from this rule. The previous prefix-only matching works for flat namespace structures but breaks down for deeper hierarchies. A project might need to exclude `App\Entity\User\`, `App\Entity\Order\`, etc., which required enumerating every sub-namespace explicitly.

**What changed:**

- New `crates/linter/src/rule/utils/namespace_pattern.rs` — a namespace FQN pattern matcher extracted from the guard crate's `matcher.rs`, supporting `*`, `**`, partial wildcards, and brace expansion. Placed in `rule/utils/` so other linter rules can reuse it.
- `no_service_state_mutation.rs` — replaces the old `starts_with`-based closure with `matches_namespace_pattern(..., false, true)`. Backward-compatible: existing patterns like `App\Entity\` continue to match as namespace prefixes.
- `RuleMeta` description updated to document the new pattern syntax.
- 4 new test cases covering `**`, `*`, brace expansion, and the new `include-namespaces` wildcard path.
- Docs regenerated via `just doc-linter-rules`.

**Example config:**
```toml
[linter.rules.no-service-state-mutation]
exclude-namespaces = [
  "App\\{Entity,DTO,ValueObject}\\",
  "App\\Domain\\**",
]
```

## Test plan

- [x] `cargo test -p mago-linter no_service_state_mutation` — 34 tests pass (30 existing + 4 new)
- [x] `just fix` — no clippy warnings, formatting clean
- [x] `just doc-linter-rules` — docs regenerated, glob paragraph visible in `docs/tools/linter/rules/safety.md`

Closes #1745

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 .mago/
 /corpus/vendor/
 /corpus/composer.lock
+.worktrees
--- a/crates/linter/src/rule/utils/namespace_pattern.rs
+++ b/crates/linter/src/rule/utils/namespace_pattern.rs
@@ -1,0 +1,199 @@
+use mago_atom::starts_with_ignore_case;
+
+const SEPARATOR: char = '\\';
+
+pub fn matches_namespace_pattern(fqcn: &str, pattern: &str, is_constant: bool, treat_as_namespace: bool) -> bool {
+    if pattern.contains('{') {
+        return expand_braces(pattern)
+            .iter()
+            .any(|p| matches_namespace_pattern(fqcn, p, is_constant, treat_as_namespace));
+    }
+
+    if !pattern.contains('*') {
+        let p = pattern.trim_matches(SEPARATOR);
+        let f = fqcn.trim_matches(SEPARATOR);
+
+        if p.is_empty() {
+            return f.is_empty();
+        }
+
+        if treat_as_namespace || pattern.ends_with(SEPARATOR) {
+            if !starts_with_ignore_case(f, p) {
+                return false;
+            }
+
+            return f.len() == p.len() || f.as_bytes().get(p.len()) == Some(&(SEPARATOR as u8));
+        }
+
+        if is_constant {
+            let f_last = f.rsplit_once(SEPARATOR);
+            let p_last = p.rsplit_once(SEPARATOR);
+
+            return match (f_last, p_last) {
+                (Some((f_ns, f_name)), Some((p_ns, p_name))) => f_ns.eq_ignore_ascii_case(p_ns) && f_name == p_name,
+                (None, None) => f == p,
+                _ => false,
+            };
+        }
+        return f.eq_ignore_ascii_case(p);
+    }
+
+    let fqcn = fqcn.trim_matches(SEPARATOR);
+    let pattern = pattern.trim_matches(SEPARATOR);
+
+    if pattern == "**" {
+        return true;
+    }
+
+    if pattern == "*" {
+        return !fqcn.contains(SEPARATOR);
+    }
+
+    if fqcn.is_empty() || pattern.is_empty() {
+        return fqcn == pattern;
+    }
+
+    let fqcn_parts: Vec<&str> = fqcn.split(SEPARATOR).collect();
+    let pattern_parts: Vec<&str> = pattern.split(SEPARATOR).collect();
+
+    do_match(&fqcn_parts, &pattern_parts, is_constant)
+}
+
+fn do_match(fqcn_parts: &[&str], pattern_parts: &[&str], is_constant: bool) -> bool {
+    match (fqcn_parts.first(), pattern_parts.first()) {
+        (None, None) => true,
+        (_, Some(&"**")) => {
+            if pattern_parts.len() == 1 {
+                return true;
+            }
+            if fqcn_parts.is_empty() {
+                return do_match(fqcn_parts, &pattern_parts[1..], is_constant);
+            }
+            do_match(fqcn_parts, &pattern_parts[1..], is_constant)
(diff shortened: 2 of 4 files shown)


azjezz:
@Swahjak please rebase. note that you don't need to generate linte documentation anymore. that is done automatically.

Swahjak:
@azjezz done

crates/linter/src/rule/utils/namespace_pattern.rs
```diff
@@ -0,0 +1,199 @@
+use mago_atom::starts_with_ignore_case;
+
+const SEPARATOR: char = '\\';
+
+pub fn matches_namespace_pattern(fqcn: &str, pattern: &str, is_constant: bool, treat_as_namespace: bool) -> bool {
```
azjezz, line 5: `is_constant` is a dead parameter, not used anywhere, and always set to `false`.

crates/linter/src/rule/utils/namespace_pattern.rs
```diff
@@ -0,0 +1,199 @@
...
+    parts.push(&alternatives[start..]);
+
+    let mut results = Vec::with_capacity(parts.len());
+    for part in parts {
+        let expanded = format!("{prefix}{part}{suffix}");
+        results.extend(expand_braces(&expanded));
+    }
+
+    results
+}
```
azjezz, line 199: There is no tests for this matcher at all. while tested via the linter rule, it should also have its own tests because this is a big chunk of code.

crates/linter/src/rule/utils/namespace_pattern.rs
```diff
@@ -0,0 +1,199 @@
+use mago_atom::starts_with_ignore_case;
+
+const SEPARATOR: char = '\\';
+
+pub fn matches_namespace_pattern(fqcn: &str, pattern: &str, is_constant: bool, treat_as_namespace: bool) -> bool {
+    if pattern.contains('{') {
```
azjezz, line 6: `App\{Entity,DTO\` pattern ( missing `}` ) falls to `expand_braces`, but there is no closing brace here.

.gitignore
```diff
@@ -8,3 +8,4 @@
 .mago/
 /corpus/vendor/
 /corpus/composer.lock
+.worktrees
```
azjezz, line 11: Should be removed.

azjezz (changes requested):
I'm not sure this addition is good.

Unlike guard, which compiles regular expressiosn once per invocation, this rule now has its own handrolled matcher which has some issues, but also, is expensive to invoke for every single class and trait within a codebase.

brace expansion is not something we do in guard either.

Overall, i feel like this rule is becoming too big for little benefit.

azjezz:
thanks for the rebase.

after looking more closely i'm going to close this. three of the four new tests work today with the existing prefix matcher, `App\Domain\**` matches the same set as `App\Domain\`, and `App\{Entity,DTO,VO}\` is three entries in an array. the only genuinely new capability is mid-pattern wildcards like `App\*\Service`, and i don't think one use case justifies a new pattern dialect that every namespace-filtered rule will end up being asked for.

your motivating example (`App\Entity\User\`, `App\Entity\Order\`) already works with `App\Entity\` as a prefix.

if mid-pattern wildcards turn out to be a real need across multiple rules, the right move is consolidating with guard's matcher into a shared crate, not a parallel handrolled one.

happy to design that in an issue or in discord if you have a case prefixes can't cover.

PR #1779, opened 2026-05-04
Type patching

### LINKED ISSUE #1288: Support for patching types in vendor code
### Feature

Sometimes type annotation in vendor code aren't correct or not expressed in a way that allows proper static analysis. I'd love to be able to patch the type in these cases. With PHPStan we can do that with a stub file which overrides the typings on just the given items, rather than replacing e.g. the whole class.

### Example

From Yii 1.1 (don't judge me :joy:)

```php
abstract class CApplication {
	/**
	 * Terminates the application.
	 * This method replaces PHP's exit() function by calling
	 * {@link onEndRequest} before exiting.
	 * @param integer $status exit status (value 0 means normal exit while other values mean abnormal exit).
	 * @param boolean $exit whether to exit the current request. This parameter has been available since version 1.1.5.
	 * It defaults to true, meaning the PHP's exit() function will be called at the end of this method.
	 */
	public function end($status=0,$exit=true) { /* ... */ }
```

Since there is no `never` return type, mago will complain about functions in the application code that are supposed to return `never` and just call this `end()` method.

With PHPStan we have a stub:

```php
class CApplication
{
    /**
     * @return ($exit is true ? never : void)
     */
    public function end(int $status=0, bool $exit=true);
}
```

And this just patches the `end()` method, without affecting the rest of the class.

dotdash:
## 🛠️ Summary of Changes

- **Refactor:** Improved `mago_ast` structure.
- **Docs:** Updated installation guide.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #1288

### DIFF
--- a/crates/codex/src/metadata/function_like.rs
+++ b/crates/codex/src/metadata/function_like.rs
@@ -6,10 +6,12 @@ use serde::Serialize;
 use mago_atom::Atom;
 use mago_atom::AtomMap;
 use mago_atom::AtomSet;
+use mago_reporting::Annotation;
 use mago_reporting::Issue;
 use mago_span::Span;
 
 use crate::assertion::Assertion;
+use crate::issue::ScanningIssueKind;
 use crate::metadata::attribute::AttributeMetadata;
 use crate::metadata::class_like::TemplateTypes;
 use crate::metadata::flags::MetadataFlags;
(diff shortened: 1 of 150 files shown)


crates/codex/src/metadata/function_like.rs
```diff
@@ -292,4 +294,104 @@ impl FunctionLikeMetadata {
...
+            let static_mismatch = patch_m.is_static != self_m.is_static;
+            let final_removed = self_m.is_final && !patch_m.is_final;
+
+            if visibility_mismatch || abstract_mismatch || static_mismatch || final_removed {
+                let name = patch.original_name.as_deref().or(patch.name.as_deref()).unwrap_or("<anonymous>");
+                diagnostics.push(
+                    Issue::warning(format!(
+                        "Patch for `{}` declares structural attributes (visibility, abstract, static, or \
+                         removing final) that differ from the original; only type annotations are applied.",
+                        name,
```
azjezz, line 349: error messages should be shorter, to stay in a single line. we can use `notes` and `help` for additional info.

crates/codex/src/ttype/comparator/mod.rs
```diff
@@ -88,9 +88,7 @@ mod tests {
             let program = parse_file(&arena, &file);
             assert!(!program.has_errors(), "Parse failed: {:?}", program.errors);
             let resolved_names = NameResolver::new(&arena).resolve(program);
-            let program_codebase = scan_program(&arena, &file, program, &resolved_names);
```
azjezz, line 91: unrelated

crates/codex/src/ttype/expander.rs
```diff
@@ -1108,9 +1108,7 @@ mod tests {
             let program = parse_file(&arena, &file);
             assert!(!program.has_errors(), "Parse failed: {:?}", program.errors);
             let resolved_names = NameResolver::new(&arena).resolve(program);
-            let program_codebase = scan_program(&arena, &file, program, &resolved_names);
-
-            codebase.extend(program_codebase);
+            codebase.extend(scan_program(&arena, &file, program, &resolved_names));
```
azjezz, line 1111: unrelated

crates/codex/tests/comparator_common/mod.rs
```diff
@@ -59,9 +59,8 @@ pub fn codebase_from_php(code: &'static str) -> CodebaseMetadata {
         let program = parse_file(&arena, &file);
         assert!(!program.has_errors(), "Parse failed: {:?}", program.errors);
         let resolved_names = NameResolver::new(&arena).resolve(program);
-        let program_codebase = scan_program(&arena, &file, program, &resolved_names);
 
-        codebase.extend(program_codebase);
+        codebase.extend(scan_program(&arena, &file, program, &resolved_names));
     }
```
azjezz, line 64: unrelated

dotdash:
Whoopsie

PR #1785, opened 2026-05-05
feat(linter): add deny mode to yoda-conditions rule

dotdash:
## 📌 What Does This PR Do?

Extends the `yoda-conditions` rule with a `deny` mode that forbids                                                                                                                                                                                                             
Yoda style, complementing the existing `require` mode.

## 🔍 Context & Motivation

The rule currently only enforces Yoda style (`require` mode). Some                                                                                                                                                                                                             
codebases prefer the opposite — variable on the left for readability                                                                                                                                                                                                           
— and have no way to mechanically enforce that preference. A `deny`                                                                                                                                                                                                            
mode fills this gap.                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                 
When Yoda style is forbidden, the accidental-assignment protection it                                                                                                                                                                                                          
provides (a mistyped `=` causes a fatal error in Yoda form but a                                                                                                                                                                                                             
silent bug in non-Yoda form) is lost. The deny mode therefore suggests                                                                                                                                                                                                         
enabling the `no-assign-in-condition` rule as a replacement safeguard.                                                                                                                                                                                                         

The default remains `require`, preserving existing behavior.

## 🛠️ Summary of Changes

- **Feature:** Added `YodaConditionsMode` enum (`require`/`deny`) and a `mode` config field to `yoda-conditions`.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

nada

## 📝 Notes for Reviewers

In `deny` mode the auto-fix is the same swap as in `require` mode —                                                                                                                                                                                                            
the symmetry is intentional: both modes simply swap the two operands
to produce the correct form.

### DIFF
--- a/crates/linter/src/rule/best_practices/yoda_conditions.rs
+++ b/crates/linter/src/rule/best_practices/yoda_conditions.rs
@@ -29,15 +29,26 @@ pub struct YodaConditionsRule {
     cfg: YodaConditionsConfig,
 }
 
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq, Hash, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "kebab-case")]
+pub enum YodaConditionsMode {
+    /// Require Yoda style: constant/literal on the left, variable on the right.
+    #[default]
+    Require,
+    /// Deny Yoda style: variable on the left, constant/literal on the right.
+    Deny,
+}
+
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash, Serialize, Deserialize, JsonSchema)]
 #[serde(default, rename_all = "kebab-case", deny_unknown_fields)]
 pub struct YodaConditionsConfig {
     pub level: Level,
+    pub mode: YodaConditionsMode,
 }
 
 impl Default for YodaConditionsConfig {
     fn default() -> Self {
-        Self { level: Level::Help }
+        Self { level: Level::Help, mode: YodaConditionsMode::Require }
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/linter/src/rule/best_practices/yoda_conditions.rs
```diff
@@ -112,36 +130,46 @@ impl LintRule for YodaConditionsRule {
...
+                )),
+            YodaConditionsMode::Deny => (is_constant_like(binary.lhs) && is_writable_variable(binary.rhs)).then_some((
+                "Avoid Yoda condition style",
+                "Constant should be on the right side",
+                "Yoda conditions are harder to read; prefer `$count === 5`.",
+                "Move variable to left: `$count === 5`",
+            )),
+        }) else {
+            return;
+        };
```
azjezz: Took me a minute to understand this :D could be made more readable:

```suggestion
        let (message, annotation, reason, help) = match self.cfg.mode {
            YodaConditionsMode::Require if is_writable_variable(binary.lhs) && is_constant_like(binary.rhs) => (
                "Use Yoda condition style for safer comparisons",
                "Variable should be on the right side",
                "Yoda conditions help prevent accidental assignment bugs where `=` is used instead of `==`",
                "Move constant/literal to left: `5 === $count`",
            ),
            YodaConditionsMode::Deny if is_constant_like(binary.lhs) && is_writable_variable(binary.rhs) => (
                "Avoid Yoda condition style",
                "Constant should be on the right side",
                "Yoda conditions are harder to read; prefer `$count === 5`.",
                "Move variable to left: `$count === 5`",
            ),
            _ => { return; }
        };
```

azjezz (changes requested):

azjezz (approved):

PR #1805, opened 2026-05-07
feat(testing): introduce standalone integration test runner

dotdash:
## 📌 What Does This PR Do?

The runner lives in crates/testing/ and operates on a root-level cases/ directory. Each case is a subdirectory containing:

  source.php    — the PHP input
  results.toml  — expected issues, keyed by tool and code
  mago.toml     — required per-case tool configuration

The runner panics early if mago.toml is absent, since without it mago has no source paths and the test would silently pass empty.

results.toml uses [analyzer] and [linter] sections that map issue codes to sorted arrays of 1-based line numbers. Each tool is run twice per case to verify output stability. Snapshots can be regenerated with MAGO_UPDATE_SNAPSHOTS=1.

A defaults.toml at the root of cases/ enables strict analyzer settings shared across all cases.

A smoke_test case exercises the runner end-to-end: a PHP file with a constant-valued variable triggers impossible-condition and redundant-comparison in the analyzer, and strict-types in the linter.

## 🔍 Context & Motivation

The existing per-crate test suites (analyzer, linter) are each siloed — adding a cross-tool test means touching one of them arbitrarily. A standalone runner in its own crate can exercise any combination of tools against a single PHP input.

## 🛠️ Summary of Changes

- **Feature:** Added a test suite runner that can exercise the mago cli in a snapshot based test setup.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): Test suite

## 🔗 Related Issues or PRs

:see_no_evil: 

## 📝 Notes for Reviewers

:crab:


cases/smoke_test/results.toml
```diff
@@ -0,0 +1,6 @@
+[analyzer."source.php"]
+impossible-condition = [4]
+redundant-comparison = [4]
```
azjezz: why is it reported 4 times here?

dotdash: that's a line number, not a count :D

dotdash: I guess we could record the whole span data here. Would check for accidental changes. Might generate noise but auto-updating the snapshots is easy enough. Thoughts?

azjezz: Actually.. can't we just use baseline? 2 baseline files, one for analyzer, one for linter. this verifies that our baseline feature also works! just adding `analyzer-baseline.toml` and `linter-baseline.toml` in the default config.

dotdash: Done. With the new baseline diff-reporting the test runner looks like this when things don't match up:
image

azjezz (approved):

PR #1816, opened 2026-05-10
refactor(analyzer): migrate test cases to directory-based suite

dotdash:
## 📌 What Does This PR Do?

Using the real binary with baseline files gives each case an exact, unambiguous record of expected output. Running each case twice also catches non-determinism that the in-process harness cannot detect.

A small set of cases surface new findings under the real binary that the old in-process harness missed (notably a new semantics check and a handful of isolated issues); these are recorded as-is in their baselines.

## 🔍 Context & Motivation

The test framework runs the analyzer in-process and matches reported issues against @mago-expect annotations embedded in source files. Adding a test requires a corresponding entry in mod.rs, which has grown to over 2,500 lines. One docblock annotation can suppress an arbitrary number of matching issues within a function's scope, making the expected/actual contract ambiguous.

## 🛠️ Summary of Changes

- **Refactor:** Ported the analyzer test suite to the new test runner.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): Tests

## 🔗 Related Issues or PRs

not really, https://github.com/carthage-software/mago/pull/1779 kind of triggered the whole new test runner thing

## 📝 Notes for Reviewers

I tried to make sure that all tests are ported 1:1, i.e. the old expectations and the new baseline mirror each other, but the semantics checks is strictly new, so there are more things in the base line in some cases. Also, the plugins are no longer all enabled by default for all test cases. I hope I catched all cases where some where needed to be enabled. I partially only learned about the problems after porting everything, so fixing things after the fact was a bit tricky.

If needed, I can go ahead and split this into severaval commits/PRs for easier review. But I'm a bit tired of this commit right now :sweat_smile:


azjezz:
As discussed internally, we can skip doing this. I think Its too much to do, and we can instead do this either slowly ( migrating some tests at a time ), or simply keep the current analyzer test framework - and use the new one for new tests.

PR #1857, opened 2026-05-15
fix(formatter): restore indent alignment for mid-line echo tags witho…

dotdash:
## 📌 What Does This PR Do?

Fixes a regression where function call arguments inside a `<?= ... ?>`
echo tag lost their indentation when the tag appeared after other HTML
content on a line with leading whitespace.

## 🔍 Context & Motivation

715db54b7 "fix(formatter): converge corpus idempotency on a single pass"
tightened the `should_apply_align` condition in `statement.rs` to only
apply `Document::Align` (which provides the indentation context for PHP
line continuations) to opening tags and tags whose entire line prefix is
whitespace. This was correct for the idempotency case it fixed, but it
also stopped applying alignment to legitimate mid-line echo tags,
causing argument continuation lines to lose the HTML indentation context:

## 🛠️ Summary of Changes

- **Bug Fix:** The formatter now preserves indentation for multi-line
  function calls inside inline echo tags that appear mid-line after
  HTML content.- **Refactor:** Improved `mago_ast` structure.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

### DIFF
--- a/crates/formatter/src/internal/format/statement.rs
+++ b/crates/formatter/src/internal/format/statement.rs
@@ -173,8 +173,36 @@ fn print_statement_slice<'ctx, 'arena>(
             if let Some(line_start_offset) = f.file.get_line_start_offset(line) {
                 let c = &f.source_text[line_start_offset as usize..offset as usize];
                 let ws = c.chars().take_while(|c| c.is_whitespace()).collect::<String>();
-                let should_apply_align =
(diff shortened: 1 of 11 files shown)


dotdash:
found remaining, related issues in our code base

PR #1873, opened 2026-05-19
feat(codex): add patches source for vendor type overrides

### LINKED ISSUE #1288: Support for patching types in vendor code

dotdash:
## 📌 What Does This PR Do?

A new `patches` source category (alongside `includes`) lets users point Mago at plain PHP files that redeclare vendor or built-in symbols with corrected type information. Like includes, patches are not analyzed, linted, or formatted — but their type declarations and PHPDoc take precedence over both vendored and built-in definitions.

A `PATCH` metadata flag and a new priority tier — USER_DEFINED > PATCH > BUILT_IN > VENDORED — implement this. For class-like symbols the replacement is partial: `apply_patch` merges only the method, property, and constant maps from the patch into the existing vendored class, leaving the class-level structure (parent, interfaces, traits, flags) intact. A patch that redeclares only `foo()` still preserves `bar()` from the original vendor class.

Correct operation requires non-patch files to be merged before patches so that `apply_patch` always finds an occupied entry to update rather than inserting a patch into a vacant slot. The pipeline scanner and both incremental paths enforce this ordering by partitioning partials into non-patch and patch groups before extending the codebase metadata. On incremental runs, unchanged patches are re-applied whenever a non-patch file changes, ensuring patches remain in effect after re-scans.

## 🔍 Context & Motivation

Vendor code included via `includes` may have incomplete or wrong type declarations. There is currently no way to correct those without editing vendor files directly or providing full class replacements that also obscure the original implementation.

## 🛠️ Summary of Changes

- **Feature:** Added support for patching types in vendor files.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #1288

## 📝 Notes for Reviewers

I guess I should just have someone else have a look for a change. I don't think it's "finished", I keep running into things I forgot about. Maybe someone else can give a more comprehensive feedback than my constant small incremental tinkerings.


crates/codex/src/metadata/class_like.rs
```diff
@@ -503,6 +506,301 @@ impl ClassLikeMetadata {
...
+    /// declarations must match the original exactly — a mismatch is a hard error indicating
+    /// the patch targets a different class. `use` trait declarations are never valid in a
+    /// patch and are always reported as an error.
+    ///
+    /// Diagnostics about the patch itself (mismatches, attempts to introduce new members)
+    /// are returned to the caller rather than pushed onto `self.issues`. The patched class
+    /// is vendor/built-in metadata; its `issues` list is reserved for diagnostics about
+    /// that definition. Patch diagnostics belong to the patch and the caller routes them to
+    /// the codebase-level patch-diagnostics collection.
+    #[must_use = "patch diagnostics must be routed to CodebaseMetadata::patch_diagnostics"]
```
azjezz: Why don't we keep diagnostics in this class_like issues map?

crates/codex/src/metadata/function_like.rs
```diff
@@ -324,4 +326,117 @@ impl FunctionLikeMetadata {
...
+                    "<anonymous>"
+                };
+                diagnostics.push(
+                    Issue::error(format!(
+                        "Patch for `{}` declares structural attributes (visibility, abstract, static, or \
+                         removing final) that differ from the original; only type annotations are applied.",
+                        name,
+                    ))
+                    .with_code(ScanningIssueKind::PatchMethodStructuralMismatch)
+                    .with_annotation(Annotation::primary(patch.span)),
```
azjezz: I wthink all issues should have help/note messages to further explain the problem.

crates/codex/src/metadata/class_like.rs
dotdash: `self` is a vendor/built-in here, so issues stored there won't be reported as they're not user defined. And the patch side currently doesn't make it into the final codebase, they just modify other entries, so any issues stored there would be lost.

dotdash:
Disclaimer: I had Claude rewrite the whole thing to use patch members in the CodebaseMetadata struct instead of maintaining a wholly separate codebase instance. I did not fully review the results, yet. For now, I'd just like feedback on the new approach, which eliminates the old "side-channel" for the patch diagnostics.

crates/codex/src/metadata/function_like.rs
dotdash: There should be a help note on all of them now. The new diagnostic for duplicate patches might still miss it. If so, I'll add that when I get around to review Claude's work in detail.

PR #1890, opened 2026-05-23
feat(docs): allow copying linter rule names without toggling details

### LINKED ISSUE #1882: Copping rule name from documentation
### Feature

Right now we can't copy rule name from documentation: https://mago.carthage.software/1.28.0/en/tools/linter/rules/#redundancy.
If we click on rule name, it collapse. We cannot simply copy rule name.

The idea is to allow users to easily copy rule name. By clicking on icon, or name of rule don't collapse tab.

mvanhorn:
## Summary

Lets users copy a linter rule code from the rules page in the docs without collapsing its `<details>` panel. Closes #1882.

## Why this matters

The rules page renders each rule inside `<details>` so the catalog stays scannable. The rule name sits inside `<summary>`, so every click on it flips the open/closed state before the user can finish a select-and-copy — confirmed against `docs/content/en/tools/linter/rules.md` (each rule wraps `<code class="rule__code">` in a `<summary>`). The issue links the live page at `https://mago.carthage.software/1.28.0/en/tools/linter/rules/#redundancy`, where the same behavior reproduces. Folks reaching for a rule name to drop into `mago.toml`, a chat reply, or an issue keep hitting the toggle instead.

The existing `docs/static/js/anchors.js` already stops `<summary> .rule__anchor` clicks from toggling (lines 4–6), so the precedent and the file are already there.

## 📌 What Does This PR Do?

Click a `<code class="rule__code">` inside a `<summary>` and the rule name lands on the clipboard, with a brief flash for confirmation. The `<details>` panel does NOT toggle on that click. Clicks elsewhere on the summary (the rule level chip, the surrounding padding) keep toggling exactly as before.

## 🔍 Context & Motivation

See "Why this matters" above. Behaviour drift from the existing `.rule__anchor` precedent is what makes the rule code uniquely uncopyable today.

## 🛠️ Summary of Changes

- `docs/static/js/anchors.js`: handler on `summary .rule__code` that stops the toggle (`preventDefault` + `stopPropagation`) and copies `code.textContent` via `navigator.clipboard.writeText()`. The handler returns early when `navigator.clipboard` is unavailable, leaving the native `<details>` toggle intact as a fallback. The success flash only fires after the write resolves so unsupported / blocked clipboard contexts don't get a fake "copied" highlight. A `WeakMap` keyed on the code element holds the cleanup timer so rapid re-clicks don't leak handlers.
- `docs/static/css/components.css`: `cursor: copy` on `.rule__code` for discoverability, plus a transient `.rule__code--copied` state that flashes for ~1.2s.

Diff is 36 lines across two files. No markdown content under `docs/content/**` changes.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [x] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #1882.

## 📝 Notes for Reviewers

- A `title="Click to copy rule name"` is added for discoverability; happy to swap to a small icon if you prefer.
- The handler is scoped to `summary .rule__code` only, so the rule level chip (`.rule__level`) and the bare summary background still toggle the section as before.

### DIFF
--- a/docs/static/css/components.css
+++ b/docs/static/css/components.css
@@ -1320,6 +1320,14 @@
     padding: 0;
     word-break: keep-all;
     flex: 0 0 auto;
+    cursor: copy;
+  }
+
+  details.rule .rule__code--copied {
+    background: var(--bg-soft);
+    box-shadow: 0 0 0 2px var(--accent-deep);
+    border-radius: 3px;
+    transition: background var(--t-snap), box-shadow var(--t-snap);
   }
 
   details.rule .rule__anchor {
--- a/docs/static/js/anchors.js
+++ b/docs/static/js/anchors.js
@@ -13,11 +13,13 @@
       event.preventDefault();
       event.stopPropagation();
 
+      if (!navigator.clipboard || !navigator.clipboard.writeText) return;
+
       try {
-        if (navigator.clipboard && navigator.clipboard.writeText) {
-          await navigator.clipboard.writeText(code.textContent || "");
-        }
-      } catch (_) {}
+        await navigator.clipboard.writeText(code.textContent || "");
+      } catch (_) {
+        return;
+      }
 
       code.classList.add("rule__code--copied");
       const previousTimer = copiedTimers.get(code);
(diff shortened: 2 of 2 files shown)


mvanhorn:
Thanks @azjezz for landing the docs change to copy linter rule codes without collapsing the details panel.

PR #1894, opened 2026-05-25
refactor(language-server): only publish diagnostics deltas

dotdash:
## 📌 What Does This PR Do?

Makes the LSP server only send the meaningful delta instead of all diagnostics to reduce load on the LSP client.

## 🔍 Context & Motivation

The server owns the diagnostics and unless a change is sent, the client is supposed to preserve the last sent set of diagnostics for a file. That means that instead of sending all diagnostics, we can just send them for the set of files where the diagnostics have changed. This saves the client from updating its view of the diagnostics, which can be quite slow.

## 🛠️ Summary of Changes

- **Refactor:** Made the LSP send only deltas for diagnostics.
- **Docs:** Updated installation guide.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): LSP


dotdash:
In a project with >80k diagnostics, this takes updating diagnostics in nvim from over a second to feeling almost instant.

PR #1962, opened 2026-06-09
feat: add formatter setting to combine opening tag and declare

### LINKED ISSUE #1959: Formatter won't move first opening tag and declare strict_types to the same line
This is a really cool project and as far as formatting goes, is the only serious formatter/beautifier for php that I could find in the sense that it is trying to be a full formatter with many options.

### Description

As the first line of each php file, I would like to have:

`<?php declare(strict_types=1);`

I am able to get the formatter to keep these together, but it won't combine them if they start out as separate lines, even when the settings seem like they should mean to format it this way.

**Additional Thoughts**
This seems like it could be a symptom of a deeper problem where it is not really a fully deterministic formatter in a certain sense. Issues like this could maybe be caught by randomly changing the formatting settings on a large code base, and then doing a format. As long as the code started out the same and the settings end up the same, any number of formats with different random settings in between should result in the same code in the end. Although if that large code base never had this starting code sometimes together and sometimes split up to begin with, that would still not have found the bug. So maybe the starting code also needs to be randomly changed in ways that still result in valid code.

### Playground Link

https://mago.carthage.software/latest/en/playground/#019ea84a-fece-8937-f022-abcdf26bafa9

### Reproduction

```
[formatter]
opening-tag-on-own-line = false
empty-line-after-opening-tag = false
empty-line-after-declare = true
```

I guess this is where the debate lies, do those settings mean it should format the code the way I want? It seems to me they should because if the opening tag is not on its own line, and there is not an empty line right after it, then shouldn't the declare be on the same line?

### Environment

- Mago 1.30.0
- Ubuntu 24.04
- PHP 8.4
- Mago running from downloaded release binary mago-1.30.0-x86_64-unknown-linux-gnu.tar.gz

mvanhorn:
## Summary

Adds an opt-in formatter setting, `combine-opening-tag-and-declare` (default `false`), that collapses an opening `<?php` tag and an immediately following `declare` statement onto a single line:

```php
```

When enabled, the tag and `declare` are combined regardless of how the source laid them out. The flag takes precedence over `opening-tag-on-own-line` and `empty-line-after-opening-tag` for this specific pairing.

## Why this matters

Resolves #1959. As discussed in that issue, Mago currently *preserves* `<?php declare(...);` on one line when the source already has them together, but never *combines* them when they start on separate lines, even with `opening-tag-on-own-line = false`. @azjezz noted that for users relying on `mago fmt` (not `mago lint`) supporting an opt-in collapse "might make sense". This adds exactly that, behind a flag, with the default behavior unchanged.

## Changes

- `crates/formatter/src/settings.rs`: new `combine_opening_tag_and_declare` field (`default_false`), documented inline.
- `crates/formatter/src/internal/format/statement.rs`: in the `OpeningTag` arm of `should_add_new_line_or_space_after_stmt`, when the flag is on and the next statement is a `Declare`, emit a space instead of a line break. Skipped when a trailing comment is on the opening tag or any comment sits between the tag and the `declare`, so comment placement is never rewritten.
- `crates/formatter/src/presets.rs`: set the new field to `false` in every preset (PSR-12, Pint, Tempest, Hack, Drupal, PER) to keep their output unchanged.
- `crates/formatter/tests/`: two snapshot cases — `combine_opening_tag_and_declare` (collapses) and `combine_opening_tag_and_declare_with_comment` (comment keeps them apart).
- `docs/content/en/tools/formatter/configuration-reference.md`: documents the new option.

## Testing

- `cargo test -p mago-formatter` — all 431 formatter cases pass (including the two new ones; the harness also checks idempotency on a reformat pass).
- `cargo clippy -p mago-formatter --all-targets` — clean.
- `cargo +nightly fmt` — no changes.
- Default output verified unchanged (flag defaults `false`; presets pin it `false`).

Fixes #1959

AI was used for assistance.


mvanhorn:
Thank you for merging, @azjezz. The combined opening tag and declare formatting option is a nice ergonomics win.

PR #1964, opened 2026-06-09
fix(analyzer): don't report trait property conflicts against @property tags

dotdash:
## 📌 What Does This PR Do?

Disable bogus reports about visibility mismatch conflicts

## 🔍 Context & Motivation

A class can declare a magic property via a `@property` docblock tag while also using a trait that declares a real property of the same name. Magic properties default to public read/write visibility, so the trait property conflict check compares them against the trait's declaration and reports a spurious `incompatible-property-visibility` error, claiming PHP will raise a fatal error.

Magic properties are not real declarations and never participate in PHP's trait composition, so they cannot conflict with a trait property. Skip them when collecting the class's own properties for the conflict check.

## 🛠️ Summary of Changes

- **Bug Fix:** don't report trait property conflicts against @property tags

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): analyzer

### DIFF
--- a/crates/analyzer/src/statement/class_like/mod.rs
+++ b/crates/analyzer/src/statement/class_like/mod.rs
@@ -2369,6 +2369,12 @@ fn check_trait_property_conflicts<'ctx, 'ast, 'arena, A>(
 
     let mut class_properties: IndexMap<Word, &PropertyMetadata> = IndexMap::new();
     for (property_name, property_metadata) in class_like_metadata.properties.iter().sorted_by_key(|(k, _)| *k) {
+        // Magic properties (from `@property` docblock tags) are not real declarations and
+        // never participate in PHP's trait composition, so they cannot conflict with a trait property.
+        if property_metadata.flags.is_magic_property() {
+            continue;
+        }
+
         if let Some(declaring_class) = class_like_metadata.declaring_property_ids.get(property_name)
             && declaring_class == &class_like_metadata.name
         {
--- a/crates/analyzer/tests/cases/trait_class_property_conflicts.php
+++ b/crates/analyzer/tests/cases/trait_class_property_conflicts.php
@@ -116,3 +116,24 @@ class C8
     public readonly string $prop = '';
     use T9;
 }
+
+// Test 9: Magic @property tag vs trait property (OK)
+// PHP: No error - magic properties are not real declarations and do not
+// participate in trait composition.
+trait T10
+{
+    protected int $prop;
+}
+
+/**
+ * @property int $prop
+ */
+class C9
+{
+    use T10;
+
+    public function __get(string $name): int
+    {
+        return $this->prop;
+    }
+}


crates/analyzer/src/statement/class_like/mod.rs
```diff
@@ -2369,6 +2369,12 @@ fn check_trait_property_conflicts<'ctx, 'ast, 'arena, A>(
 
     let mut class_properties: IndexMap<Word, &PropertyMetadata> = IndexMap::new();
     for (property_name, property_metadata) in class_like_metadata.properties.iter().sorted_by_key(|(k, _)| *k) {
+        // Magic properties (from `@property` docblock tags) are not real declarations and
+        // never participate in PHP's trait composition, so they cannot conflict with a trait property.
+        if property_metadata.flags.is_magic_property() {
+            continue;
+        }
```
azjezz, line 2376: True, but, this hides errors. potentially.

e.g, If you have `@property string $foo` and the trait declares the property as `int`. what happens in:

$obj->foo = "hello";

we need to tell the user that their declaration and annotation disagree.

If PHP does not complain, it does not mean we should not complain 😛

dotdash: The thing is that the pseudo property and the real property aren't really related to one another. The pseudo-property just documents the behavior of `__get`/`__set`. It has nothing to do with an actual property of the same name.

Consider this ([playground](https://mago.carthage.software/1.27.0/en/playground/#019eb1b6-dc40-1eb4-b2d2-738fe2d7f3da)):
```php
<?php

declare(strict_types=1);

trait XTrait
{
    protected int $valA;
    protected string $valB;
}

/**
 * @property int|null $valA
 * @property int|null $valB
 * @mago-ignore lint:single-class-per-file
 */
class XClass
{
    use XTrait;

    /**
     * @var array{ valA?: int, valB?: int }
     */
    private array $publicData = [];

    public function __construct() {
        $this->valA = 1;
        $this->valB = 'Peter';
    }

    public function __get(string $name): int
    {
        return match ($name) {
            'intVal' => $this->publicData['valA'] ?? 0,
            'stringVal' => $this->publicData['valB'] ?? 0,
            default => throw new Exception('Bad name'),
        };
    }

    public function __set(string $name, mixed $val): void
    {
        switch ($name) {
            case 'intVal':
                $this->publicData['valA'] = (int) $val;
                break;
            case 'stringVal':
                $this->publicData['valB'] = (int) $val;
                break;
            default:
                throw new Exception('Bad name');
        }
    }
}
```

Mago reports the visibility mismatch claiming that it's a crash, but the code works just fine, because the pseudo property just collides by name, but is otherwise completely separated from the actual property. Internal usage sees `$valB` as a string, external usage sees it as an int. 
And the type from the `@property` causes a bogus report for a type error on the assignment in `__set`. This seems like a bigger bug than I had expected.

PR #1975, opened 2026-06-12
fix(prelude): more precise types for some standard functions

gsteel:
Adds simple conditional return for `str_repeat()` + adds a handful of other more precise return types to other standard functions.


gsteel:
Adding `@param non-negative-int` to `str_repeat` has probably broken Psl checks. However, a negative number to `$times` is a `ValueError`

azjezz:
>Adding @param non-negative-int to str_repeat has probably broken Psl checks.

That's fine, will fix it in Psl.

PR #1977, opened 2026-06-13
feat(formatter): add preserve_breaking_binary_expression setting

### LINKED ISSUE #1853: Formatter: Add `preserve-breaking-binary-expression` option for logical/comparison operators
### Feature

Request a `preserve-breaking-binary-expression` option (analogous to the existing `preserve-breaking-*` family) that prevents the formatter from collapsing intentionally multi-line binary expressions onto a single line.

This extends [#1564](https://github.com/carthage-software/mago/issues/1564) (which covers `??` and `?:`) to also cover logical operators (`&&`, `||`, `and`, `or`) and comparison chains.

### Motivation

When integrating Mago as a fast pre-formatter alongside PHP-CS-Fixer in a large codebase (~7k files), the only remaining source of unwanted diffs is binary expression collapsing. All other formatting conflicts can be resolved with the existing `preserve-breaking-*` options:

- ✅ `preserve-breaking-member-access-chain`
- ✅ `preserve-breaking-argument-list`
- ✅ `preserve-breaking-parameter-list`
- ✅ `preserve-breaking-array-like`
- ✅ `preserve-breaking-conditional-expression`
- ✅ `preserve-breaking-condition-expression`
- ❌ **No equivalent for binary expressions**

### Examples

**Input (intentionally multi-line for readability):**
```php
return $this->property
    && Str::contains($this->property, $searchTerm, ignoreCase: true);
```

**Current output with `print-width = 999`:**
```php
return $this->property && Str::contains($this->property, $searchTerm, ignoreCase: true);
```

**Expected with `preserve-breaking-binary-expression = true`:**
```php
return $this->property
    && Str::contains($this->property, $searchTerm, ignoreCase: true);
```

### More complex example

**Input:**
```php
return (
    $this->conditionA
    && $this->checkSomething($this->conditionA, Category::FOO, ignoreCase: true)
)
    || (
        $this->conditionB
        && $this->checkSomething($this->conditionB, Category::BAR, ignoreCase: true)
    )
    || (
        $this->conditionB
        && $this->checkSomething($this->conditionB, 'some literal', ignoreCase: true)
    );
```

**Current output (`print-width = 999`):**
```php
return $this->conditionA && $this->checkSomething($this->conditionA, Category::FOO, ignoreCase: true) || $this->conditionB && $this->checkSomething($this->conditionB, Category::BAR, ignoreCase: true) || $this->conditionB && $this->checkSomething($this->conditionB, 'some literal', ignoreCase: true);
```

### Null-coalescing / throw pattern

Same issue from #1564, also applies to `?? throw`:
```php
// Input (readable):
return $this->findRecord(id: $id)
    ?? throw new RecordNotFoundException(model: $this, id: $id);

// Current output (collapsed):
return $this->findRecord(id: $id) ?? throw new RecordNotFoundException(model: $this, id: $id);
```

### Proposed behavior

When `preserve-breaking-binary-expression = true` (or similar name), the formatter should preserve intentional line breaks before binary operators (`&&`, `||`, `and`, `or`, `??`, `?:`, `.`) — same semantics as the other `preserve-breaking-*` options: if the developer placed the break, keep it; if they wrote it on one line, keep that too.

### Environment

- Mago version: 1.27.0
- Config: `print-width = 999` with all available `preserve-breaking-*` options enabled
- Use case: Mago as fast local pre-formatter alongside PHP-CS-Fixer in CI

mvanhorn:
## 📌 What Does This PR Do?

Adds a `preserve_breaking_binary_expression` setting to the formatter. When enabled, the formatter detects intentional newlines before binary operator right-hand operands and keeps the expression broken rather than collapsing it onto a single line.

## 🔍 Context & Motivation

The formatter currently collapses multi-line binary expressions (`&&`, `||`, `and`, `or`, `??`, `.`) onto a single line whenever they fit within `print-width`. This behavior is a blocker for projects that integrate Mago alongside PHP-CS-Fixer or that enforce a style where each operand appears on its own line.

The `preserve_breaking_*` family already handles member-access chains, argument lists, array literals, conditional expressions, and condition expressions. This PR extends the family with `preserve_breaking_binary_expression` so intentional line breaks before binary operators are respected.

Fixes #1853

## 🛠️ Summary of Changes

- **Feature:** Added `preserve_breaking_binary_expression: bool` (default `false`) to `FormatterSettings`, following the existing `preserve_breaking_*` pattern with the same doc-comment style.
- **Feature:** Extended `should_break` in `print_binaryish_expression_parts` (`binaryish.rs`) with a `has_new_line_in_range` check that fires when the setting is enabled and the source has a newline between the left operand and the right operand.
- **Chore:** Added `preserve_breaking_binary_expression: false` to all six preset structs in `presets.rs`.
- **Tests:** Registered two new formatter test cases (`preserve_breaking_binary_expression` and `preserve_breaking_binary_expression_disabled`) covering the enabled behavior and verifying that the default-off path is unchanged.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #1853

## 📝 Notes for Reviewers

The implementation mirrors exactly how `preserve_breaking_condition_expression` works: `has_new_line_in_range(f.source_text, left.span().end.offset, right.span().start.offset)` detects whether the developer placed a newline before the right operand. The setting defaults to `false` so existing behavior is entirely unchanged unless opted in.


crates/formatter/src/internal/format/binaryish.rs
```diff
@@ -322,7 +323,9 @@ where
         || f.has_placed_trailing_line_comment(right.span())
         || f.has_placed_leading_own_line_comment(right.span())
         || has_own_line_comment_in_left_chain(f, left)
-        || (is_original_right_parenthesized && has_placed_leading_comment_in_leftmost(f, original_right));
+        || (is_original_right_parenthesized && has_placed_leading_comment_in_leftmost(f, original_right))
+        || (f.settings.preserve_breaking_binary_expression
+            && misc::has_new_line_in_range(f.source_text, left.span().end.offset, right.span().start.offset));
```
azjezz:
```suggestion
            && misc::has_new_line_in_range(f.source_text, left.end_offset(), right.start_offset()));
```

mvanhorn:
The failing `type-syntax / parser` check is a pre-existing fuzz-target build break on `main`, not from this PR. `crates/type-syntax/fuzz/fuzz_targets/parser.rs` calls `arena.alloc_slice_copy(data)` but only imports `LocalArena` — not the `Arena` trait that provides `alloc_slice_copy` (via the blanket `impl<A: Allocator + ?Sized> Arena for A` in `mago-allocator`). With the trait out of scope, rustc raises `error[E0599]: no method named alloc_slice_copy found for struct LocalArena`. One-line fix in that fuzz target: `use mago_allocator::Arena;`.

This PR only touches `crates/formatter/`, so it's unaffected by the break. Happy to open a separate PR adding the import to unblock the fuzz CI if that's useful.

azjezz (changes requested):

azjezz:
@mvanhorn Could you please also update docs/? Thanks!

mvanhorn:
Done in c846b0bc:
- Applied the suggestion to use `left.end_offset()` / `right.start_offset()`.
- Added `preserve-breaking-binary-expression` to docs/content/en/tools/formatter/configuration-reference.md, grouped with the other `preserve-breaking-*` settings.

`cargo test -p mago-formatter` is green. Happy to sync the zh/fr reference tables too if you'd like them in the same PR.

mvanhorn:
Huge thanks for the merge, @azjezz. The preserve_breaking_binary_expression setting gives folks control over how the formatter handles those.

mvanhorn:
Appreciate the formatter change, @azjezz. preserve_breaking_binary_expression gives people a way to keep their intentional line breaks.

PR #1991, opened 2026-06-19
fix(linter): only flag boolean parameters used as flags (#1988)

### LINKED ISSUE #1988: False positive for Linter Rule: `no-boolean-flag`
### Description

`no-boolean-flag` does not check if the function contains any conditional branches.

My expectation would be that it should check the function body for any `if` statements or `?` ternary expressions to determine if the boolean is used as a flag or not.

### Playground Link

https://mago.carthage.software/latest/en/playground/#019eda01-3cad-601d-4fdb-6540bcf6bde4

### Reproduction

There is a playground link but this should make sense from a glance.

```php
    public function update(
        bool $myBool,
    ): void {
        $this->myBool = $myBool;
    }
```

### Environment

- 1.2.7.1 but it is also reproduceable in the playground.
- Linux (Debian)

### Word of Appreciation

Just FYI: I read the [performance promise](https://mago.carthage.software/1.30.0/en/benchmarks/#the-performance-promise) and was intrigued. When I tried it I was amazed. Tens of thousands of lines of code, without any cache, it finished running before I lifted my finger off the return key. 🥇

vjsai:
## Summary

Fixes #1988.

`no-boolean-flag-parameter` currently reports **every** boolean parameter, even when the
boolean is never used to branch behaviour. The minimal reproduction from the issue:

```php
public function update(
    bool $myBool,
): void {
    $this->myBool = $myBool;
}
```

Here `$myBool` is only stored on a property — it does not drive any control flow — so flagging
it as a boolean *flag* is a false positive. As suggested in the issue thread, this PR teaches the
rule to check how the parameter is actually used.

## Change

The rule now inspects the enclosing function-like body and only reports a boolean parameter when
it is used **as a flag** — that is, referenced somewhere other than as the value stored by a plain
assignment (`$x = $param;` / `$this->x = $param;`). Any other reference (a condition, a ternary, a
match/switch subject, a call argument, a compound expression, a `return`, …) still counts as flag
usage and is reported.

When there is no concrete body to inspect — abstract methods, interface methods, and arrow
functions — the parameter is reported exactly as before, so no existing report on those is lost.

The implementation reuses the existing `function_like_parts` helper to locate the body and walks it
with a small `MutWalker` that skips the right-hand side of plain assignments. The rule's
documentation examples are updated so the bad example demonstrates a real flag and the good example
documents the now-allowed "stored boolean" case.

## Before / after

```php
final class Connection
{
    private bool $secure;

    public function configure(bool $secure): void
    {
        $this->secure = $secure;   // before: reported  →  after: not reported (only stored)
    }
}

function get_difference(string $a, string $b, bool $ignore_case): string
{
    if ($ignore_case) {            // before: reported  →  after: still reported (drives behaviour)
        return strtolower($a) === strtolower($b) ? '' : $a;
    }

    return $a === $b ? '' : $a;
}
```

## Tests

Added `crates/linter/tests/no_boolean_flag_parameter.rs` covering both directions
(flagged: condition / ternary / call argument / compound expression; not flagged: stored to a
property, stored to a local, unused). The rule's good/bad examples are also exercised by the
existing `rule_examples` test.

```
$ cargo test -p mago-linter --test no_boolean_flag_parameter
running 7 tests
test does_not_flag_unused_boolean_parameter ... ok
test does_not_flag_boolean_parameter_only_assigned_to_property ... ok
test does_not_flag_boolean_parameter_only_assigned_to_local ... ok
test flags_boolean_parameter_passed_to_call ... ok
test flags_boolean_parameter_used_in_ternary ... ok
test flags_boolean_parameter_used_in_condition ... ok
test flags_boolean_parameter_used_in_compound_expression ... ok
test result: ok. 7 passed; 0 failed
```

All `mago-linter` tests pass, and `cargo fmt --all -- --check` plus
`cargo clippy -p mago-linter --all-features -- -D warnings` are clean. (The only failure I see
locally is an unrelated, pre-existing `mago-syntax` parser fixture — `expr_first_class_callables` —
which this change does not touch.)


crates/linter/tests/no_boolean_flag_parameter.rs
```diff
@@ -0,0 +1,69 @@
...
+fn count_issues(code: &str) -> usize {
+    let arena = LocalArena::new();
+    let file = File::ephemeral(Cow::Borrowed(b"test.php"), Cow::Owned(code.as_bytes().to_vec()));
+    let program = parse_file(&arena, &file);
+
+    let resolver = NameResolver::new(&arena);
+    let resolved_names = resolver.resolve(program);
+
+    let settings = Settings::default();
+    let registry = RuleRegistry::build(&settings, Some(&["no-boolean-flag-parameter".to_string()]), true);
```
azjezz: Please use `lint_success!` and `lint_failure!` macros under the rule definition. we don't create separate test files for rules.

crates/linter/src/rule/maintainability/no_boolean_flag_parameter.rs
```diff
@@ -140,3 +180,69 @@ impl LintRule for NoBooleanFlagParameterRule {
...
+                assignment.rhs,
+                Expression::Variable(Variable::Direct(variable)) if variable.name == self.target
+            );
+
+        if !is_plain_store {
+            self.walk_expression(assignment.rhs, context);
+        }
+
+        self.walk_expression(assignment.lhs, context);
+    }
```
azjezz, line 463: Wouldn't this flag `$this->enableFoo = $this->enableAll || $enableFoo`? I think we should just consider it "plain" if the operator is `=` without looking further. Or, only mark it as being used as a flag if it's used as condition? 🤔

Issue #1931, opened 2026-06-01
Mago 2.0 - Tracking Issue


azjezz:
# Mago 2.0 - Tracking Issue

Meta issue for the 2.0 release. 2.0 is a ground-up rework of the analysis pipeline plus the tooling and stability commitments that ride on top of it. This issue tracks the whole effort; each box links to (or will spawn) a dedicated issue/PR.

## Why 2.0

PHP is effectively **two languages**: the surface syntax, and the type language embedded in docblocks (`@var`, `@param`, `@template`, `@method`, `@psalm-type`, ...). They have different grammars and, in 1.x, are parsed by different code at different times, the syntax tree as one artifact, docblocks re-parsed on demand across a multiple scattered call sites.

2.0 fuses them. A single **lowering** pass turns the syntax tree into one typed **IR** that already has docblocks parsed, names resolved, and sugar normalized — and analysis is split into honest phases (scan → inference → check) that hand a typed program *forward* instead of re-deriving it. As a side effect we stop lying about the tree: what 1.x calls the "AST" is really a lossless CST (it keeps trivia and spans, which is why the formatter round-trips), so 2.0 names it a CST and reserves "the real, lowered tree" for the IR.

## The new pipeline

```
    source
      │ parse
      ▼
    CST            (mago_syntax::cst — lossless: trivia, spans, round-trips)
      │ lower      (per-file, NO cross-file info: name resolution + docblock parse + desugar)
      ▼
    IR<(), ()>     (typed-in-node tree, generic over context C and type T)
      │ scan
      ├─▶ linter
      ▼
    CodebaseMetadata
      │            IR + CodebaseMetadata
      ▼ inference
    IR<(), T>      (type info filled in the nodes)
      │ check
      ▼
    IR<C, T>       (context / "Env" filled: cross-references resolved, etc.)
      │
      ├─▶ security / taint analysis
      └─▶ reporting
```

Key decisions already settled in design:
- The IR is **arena-allocated**, generic over `C` (context) and `T` (type), with annotationion** (no interior mutability), the generics encode the analysis *phase* in the type system.
- **Lowering is fully local**, no codebase access. Linking a name to another file is the checker's on-demand job, not lowering's.
- The IR must be **retained and movable across phase boundaries** (inference output is the check input), so it lives in a cross-phase arena (`bumpalo-herd`) or as owned-per-file values, not borrowed from a transient per-thread bump.

### 1. Compiler core (critical path — roughly in order)

[...]

PR #2013, opened 2026-06-24
Add linter rule to disallow legacy array shorthand syntax (T[])

### LINKED ISSUE #1016: New Rule: Disallow legacy shorthand array syntax (`T[]`)

mvanhorn:
## 📌 What Does This PR Do?

Adds a linter rule that flags the PHP array shorthand type `T[]` in PHPDoc and suggests the explicit generic form (`array<T>` / `list<T>`), with an optional `check-generic-array` setting that also flags array-like generics missing an explicit key type.

## 🔍 Context & Motivation

`T[]` is ambiguous about whether the collection is a list or a keyed array, which the project's other consistency rules already discourage in favor of explicit `array<K, V>` / `list<T>`. This makes that preference enforceable. Requested in #1016.

## 🛠️ Summary of Changes

- **Feature:** Added the `no-shorthand-array-type` consistency rule, registered in the rule registry and exposed in linter settings, with an optional `check-generic-array` mode.

## 📂 Affected Areas

- [x] Linter

## 🔗 Related Issues or PRs

Fixes #1016

## 📝 Notes for Reviewers

The rule walks the structured PHPDoc type positions and reports shorthand array types; inline rule tests cover the shorthand cases and the optional generic-array check.


maximal:
Cool rule!

crates/linter/src/rule/consistency/no_shorthand_array_type.rs
```diff
@@ -0,0 +1,531 @@
...
+        for trivia in &program.trivia {
+            if trivia.kind != TriviaKind::DocBlockComment {
+                continue;
+            }
+
+            let document = PHPDocParser::parse_with_span(ctx.arena, trivia.value, trivia.span);
+            for tag in document.tags() {
+                self.check_tag_value(ctx, &tag.value);
+            }
+        }
```
azjezz, line 132: This is not ideal.

The problem here is this will re-parse docblocks, every rule that needs to work on docblocks requires re-parsing.

And this feature is delayed for 2.0 ( see #1931 ).

The reason we are delaying this for 2.0 is we want to have rules that work directly on PHPDoc CST, so we can parse docblocks just once for the whole file, and rules can target a specific PHPDoc CST node instead of targeting Program. this also means the `check_tag_value` function below becomes redundant, as this rule would only need to target `Type` node.

And for that reason, I'll be closing this. until the re-work is done where we are able to lint a specific PHPDoc CST node by itself without having to do parsing, i don't want to introduce linting rules for PHPDoc.

PR #2021, opened 2026-06-25
fix(analyzer): subtract enum cases from named enum subjects

### LINKED ISSUE #2019: False positive for Non-exhaustive `match` expression
### Description

When an enum is constructed using a custom method, it seems that the arms are not covered properly

### Playground Link

https://mago.carthage.software/1.40.2/en/playground/#019efdc6-e819-c3ea-eb01-66d725d88918

### Reproduction

_No response_

### Environment

Mago 1.40.2

xHeaven:
## 📌 What Does This PR Do?

Fixes a false positive `match-not-exhaustive` analyzer issue when matching over enum values returned from a method call.

## 🔍 Context & Motivation

`match` exhaustiveness narrowing correctly handled enum subject types represented as enum objects, but missed cases where the subject was represented as a named enum object. This caused exhaustive matches like match `(SomeEnum::fromNumber(...))` to be incorrectly reported as non-exhaustive.

## 🛠️ Summary of Changes

- **Bug Fix:** Subtract enum cases from named enum subject types during negated assertion reconciliation.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation

## 🔗 Related Issues or PRs

Fixes #2019

### DIFF
--- a/crates/analyzer/src/reconciler/negated_assertion_reconciler.rs
+++ b/crates/analyzer/src/reconciler/negated_assertion_reconciler.rs
@@ -220,32 +220,33 @@ fn subtract_complex_type<A>(
                 acceptable_types.push(existing_atomic);
             }
             (
+                TAtomic::Object(TObject::Named(existing_named_object)),
+                TAtomic::Object(TObject::Enum(TEnum { name: assertion_enum_name, case: Some(assertion_case) })),
+            ) if context
+                .codebase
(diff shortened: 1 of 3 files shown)


xHeaven:
Welp, too late, fixed in `6c72db0636098ddb3bdad0cde1c36bc5744fd7e0`.

PR #2030, opened 2026-06-27
fix(formatter): keep short chain attached after multiline receiver args

### LINKED ISSUE #1711: Unwanted line return on small method chain
The following:

```php
new UploadFileToSignedUrlRequest(
    upload_url: $upload->upload_url,
    content_type: $file->getMimeType() ?? 'application/octet-stream',
    content: (string) $file->getContent(),
    upload_headers: $upload->headers,
)->send()->dtoOrFail();
```

Is being formatted as this:
```php
new UploadFileToSignedUrlRequest(
    upload_url: $upload->upload_url,
    content_type: $file->getMimeType() ?? 'application/octet-stream',
    content: (string) $file->getContent(),
    upload_headers: $upload->headers,
)
    ->send()
    ->dtoOrFail();
```

This is particularly bad because the line before the chain is just one character

I'd assume `preserve-breaking-member-access-chain` would preserve my manual formatting here

Mago 1.21.0
```toml
php-version = "8.5.0"

[formatter]
print-width = 180
tab-width = 4
use-tabs = false
null-type-hint = "question"
indent-heredoc = false
inline-empty-method-braces = false
space-before-arrow-function-parameter-list-parenthesis = true
space-after-logical-not-unary-prefix-operator = true
always-break-named-arguments-list = false
preserve-breaking-argument-list = true
preserve-breaking-array-like = true
preserve-breaking-parameter-list = true
preserve-breaking-attribute-list = true
array-table-style-alignment = false=
```

mvanhorn:
## 📌 What Does This PR Do?

Fixes #1711. It stops the formatter from breaking a short member-access chain onto separate lines just because the chain's receiver is a `new X(...)` whose argument list is spread across multiple lines.

Before, this input:

```php
new UploadFileToSignedUrlRequest(
    upload_url: $upload->upload_url,
    content_type: $file->getMimeType() ?? 'application/octet-stream',
    content: (string) $file->getContent(),
    upload_headers: $upload->headers,
)->send()->dtoOrFail();
```

was reformatted so that `->send()` and `->dtoOrFail()` each moved to their own indented line, breaking right after a closing `)` whose line holds a single character. With this change the short trailing chain stays attached: `)->send()->dtoOrFail();`.

## 🔍 Context & Motivation

The chain-break decision lives in `MemberAccessChain::get_eligibility_score`. For an `Instantiation` receiver the score was incremented by a flat `+2` unconditionally. Combined with `+2` for each of the two trailing simple calls, a 2-link chain reached a score of `6`, which exactly hits the eligibility threshold of `6` for an instantiation base, so the chain was marked eligible to break.

As noted on the issue, `preserve-breaking-member-access-chain` only governs how an already-breaking chain breaks, not whether a short chain should break at all, so the existing options do not cover this case. The receiver's own multiline argument list (kept broken by `preserve-breaking-argument-list`) has nothing to do with whether the short trailing chain should break, yet it is the receiver that tips the score over the threshold.

## 🛠️ Summary of Changes

Gate the instantiation receiver's score contribution: the `+2` is no longer counted when the chain is short (`accesses.len() <= 2`) and every link is a simple call (no breaking arguments of its own) while the receiver's argument list is spread across multiple source lines. This drops the score below the threshold and keeps `)->send()->dtoOrFail()` inline.

Genuinely long chains (3+ links), links with breaking arguments, and instantiation receivers with inline argument lists are unaffected and still break exactly as before.

## 📂 Affected Areas

- `crates/formatter/src/internal/format/member_access.rs` (the eligibility heuristic)
- `crates/formatter/tests/cases/member_access_chain_short_after_breaking_receiver_args/` (new snapshot case)
- `crates/formatter/tests/mod.rs` (test registration)

## 🔗 Related Issues or PRs

Fixes #1711

## 📝 Notes for Reviewers

- New snapshot case `member_access_chain_short_after_breaking_receiver_args` covers the issue's exact input. The harness also asserts idempotency, so re-formatting the output is verified stable.
- `cargo test -p mago-formatter` passes all 439 snapshot cases, including the existing `member_access_chain*` and `preserve_breaking_member_access_chain*` cases, confirming the change is narrow.
- The multiline detection looks at whether the receiver's argument list span contains a newline in the source, which is the condition the reporter is hitting with `preserve-breaking-argument-list = true`.

### DIFF
--- a/crates/formatter/src/internal/format/member_access.rs
+++ b/crates/formatter/src/internal/format/member_access.rs
@@ -12,6 +12,7 @@ use mago_syntax::cst::ClassLikeMemberSelector;
 use mago_syntax::cst::Expression;
 use mago_syntax::cst::FunctionCall;
 use mago_syntax::cst::Identifier;
+use mago_syntax::cst::Instantiation;
 use mago_syntax::cst::MethodCall;
 use mago_syntax::cst::Node;
 use mago_syntax::cst::NullSafeMethodCall;
@@ -27,6 +28,8 @@ use crate::document::Line;
 use crate::internal::FormatterState;
 use crate::internal::format::Format;
 use crate::internal::format::call_arguments::print_argument_list;
+use crate::internal::format::call_arguments::promote_argument_list_to_partial;
+use crate::internal::format::call_arguments::should_break_all_arguments;
 use crate::internal::format::misc;
 use crate::internal::format::misc::is_breaking_expression;
 use crate::internal::format::misc::is_simple_call_argument;
@@ -127,13 +130,13 @@ where
                 // A `new X(...)` receiver normally pushes the chain toward breaking.
                 // However, when the chain is short (<= 2 links) and every link is a
                 // simple call (no breaking arguments of its own), a receiver whose
-                // *own* argument list is already spread across multiple lines should
+                // *own* argument list will print across multiple lines should
                 // not also drag the trailing short chain onto separate lines.
                 // See: https://github.com/carthage-software/mago/issues/1711
-                let receiver_args_multiline = instantiation.argument_list.as_ref().is_some_and(|argument_list| {
-                    let span = argument_list.span();
-                    misc::has_new_line_in_range(f.source_text, span.start.offset, span.end.offset)
-                });
+                let receiver_args_multiline = instantiation
+                    .argument_list
+                    .as_ref()
+                    .is_some_and(|argument_list| receiver_argument_list_will_break(f, instantiation, argument_list));
 
                 let short_simple_chain = self.accesses.len() <= 2
                     && self.accesses.iter().all(|access| match access.get_arguments_list() {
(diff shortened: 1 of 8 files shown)


azjezz:
This feature is non idempotent, see CI failure.

mvanhorn:
Good catch, thanks. Fixed in 9f3585a and rebased onto current main.

Root cause: the attachment heuristic read the source layout (has_new_line_in_range over the receiver's argument span), so a first pass that collapsed the receiver's arguments changed the second pass's decision. It now computes whether the receiver's argument list will actually print multiline, which is stable across passes.

Added a regression case that is a fixed point under double-format. All 444 formatter case tests pass locally.

azjezz:
@mvanhorn still failing on the corpus :/

mvanhorn:
Fixed the corpus `fmt` idempotency failure in 3b46272.

Root cause: `receiver_argument_list_will_break()` returned `true` for an empty argument list (`new static()`) and for any receiver whose flat width couldn't be computed — the latter happens on a second format pass once the receiver is already printed across multiple lines, so the trailing short chain's attach/break decision depended on the current formatting state and `mago fmt` wasn't idempotent.

Now empty argument lists short-circuit to `false`, and I use `is_some_and` so an un-computable width no longer forces a "will break" verdict. The #1711 behaviour is preserved — both fixture cases still pass, and I verified `mago fmt && mago fmt --check` is clean across the full corpus (55,549 files) locally.

mvanhorn:
Thank you for the review, @azjezz - keeping short chains attached after multiline receiver args was a fun formatter corner to chase.

PR #2041, opened 2026-06-30
fix(formatter): make member-access chain formatting idempotent with both preserve options

### LINKED ISSUE #2032: Formatter: `mago format` is not idempotent
### Description

`mago format` is not idempotent; it oscillates between two layouts instead of converging. I would expect a second run to be a no-op.

Triggers only when **both** options are on (either one alone is stable):

```toml
preserve-breaking-member-access-chain-first-method-on-same-line = true
```

### Playground Link

I wasn't able to reproduce this on the playground. I couldn't figure out how to enable these formatter settings.

### Reproduction

`mago.toml`:

```toml
version = "1"
php-version = "8.5.0"

[formatter]
preserve-breaking-member-access-chain-first-method-on-same-line = true
```

Here's the line that oscillates back and forth

```php
<?php

$result = bagOf(array_values($theInputCollection))->containsAll($firstNeedle, $secondNeedle, $thirdNeedleValue ?? $thirdNeedle);
```

Run `mago format` repeatedly. Odd runs:

```php
$result = bagOf(array_values($theInputCollection))
    ->containsAll($firstNeedle, $secondNeedle, $thirdNeedleValue ?? $thirdNeedle);
```

Even runs:

```php
$result = bagOf(array_values($theInputCollection))->containsAll(
    $firstNeedle,
    $secondNeedle,
    $thirdNeedleValue ?? $thirdNeedle,
);
```

Workaround: set `preserve-breaking-member-access-chain-first-method-on-same-line = false`; output converges to the chain-broken layout and stays.

Here's an MRE PHP file that compiles:

<details><summary>MRE</summary>
<p>

```php
<?php

declare(strict_types=1);

final class NumberBag
{
    /** @param list<int> $items */
    public function __construct(
        private array $items,
    ) {}

    public function containsAll(int $first, int $second, int $third): bool
    {
        return in_array($first, $this->items, true)
            && in_array($second, $this->items, true)
            && in_array($third, $this->items, true);
    }
}

/** @param list<int> $items */
function bagOf(array $items): NumberBag
{
    return new NumberBag($items);
}

$theInputCollection = [1, 2, 3, 4, 5];
$firstNeedle = 2;
$secondNeedle = 4;
$theFinalNeedleValueToCheck = 5;

$result = bagOf(array_values($theInputCollection))->containsAll($firstNeedle, $secondNeedle, $theFinalNeedleValueToCheck);

var_dump($result);
```

</p>
</details> 

### Environment

- Mago: 1.42.0
- OS: macOS (Darwin 25.5)
- Installed via: prebuilt binary (`~/.local/bin`)

mvanhorn:
## 📌 What Does This PR Do?

Makes `mago format` idempotent when both `preserve-breaking-member-access-chain` and `preserve-breaking-member-access-chain-first-method-on-same-line` are enabled. Before this change the formatter oscillated between two layouts on alternating runs instead of converging, so a second run was never a no-op.

## 🔍 Context & Motivation

Reported in #2032. With both options on, this line never reaches a fixed point:

```php
$result = bagOf(array_values($theInputCollection))->containsAll($firstNeedle, $secondNeedle, $thirdNeedleValue ?? $thirdNeedle);
```

Odd runs break the chain onto its own line; even runs keep the chain attached and break the argument list. Each layout reformats into the other.

The decision lives in `should_inline_first_access` in `crates/formatter/src/internal/format/member_access.rs`. Its `preserve_same_line_first_method` branch inlines the first method when the chain is already broken, broke after the first access, or exceeds the print width. The first two of those are source-shape signals that differ between the two layouts, so they were meant to be backstopped by the print-width signal. But `exceeds_print_width` relies on `get_flat_width`, and the underlying `get_flat_expression_width` returns `None` for a binary expression such as `$x ?? $y`. With a `??` argument in the chain the width was never computed, the print-width signal dropped out, and the verdict flipped from pass to pass.

## 🛠️ Summary of Changes

- **Bug Fix:** Added `exceeds_print_width_from_line_start` and a binary-argument-aware width path (`get_flat_width_with_binary_arguments` and its `get_*_with_binary_arguments` helpers) in `member_access.rs`. The new width computation handles `Binary` expressions and function/method/static-method call arguments, and measures from the line start (including indentation), so the print-width signal is available even when an argument is a binary expression. `should_inline_first_access` now uses this stable signal in the `preserve_same_line_first_method` branch, so the verdict is the same whether the input is the broken or the attached layout. Both options converge on: chain attached to the receiver, argument list broken.
- **Tests:** Added `issue_2032` (the reported `??` case) and `issue_2032_nested_function_argument` (a `??` wrapped in a function call, e.g. `normalize($c ?? $d)`, which the first pass of the width helper still missed until a `Call::Function` arm was added). The `test_case!` harness asserts both `format(before) == after` and `format(after) == after`, so each case proves idempotency.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #2032

## 📝 Notes for Reviewers

The behavioral change is scoped to the `preserve_same_line_first_method` branch (both preserve options on); the single-option and default paths are untouched, and the existing member-access, idempotency, same-line-chain, and method-chain test cases still pass. Verified with `cargo test -p mago-formatter` (440 case tests + 48 precedence tests) and `cargo +nightly fmt --all -- --check --unstable-features`.

### DIFF
--- a/crates/formatter/src/internal/format/member_access.rs
+++ b/crates/formatter/src/internal/format/member_access.rs
@@ -744,6 +744,9 @@ fn get_flat_expression_width_with_binary_arguments(expression: &Expression<'_>) 
                 .zip(get_selector_width(&access.property))
                 .map(|(object, property)| object + string_width(b"?->") + property)
         }
+        Expression::Call(Call::Function(call)) => get_flat_expression_width_with_binary_arguments(call.function)
+            .zip(get_argument_list_width_with_binary_arguments(&call.argument_list))
+            .map(|(function, arguments)| function + arguments),
         Expression::Call(Call::Method(call)) => get_flat_expression_width_with_binary_arguments(call.object)
             .zip(get_selector_width(&call.method))
             .zip(get_argument_list_width_with_binary_arguments(&call.argument_list))
--- a/crates/formatter/src/internal/format/member_access.rs
+++ b/crates/formatter/src/internal/format/member_access.rs
@@ -309,6 +309,22 @@ where
         width > f.settings.print_width
     }
 
+    fn exceeds_print_width_from_line_start(&self, f: &FormatterState<'_, '_, A>) -> bool {
+        let Some(width) = self.get_flat_width_with_binary_arguments() else {
+            return false;
+        };
+
+        let base_start = self.base.span().start.offset;
+        let line = f.file.line_number(base_start);
+        let Some(line_start) = f.file.get_line_start_offset(line) else {
+            return false;
+        };
+
+        let prefix_width = string_width(&f.source_text[line_start as usize..base_start as usize]);
+
+        prefix_width + width > f.settings.print_width
+    }
+
     #[inline]
     fn has_break_after_first_access(&self, f: &FormatterState<'_, '_, A>) -> bool {
         for (i, access) in self.accesses.iter().enumerate() {
@@ -365,6 +381,15 @@ where
         let mut width = get_flat_expression_width(self.base)?;
         for access in &self.accesses {
             width += access.get_flat_width()?;
+        }
+
+        Some(width)
+    }
+
+    fn get_flat_width_with_binary_arguments(&self) -> Option<usize> {
+        let mut width = get_flat_expression_width_with_binary_arguments(self.base)?;
+        for access in &self.accesses {
+            width += get_access_flat_width_with_binary_arguments(access)?;
         }
 
         Some(width)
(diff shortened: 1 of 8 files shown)


mvanhorn:
Thanks @azjezz! Glad the member-access chain formatting is idempotent now.

caliguIa:
I'm still seeing this issue on version 1.4.5

PR #2045, opened 2026-07-02
feat(prelude): add stubs for the sqlsrv extension

### LINKED ISSUE #2039: Mago has no stubs for the PHP SQL Server extension
### Description

Seems like Mago is missing the support of ext/sqlsrv

### Playground Link

https://mago.carthage.software/latest/en/playground/#019f1477-e421-8f57-8201-f06f50a3ca67

### Reproduction

_No response_

### Environment

- Mago 1.42
- macOS 26

Soean:
## 📌 What Does This PR Do?

Adds stubs for the `ext/sqlsrv` extension (Microsoft SQL Server Driver for PHP) to the prelude, so Mago recognizes all `sqlsrv_*` functions and `SQLSRV_*` constants instead of reporting them as unknown symbols.

## 🔍 Context & Motivation

Mago currently has no stubs for `ext/sqlsrv`, so any project using the SQL Server driver gets false positives for every function call and constant reference (see #2039). All other database extensions (`mysqli`, `pgsql`, `sqlite3`, `pdo`) already ship stubs — this fills the gap.

## 🛠️ Summary of Changes

- **Feature:** Added `crates/prelude/assets/extensions/sqlsrv.php` covering all 26 `sqlsrv_*` functions documented on php.net, the 10 parameterized type-constructor functions (`SQLSRV_PHPTYPE_STRING()`, `SQLSRV_SQLTYPE_VARCHAR()`, etc.), and all constants registered by the driver. Connection/statement handles are typed as `resource` via docblocks, result-shaping functions carry precise union return types (`array|null|false` etc.), and the `VARCHAR`/`NVARCHAR`/`VARBINARY` constructors accept `int|'max'`.
- **Feature:** Registered the extension's functions in `EXTENSION_FUNCTIONS` (`crates/linter/src/rule/utils/consts.rs`) so the disallowed-functions linter rule can target the `sqlsrv` extension as a whole.

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): Prelude (extension stubs)

## 🔗 Related Issues or PRs

Fixes #2039

### DIFF
--- a/crates/prelude/assets/extensions/sqlsrv.php
+++ b/crates/prelude/assets/extensions/sqlsrv.php
@@ -1,0 +1,362 @@
+<?php
+
+// Error / warning selectors
+
+const SQLSRV_ERR_ERRORS = 0;
+
+const SQLSRV_ERR_WARNINGS = 1;
+
+const SQLSRV_ERR_ALL = 2;
+
+// Logging subsystems
+
+const SQLSRV_LOG_SYSTEM_ALL = -1;
+
+const SQLSRV_LOG_SYSTEM_OFF = 0;
+
+const SQLSRV_LOG_SYSTEM_INIT = 1;
+
+const SQLSRV_LOG_SYSTEM_CONN = 2;
+
+const SQLSRV_LOG_SYSTEM_STMT = 4;
+
+const SQLSRV_LOG_SYSTEM_UTIL = 8;
+
+// Logging severities
+
+const SQLSRV_LOG_SEVERITY_ALL = -1;
+
+const SQLSRV_LOG_SEVERITY_ERROR = 1;
+
+const SQLSRV_LOG_SEVERITY_WARNING = 2;
+
+const SQLSRV_LOG_SEVERITY_NOTICE = 4;
+
+// Fetch types
+
+const SQLSRV_FETCH_NUMERIC = 1;
+
+const SQLSRV_FETCH_ASSOC = 2;
+
+const SQLSRV_FETCH_BOTH = 3;
+
+// PHP types (values 4 and 6 belong to SQLSRV_PHPTYPE_STRING() and SQLSRV_PHPTYPE_STREAM())
+
+const SQLSRV_PHPTYPE_NULL = 1;
+
+const SQLSRV_PHPTYPE_INT = 2;
+
+const SQLSRV_PHPTYPE_FLOAT = 3;
+
+const SQLSRV_PHPTYPE_DATETIME = 5;
+
+const SQLSRV_PHPTYPE_TABLE = 7;
+
+// Encodings
+
+const SQLSRV_ENC_BINARY = 'binary';
+
+const SQLSRV_ENC_CHAR = 'char';
+
+// Nullability
+
+const SQLSRV_NULLABLE_NO = 0;
+
+const SQLSRV_NULLABLE_YES = 1;
+
+const SQLSRV_NULLABLE_UNKNOWN = 2;
+
+// SQL types
+
+const SQLSRV_SQLTYPE_BIGINT = -5;
+
+const SQLSRV_SQLTYPE_BIT = -7;
+
+const SQLSRV_SQLTYPE_CHAR = 1;
+
+const SQLSRV_SQLTYPE_DATE = 5211;
+
+const SQLSRV_SQLTYPE_DATETIME = 25177693;
+
+const SQLSRV_SQLTYPE_DATETIME2 = 58734173;
+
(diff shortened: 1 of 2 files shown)


crates/prelude/assets/extensions/sqlsrv.php
```diff
@@ -0,0 +1,362 @@
+<?php
+
+// Error / warning selectors
```
azjezz: Please remove those comments.

crates/prelude/assets/extensions/sqlsrv.php
```diff
@@ -0,0 +1,362 @@
...
+
+const SQLSRV_ERR_ERRORS = 0;
+
+const SQLSRV_ERR_WARNINGS = 1;
```
azjezz, line 5: Are the values of those constants fixed? If no, we should add `/** @var int */`.

crates/prelude/assets/extensions/sqlsrv.php
```diff
@@ -0,0 +1,362 @@
...
+
+function SQLSRV_SQLTYPE_BINARY(int $byteCount): int {}
+
+function SQLSRV_SQLTYPE_CHAR(int $charCount): int {}
+
+function SQLSRV_SQLTYPE_DECIMAL(int $precision, int $scale): int {}
+
+function SQLSRV_SQLTYPE_NCHAR(int $charCount): int {}
+
+function SQLSRV_SQLTYPE_NUMERIC(int $precision, int $scale): int {}
```
azjezz, line 347: can we add more precise types here? int ranges, non-empty strings..etc?

crates/prelude/assets/extensions/sqlsrv.php
```diff
@@ -0,0 +1,362 @@
...
+const SQLSRV_CURSOR_KEYSET = 'keyset';
+
+const SQLSRV_CURSOR_CLIENT_BUFFERED = 'buffered';
+
+/**
+ * @param array<string, mixed> $connectionInfo
+ *
+ * @return resource|false
+ */
+function sqlsrv_connect(string $serverName, array $connectionInfo = []): mixed {}
```
azjezz: Are parameter names fixed across all versions? If not, we should add `@no-named-arguments`

crates/prelude/assets/extensions/sqlsrv.php
Soean: Yes, they're fixed. I checked the values against the `microsoft/msphpsql` source across the driver's published history (v3.0.1 → v5.13):

- The simple constants map to stable underlying values — ODBC codes (`SQL_BIGINT`, `SQL_PARAM_OUTPUT`, …) and driver enums (`SQLSRV_ERR_*`, `SQLSRV_FETCH_*`, …) that are identical across every version.
- The `SQLSRV_PHPTYPE_*` enum order is stable, so `DATETIME` is `5` throughout (`STRING` has always preceded it — even v3.0.1). `SQLSRV_PHPTYPE_TABLE = 7` was added later, but additively, without shifting existing values.
- The composite `SQLSRV_SQLTYPE_*` values (datetime/money) are computed deterministically from a fixed ODBC type + size/scale (e.g. `SQL_TYPE_TIMESTAMP`, size 23, scale 3 → `25177693`), identical across versions.

So I kept the literal values rather than `/** @var int */`, which also lets the analyzer fold constant comparisons. Happy to switch to `/** @var int */` if you'd prefer that as the house style regardless (I see `pgsql.php` uses it uniformly).

crates/prelude/assets/extensions/sqlsrv.php
Soean: Yes, I compared the `arginfo` across the driver's entire published history (v3.0.1 → v5.13.1) and every parameter name is identical; the only change in 10+ years is additive, never a rename. So instead of `@no-named-arguments` I used the extension's actual (snake_case) arginfo names, which differ from the php.net prose but are what count for named arguments at runtime, so `sqlsrv_query($c, tsql: ...)` is accepted while `sqlsrv_query($c, sql: ...)` is correctly flagged.

crates/prelude/assets/extensions/sqlsrv.php
Soean: Done

crates/prelude/assets/extensions/sqlsrv.php
Soean: Done

PR #2065, opened 2026-07-07
feat: Add support for the @Final annotation during structure validation.

Norbytus:
## 📌 Add support for the ```@final``` annotation during structure validation.

Added the is_final check not only for the Class structure, but also for ClassLikeMetadata.

## 🔍 Context & Motivation

Besides using the final keyword as a class modifier, the ```@final``` annotation is also sometimes used. This is especially common when working with Doctrine.

## 🛠️ Summary of Changes

- **Feature:** Update guard structural, count ```@final``` annotation

## 📂 Affected Areas

- [X] Guard

### DIFF
--- a/crates/guard/tests/mod.rs
+++ b/crates/guard/tests/mod.rs
@@ -22,6 +22,9 @@ use mago_guard::settings::PerimeterSettings;
 use mago_guard::settings::PermittedDependency;
 use mago_guard::settings::PermittedDependencyKind;
 use mago_guard::settings::Settings;
+use mago_guard::settings::StructuralRule;
+use mago_guard::settings::StructuralSettings;
+use mago_guard::settings::StructuralSymbolKind;
 use mago_names::resolver::NameResolver;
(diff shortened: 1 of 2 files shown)


crates/guard/tests/mod.rs
```diff
@@ -597,3 +600,33 @@ pub fn test_narrow_rule_not_widened_by_broad_catchall() {
...
+#[test]
+pub fn test_is_final_annotation() {
+    let code = indoc! {r"
+        <?php
+            namespace App\Entity {
+                final class Test {
+
+                }
+            }
+    "};
```
azjezz, line 614: This does not test docblock. only modifier? the test would pass without the modification above.

Norbytus: Yes my bad.

Norbytus: Fix test, and fix search ClassLikeMetadata, cause first time i try find by ```class.name.value``` change use fqn variable

PR #2097, opened 2026-07-17
fix(cli): restore default SIGPIPE handling so piped output does not panic

### LINKED ISSUE #2090: Crash when piping "mago format --dry-run" into `head`
### Description

When running `mago format --dry-run | head -n10` to only see the first 10 lines of output, mago crashes:
```
$ mago format --dry-run | head -n10
diff of 'app/Notifications/ResetPassword.php':
--- original
+++ modified
@@ -36,8 +36,8 @@
     public function __construct($client_name, string $token, string $email)
     {
         $this->client_name = $client_name;
-        $this->token = $token;
-        $this->email = $email;
+        $this->token       = $token;

thread 'main' (1540748) panicked at /rustc/ac68faa20c58cbccd01ee7208bf3b6e93a7d7f96/library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

`RUST_BACKTRACE=full` does not help since debugging symbols have been removed
<details>
```
thread 'main' (1540717) panicked at /rustc/ac68faa20c58cbccd01ee7208bf3b6e93a7d7f96/library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
stack backtrace:
   0:     0x5644674c4d8a - <unknown>
   1:     0x5644669d48da - <unknown>
   2:     0x5644674c4385 - <unknown>
   3:     0x5644674c4247 - <unknown>
   4:     0x5644674e85d7 - <unknown>
   5:     0x5644674e8589 - <unknown>
   6:     0x5644674e8a8c - <unknown>
   7:     0x5644664eedfb - <unknown>
   8:     0x56446670ea24 - <unknown>
   9:     0x564466797f26 - <unknown>
  10:     0x5644667967fd - <unknown>
  11:     0x5644664c36d0 - <unknown>
  12:     0x5644666cbd13 - <unknown>
  13:     0x56446694b7c9 - <unknown>
  14:     0x7f95ce08724a - __libc_start_call_main
                               at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
  15:     0x7f95ce087305 - __libc_start_main_impl
                               at ./csu/../csu/libc-start.c:360:3
  16:     0x5644665986f9 - <unknown>
  17:                0x0 - <unknown>
```
</details>

### Playground Link

_No response_

### Reproduction

Run `mago format --dry-run | head -n10`

### Environment

- Mago: 1.43.0
- OS: Debian Linux 12
- Installed via download from github releases `mago-1.43.0-x86_64-unknown-linux-gnu.tar.gz`

mvanhorn:
## 📌 What Does This PR Do?

Restores the default `SIGPIPE` disposition on Unix, so piping a Mago command into a reader that exits early (`head`, `less`) terminates quietly instead of panicking.

## 🔍 Context & Motivation

`mago format --dry-run | head -n10` currently panics:

```
thread 'main' panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
```

Rust ignores `SIGPIPE` at startup, so writing to a closed pipe surfaces as an `EPIPE` error instead of terminating the process, and `println!` panics on it.

`apply_update` in `src/utils/mod.rs` is only where #2090 happened to surface. There are ~176 `println!` call sites across `src/`, so `mago ast | head`, `mago cst | less` and friends are exposed the same way, and @cweiske reported the same panic with `| less` after quitting the first page. Patching individual call sites would leave the rest of that surface broken, so this restores the default disposition once at startup, which is what `cat`, `grep` and `rg` do.

## 🛠️ Summary of Changes

- **Bug Fix:** Restore default `SIGPIPE` handling at CLI startup so piped output no longer panics with "failed printing to stdout: Broken pipe".
- **Dependencies:** Add a Unix-gated `libc` dependency to the `mago` crate. `libc` is already declared in `[workspace.dependencies]`, so this references it with `{ workspace = true }`.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [x] CLI
- [x] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Fixes #2090

## 📝 Notes for Reviewers

Verified against the reported reproduction, same fixture and same command (`mago format --dry-run | head -n5`):

| build | result |
|---|---|
| before | `thread 'main' panicked ... failed printing to stdout: Broken pipe (os error 32)` |
| after | no panic, no stderr output; Mago is terminated by `SIGPIPE` (the pipeline shows exit `0` because `head` exits `0`, or `141` under `set -o pipefail`) |

`reset_sigpipe()` is called immediately after `main_start` is captured, so the existing "captured before anything else" timing comment still holds. The `libc::signal` call runs before any threads are spawned, which is why the `unsafe` block is sound.

Windows is unaffected: it has no `SIGPIPE`, so the `#[cfg(not(unix))]` arm is a no-op.

One behavioral note worth a second opinion: on a closed pipe the process now dies from `SIGPIPE` rather than unwinding, so destructors do not run in that path. That is conventional for CLI tools, but flag it if Mago relies on teardown work at exit.

Local runs of the CI checks all pass: `cargo check --workspace --locked`, `cargo fmt --all -- --check`, `cargo clippy --workspace --all-features -- -D warnings`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -309,6 +309,10 @@ mago-prelude = { workspace = true, features = ["build"] }
 [target.'cfg(any(target_os = "macos", target_os = "windows", target_env = "musl", target_env = "gnu"))'.dependencies]
 mimalloc = { version = "0.1.47" }
 
+# TODO(azjezz): replace with https://dev-doc.rust-lang.org/beta/unstable-book/language-features/unix-sigpipe.html once stable.
+[target.'cfg(unix)'.dependencies]
+libc = { workspace = true }
+
 [features]
 dhat-heap = ["dhat"]
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -92,6 +92,22 @@ static ALLOC: dhat::Alloc = dhat::Alloc;
 /// Exit code 2 indicates the tool itself failed to complete its operation.
 const EXIT_CODE_ERROR: u8 = 2;
 
+/// Restores the default SIGPIPE behavior so output piped into `head` or `less` exits quietly.
+///
+/// Rust ignores SIGPIPE by default, turning writes to a closed pipe into EPIPE errors that make
+/// `println!` panic. Restoring the default lets standard Unix pipe behavior terminate the process.
+#[cfg(unix)]
+fn reset_sigpipe() {
+    // SAFETY: Resetting SIGPIPE to default (SIG_DFL) is safe.
+    unsafe {
+        libc::signal(libc::SIGPIPE, libc::SIG_DFL);
+    }
+}
+
+/// Windows has no SIGPIPE.
+#[cfg(not(unix))]
+fn reset_sigpipe() {}
+
 /// Entry point for the Mago CLI application.
 ///
 /// This function initializes the heap profiler (if enabled), runs the main application logic,
@@ -113,6 +129,7 @@ pub fn main() -> ExitCode {
     // trace event is emitted from inside `run` once tracing is wired up;
     // telemetry in this function only matters when `MAGO_LOG=trace`.
     let main_start = Instant::now();
+    reset_sigpipe();
 
     #[cfg(feature = "dhat-heap")]
     let _profiler = dhat::Profiler::new_heap();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1401,6 +1401,7 @@ dependencies = [
  "flate2",
  "indicatif",
  "indoc",
+ "libc",
  "libtest-mimic",
  "mago-algebra",
  "mago-allocator",


Cargo.toml
```diff
@@ -309,6 +309,9 @@ mago-prelude = { workspace = true, features = ["build"] }
 [target.'cfg(any(target_os = "macos", target_os = "windows", target_env = "musl", target_env = "gnu"))'.dependencies]
 mimalloc = { version = "0.1.47" }
 
+[target.'cfg(unix)'.dependencies]
+libc = { workspace = true }
```
azjezz, line 314: Please add a comment above this to replace it with https://dev-doc.rust-lang.org/beta/unstable-book/language-features/unix-sigpipe.html once stable.

src/main.rs
```diff
@@ -92,6 +92,22 @@ static ALLOC: dhat::Alloc = dhat::Alloc;
 /// Exit code 2 indicates the tool itself failed to complete its operation.
 const EXIT_CODE_ERROR: u8 = 2;
 
+/// Restores the default SIGPIPE behavior so output piped into `head` or `less` exits quietly.
+///
+/// Rust ignores SIGPIPE by default, turning writes to a closed pipe into EPIPE errors that make
+/// `println!` panic. Restoring the default lets standard Unix pipe behavior terminate the process.
+#[cfg(unix)]
+fn reset_sigpipe() {
+    // SAFETY: This runs before any threads are spawned, and `signal` is async-signal-safe.
```
azjezz: This comment is misleading, "async-signal-safe" is irrelevant because this call is not made from a signal handler

azjezz (commented):
Overall looks good, but i think we should add a test to CI ( probably in the corpus action ), that pipes mago via `head` or `less`.

azjezz:
>exits 0, no panic, no stderr output

This is wrong, Mago is terminated by `SIGPIPE`, the overall pipeline appears successful because `head` itself exits with 0. With pipefail, it will/should report 141. That is normal Unix behavior, but the PR description is wrong.

mvanhorn:
You're right — thanks for the correction. The pipeline exits 0 because `head` exits 0 after the `SIGPIPE`, and with `pipefail` it surfaces as 141; standard Unix behavior, not a Mago bug. I'll fix the PR description to reflect that rather than implying Mago mis-reports status.

mvanhorn:
Appreciate you taking the SIGPIPE fix too, @azjezz - piped mago output dies quietly now instead of panicking.

PR #2101, opened 2026-07-17
Fine-grained level config for disallowed functions/type instantiations linter rules

adam-ja:
## 📌 What Does This PR Do?

Adds an optional `level` attribute to the object variants of the `disallowed-functions` and `disallowed-type-instantiations` linter rule entries, so that those entries can be flagged with a different severity from the rule default.

## 🔍 Context & Motivation

Some "disallowed" functions or type instantiations might be more or less undesirable than others. This allows more fine-grained control over the severity of each entry, falling back to the level at the top of the rule config if one is not provided for a given entry.

## 🛠️ Summary of Changes

- **Feature:** Added optional `level` to the `disallowed-functions` and `disallowed-type-instantiations` entry config objects.

## 📂 Affected Areas

- [x] Linter

### DIFF
--- a/crates/linter/src/rule/security/disallowed_functions.rs
+++ b/crates/linter/src/rule/security/disallowed_functions.rs
@@ -25,18 +25,21 @@ pub struct DisallowedFunctionsRule {
     cfg: DisallowedFunctionsConfig,
 }
 
-/// An entry that can be either a simple string or an object with name and optional help.
+/// An entry that can be either a simple string or an object with name and optional help and level override.
 #[derive(Debug, Clone, Eq, PartialEq, JsonSchema)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(untagged))]
 pub enum DisallowedEntry {
     /// Simple string entry (just the name).
     Simple(String),
-    /// Entry with name and optional help message.
-    WithHelp {
+    /// Entry with name and optional help message and level override.
+    Advanced {
         name: String,
         #[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "Option::is_none"))]
         help: Option<String>,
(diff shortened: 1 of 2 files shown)


crates/linter/src/rule/security/disallowed_functions.rs
```diff
@@ -155,7 +170,9 @@ impl LintRule for DisallowedFunctionsRule {
                 "Use an alternative function or update your configuration if this restriction is no longer needed.",
             );
 
-            let issue = Issue::new(self.cfg.level, format!("Function `{function_name}` is disallowed."))
+            let level = entry.level().unwrap_or(self.cfg.level);
```
azjezz, line 173: I think we should make it so that the entry level can never be higher than the rule level itself, if `disallowed-functions` is configured at warning, all entries should be warning or lower, It feels weird that a specific entry can contain a level higher than the global rule level.

azjezz (changes requested):

crates/linter/src/rule/security/disallowed_functions.rs
adam-ja: Thanks for the feedback. It's your project so happy to take your lead on this if you think it's more consistent.

Personally, I'd say more flexibility is better as long as it's clearly documented. For example, a team might have a whole load of functions they want to flag as warnings or even just help, but then one or two they absolutely don't want so set those to error. If the rule level sets the ceiling, then to accomplish that they'd have to set to rule to error and add overrides for the majority of the functions. To me, less config is better so using the most common level as the default and then only needing overrides for a few exceptions (whether higher or lower) seems cleaner.

I'll be away from my laptop until Thursday but can resolve this either way when I'm back, unless you want to merge it sooner in which case it's all yours 🙂

azjezz: Make sense, as someone who does not use this rule a lot, I think this is fine, we will see if people complain later :D

Issue #2114, opened 2026-07-21
Ingore abstract class in guard structural with must-be-final


Norbytus:
### Feature

Skip abstract class for guard structural if must-be-final = true

### Example

```php
<?php
namespace Entity {
    abstract class A
    {
    }
}

/// TOML
/// [[guard.structural.rules]]
/// on = "**\\Entity\\**"
/// target = "class"
/// must-be-final = true
```

PR #2115, opened 2026-07-21
feat(guard): Complete final class guard

Norbytus:
## 📌 What Does This PR Do?

In guard structural check class is abstract or not, If its abstract ignore must-be-final

## 🔍 Context & Motivation

Abstract class cannot be final 

## 🛠️ Summary of Changes

- **Feature:** Ignore abstract class on must-be-final config

## 📂 Affected Areas

- [X] Guard

## 🔗 Related Issues or PRs

https://github.com/carthage-software/mago/issues/2114

### DIFF
--- a/crates/guard/src/structural/mod.rs
+++ b/crates/guard/src/structural/mod.rs
@@ -133,9 +133,10 @@ impl<'ast, 'ctx, 'arena> MutWalker<'ast, 'arena, GuardContext<'ctx, 'arena>> for
                     .get_class(fqn)
                     .map(|c| c.flags.is_final())
                     .unwrap_or(class.modifiers.contains_final());
+                let is_abstract = class.modifiers.contains_abstract();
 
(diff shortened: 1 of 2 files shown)


crates/guard/src/structural/mod.rs
```diff
@@ -144,7 +145,7 @@ impl<'ast, 'ctx, 'arena> MutWalker<'ast, 'arena, GuardContext<'ctx, 'arena>> for
                             reason: structural_rule.reason.clone(),
                         });
                     }
-                    (false, true) => {
+                    (false, true, true) => {
```
dotdash: This would disable the check on non-abstract classes.

```suggestion
                    (false, true, _) => {
```

Norbytus: Done

dotdash:
LGTM, but please sign your commits

PR #2122, opened 2026-07-22
fix(analyzer): preserve array_map shape for untyped closures

dotdash:
## 📌 What Does This PR Do?

Keeps `array_map()` from widening a sealed, only-optional array shape
(e.g. `array{foo?: string}`) when the callback is an untyped closure.

## 🔍 Context & Motivation

The `array_map` return-type provider only reads a callback's *declared*
return type. An untyped arrow function like `fn($v) => $v` has none, so
the provider bails and `array_map` falls back to its stub signature. That
widens the shape to `array<string('foo'), string>|list<string>`. The
`list<...>` part is bogus — an only-optional sealed shape can never carry
integer keys, so it is never a list — and it then clashes with a declared
`array<string, ...>` return type, producing a spurious error.

The analyzer has in fact already inferred a return type for the closure
and recorded it on the callback expression's callable signature (e.g.
`(closure(mixed): string)`). Fall back to that inferred type so the
precise shape is preserved.

## 🛠️ Summary of Changes

- **Bug Fix:** `array_map()` preserves array shapes when the callback is
  an untyped closure/arrow function.

## 📂 Affected Areas

- [x] Other: analyzer

### DIFF
--- a/crates/analyzer/tests/cases/arrays_array_map_optional_keys_untyped_closure.php
+++ b/crates/analyzer/tests/cases/arrays_array_map_optional_keys_untyped_closure.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * `array_map` must preserve the sealed shape rather than widen it to include
+ * a `list<string>` that would clash with the declared return type.
+ *
+ * @param array{foo?: string} $x
+ * @return array<string, string>
+ */
+function map_optional_shape(array $x): array
+{
+    return array_map(fn($value) => $value, $x);
+}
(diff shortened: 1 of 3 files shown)


crates/analyzer/tests/cases/arrays_array_map_optional_keys_untyped_closure.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * `array_map` must preserve the sealed shape rather than widen it to include
+ * a `list<string>` that would clash with the declared return type.
+ *
+ * @param array{foo?: string} $x
+ * @return array<string, string>
```
azjezz, line 10: we could narrow this to `array{foo?: string}`

dotdash: True, I just copied the signature from a method in our project where the thing that triggered it was an internal value, not a param.

PR #2130, opened 2026-07-23
feat(cli): add `--stats` shorthand for count output

epic-64:
## Summary

Adds a `--stats` flag to the shared reporting options, so `mago lint --stats` (and `analyze` / `guard` / `cst`) shows a summary of issue counts per code, ordered from highest to lowest, instead of listing individual issues.

```console
$ mago lint --stats
help[constant-condition]: 3
warning[strict-types]: 1
```

## Motivation

When adopting Mago on an existing codebase (or after enabling new rules, or running `--pedantic`), a run can produce thousands of individual issues -- too many to read one by one. The first thing you actually need is the shape of the problem: which rules fire and how often. That tells you what to fix first, which rules to disable or downgrade, and what belongs in a baseline.

Mago can already produce this summary via `--reporting-format code-count`, but it is easy to miss as one of many format values. A dedicated `--stats` flag makes the workflow discoverable directly in `--help`, matches what other linters offer as a first-class option, and is quicker to type and remember.

## Implementation

`--stats` is a thin shorthand: when set, the issue processor selects `ReportingFormat::CodeCount`. No new output code paths -- baseline filtering, `--minimum-report-level`, `--retain-code`, reporting targets, and exit-code semantics all behave exactly like any other report.

- `--stats` conflicts with `--fix` and with an explicit `--reporting-format`.
- Added `stats` to the `--explain` / `--list-rules` conflict lists in `mago lint`.
- Documented the flag on the shared reporting options page (en/fr/zh) and in the lint command docs.

## Testing

- New integration tests in `tests/lint_stats.rs` run the binary against a temp workspace and cover: output shape and descending count order, byte-for-byte equivalence with `--reporting-format code-count`, `--minimum-fail-level` exit-code semantics, and the clap conflicts with `--fix` / `--reporting-format`.
- `cargo build` / `cargo clippy` clean.

### DIFF
--- a/docs/content/en/fundamentals/shared-reporting-options.md
+++ b/docs/content/en/fundamentals/shared-reporting-options.md
@@ -28,6 +28,7 @@ How Mago presents the issues it finds.
 | Flag | Description |
 | :--- | :--- |
 | `--sort` | Sort reported issues by level, then code, then location. |
+| `--stats` | Show a summary of issue counts per code, ordered from highest to lowest, instead of individual issues. Shorthand for `--reporting-format code-count`. |
 | `--reporting-target <TARGET>` | Where to write the report. Values: `stdout` (default), `stderr`. |
 | `--reporting-format <FORMAT>` | Output format. See below; defaults to auto-detected. |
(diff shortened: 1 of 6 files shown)


epic-64:
Sorry, last time I forgot to run cargo fmt and it only failed in the mac-os build/test job. Added a commit with the fix, but from another PC where I didn't have signing enabled yet. Force pushed for clean commit history

PR #2146, opened 2026-07-28
feat(codex): record attributes on property metadata

cmukanisa:
Class-likes, function-likes, parameters, constants and enum cases already keep
their attributes in the codex. Properties were the only symbol kind that did
not, so nothing downstream can reason about attribute-mapped properties.

Two shapes needed care, both matching PHP semantics:

- a grouped declaration (`public int $a, $b;`) gives every item the attributes
  of the shared declaration;
- a promoted constructor parameter passes its attributes to the property it
  creates. `scan_promoted_property` was dropping them — the synthesized
  `PropertyMetadata` had none even though the parameter metadata carried them.

The test covers the three shapes in the existing scanner harness.

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`
and `cargo test --workspace --locked --all-targets` all pass locally.

### DIFF
--- a/crates/codex/src/metadata/property.rs
+++ b/crates/codex/src/metadata/property.rs
@@ -4,6 +4,7 @@ use mago_php_version::PHPVersionRange;
 use mago_span::Span;
 use mago_word::WordMap;
 
+use crate::metadata::attribute::AttributeMetadata;
 use crate::metadata::flags::MetadataFlags;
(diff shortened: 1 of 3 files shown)


azjezz:
@cmukanisa a rebase is needed :)

azjezz:
@cmukanisa a rebase is needed.

PR #2168, opened 2026-08-04
fix(analyzer): resolve `self` in attributes on class-likes

### LINKED ISSUE #2166: Regression(analyzer): no-value, self-outside-class-scope
### Description

This worked previously and is valid PHP

### Playground Link

https://mago.carthage.software/main/en/playground/#019fca12-52a8-4430-33c1-275d69a7829b

### Reproduction

_No response_

### Environment

main branch

dotdash:
## 📌 What Does This PR Do?

Fixes false positives on `self::CONST` and `parent::CONST` inside
attribute arguments on a class, interface, trait, enum, or anonymous
class, and closes two false negatives that surface once those
references resolve.

## 🔍 Context & Motivation

PHP evaluates the arguments of an attribute on a class-like in the
scope of that class-like, so `#[Route(self::ROUTE_NAME)]` on a class
is valid. We analyze those arguments with the block context of the
enclosing scope, which has no class attached, so any file-level class
reports `self-outside-class-scope` and the argument degrades to
`never` — a hard error on correct code. Members are unaffected,
because the class-scoped context is built before descending into them;
only the attributes on the declaration itself are analyzed too early.

Attaching the class-like to that scope is necessary but not
sufficient, because the arguments are still evaluated *outside* its
body. Reflection instantiates the attribute from no class scope at
all, so a non-public `__construct` is unreachable even when the
annotated declaration is the attribute class itself; and `self::CONST`
on a trait resolves the trait directly, which PHP rejects at runtime.

Correcting the constructor case means visibility must be checkable
from deliberately-no-class, so `check_method_visibility` now takes the
calling class rather than a whole `BlockContext`. That also repairs a
pre-existing miss: an attribute with a non-public constructor used on
a *member* of its own class was silently accepted in every position
(class constant, property, method, parameter).

Regressions are pinned in both directions — `self::CONST` inside a
trait *method* body stays clean, and `static::` in attribute arguments
is still rejected by the semantics pass.

## 🛠️ Summary of Changes

- **Bug Fix:** Resolve `self`/`parent` in attribute arguments on
  class-likes against the annotated declaration, and report the
  non-public-constructor and direct-trait-constant accesses that
  become visible once they do.

## 📂 Affected Areas

## 🔗 Related Issues or PRs

Fixes #2166.

### DIFF
--- a/crates/analyzer/src/context/block_flags.rs
+++ b/crates/analyzer/src/context/block_flags.rs
@@ -21,6 +21,10 @@ impl BlockContextFlags {
     pub const COLLECT_INITIALIZATIONS: u32 = 1 << 15;
     pub const CALLS_PARENT_CONSTRUCTOR: u32 = 1 << 16;
     pub const INSIDE_PIPE_CALLABLE: u32 = 1 << 17;
+    /// Set while analyzing the arguments of an attribute on a class-like declaration. The
+    /// class-like is in scope for name resolution, but the arguments are evaluated outside
+    /// its body.
+    pub const INSIDE_CLASS_LIKE_ATTRIBUTE: u32 = 1 << 18;
 
     #[inline]
     pub const fn new() -> Self {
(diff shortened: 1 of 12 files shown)


dotdash:
Merging, CI error is pre-existing

PR #2191, opened 2026-08-10
feat(analyzer): add --skip-ignores to disable configured ignores

dotdash:
## 📌 What Does This PR Do?

Adds a `--skip-ignores` flag to `mago analyze` that disables the
`ignore` list from `[analyzer]` for a single run.

## 🔍 Context & Motivation

The `ignore` list is a permanent configuration setting, so auditing what
it currently hides means editing `mago.toml` and putting it back
afterwards. That is awkward in CI and when reviewing an inherited
configuration.

Inline suppressions and the baseline are deliberately left alone —
`--ignore-baseline` already covers the latter.

## 🛠️ Summary of Changes

- **Feature:** `mago analyze --skip-ignores` bypasses the configured
  `ignore` entries, in both one-shot and watch mode.
- **Docs:** Documented the flag in the analyzer command and
  configuration references (en/fr/zh).

## 📂 Affected Areas

- [x] CLI
- [x] Documentation


dotdash:
Not merging on my own as it adds CLI surface. Are you fine with this @azjezz?

azjezz:
@dotdash LGTM, but shouldn't we also have this for guard/lint? for consistency.

dotdash:
I didn't know they had ignore rules?

azjezz:
Linter has `exclude`, which i guess is different 🤔 NVM me. 

LGTM. 🚢 it

PR #2192, opened 2026-08-11
feat: worker extensions

### LINKED ISSUE #39: feature request: support gritql for writing plugins
https://biomejs.dev/reference/gritql/
https://docs.grit.io/language/syntax

GritQL is a querying tool for source code that can be used to refactor, lint, or format the input.

It seems like Biome is looking to use it to allow plugins to be written in Grit instead of just in Rust. Given the overlap between Biome and Mag (formatting and linting code), I thought this might be worth looking at doing in Mago as well?

### LINKED ISSUE #177: feature: allow creating custom linting rules using PHP.

### LINKED ISSUE #443: feature: phpunit support
### 🚀 Describe the Feature

Add comprehensive PHPUnit testing framework support to Mago's `analyze` command, providing the same level of type inference, mock object analysis, and assertion handling capabilities as the official `phpstan/phpstan-phpunit` PHPStan extension.

While Mago currently has basic PHPUnit support in the linter (assertion method recognition and some style rules), it lacks the advanced static analysis capabilities that modern PHP developers expect when working with PHPUnit tests. The analyzer should provide full type inference for mock objects, understand assertion type narrowing, handle PHPUnit-specific patterns, and offer comprehensive testing framework integration.

### 1. Mock Object Type Inference

**Intersection Types for Mock Objects:**
- `createMock($className)` should return `MockObject&$className` intersection type
- `createStub($className)` should return `Stub&$className` intersection type
- `createConfiguredMock($className)` should return `MockObject&$className`
- `createPartialMock($className, $methods)` should return `MockObject&$className`
- `getMockForAbstractClass($className)` should return `MockObject&$className`

**MockBuilder Chain Support:**
- `getMockBuilder($className)` should return `MockBuilder<$className>`
- MockBuilder method calls should maintain the type through the chain
- `getMock()`, `getMockForAbstractClass()`, `getMockForTrait()` on MockBuilder should return proper intersection types

**PHPDoc Intersection Type Support:**
- Convert `Foo|MockObject` in PHPDoc to intersection type `Foo&MockObject`
- Support both legacy union and modern intersection syntax
- Handle complex scenarios with multiple interfaces

### 2. Assertion Type Specifying

**Type Narrowing After Assertions:**
- `assertInstanceOf($class, $object)` should narrow `$object` to `$class` type
- `assertTrue($condition)` should apply condition as type guard
- `assertFalse($condition)` should apply negated condition as type guard
- `assertNull($value)` should narrow `$value` to `null`
- `assertNotNull($value)` should exclude `null` from `$value` type
- `assertSame($expected, $actual)` should handle type narrowing
- `assertCount($expected, $array)` should validate countable types

**Type Checking for Assert Parameters:**
- `assertIsArray($value)` should narrow to `array` type
- `assertIsString($value)` should narrow to `string` type
- `assertIsBool($value)` should narrow to `bool` type
- All `assertIs*()` and `assertIsNot*()` methods with proper type narrowing
- `assertInternalType($type, $value)` with type string validation

**Container and Collection Assertions:**
- `assertArrayHasKey($key, $array)` should validate array access
- `assertContains($needle, $haystack)` with proper type validation
- `assertContainsOnly($type, $array)` should narrow array element types
- `assertContainsOnlyInstancesOf($class, $array)` should narrow to `$class[]`

### 3. Test Structure Understanding

**TestCase Recognition:**
- Identify classes extending `PHPUnit\Framework\TestCase`
- Understand test method patterns (`test*()` methods, `@test` annotation)
- Recognize `setUp()`, `tearDown()`, `setUpBeforeClass()`, `tearDownAfterClass()`
- Handle data provider methods and `@dataProvider` annotations

**Early Terminating Methods:**
- Recognize `fail()`, `markTestIncomplete()`, `markTestSkipped()` as early terminating
- Prevent "undefined variable" errors after these calls
- Handle control flow properly in test methods

**Exception Handling:**
- Support `expectException()`, `expectExceptionMessage()`, `expectExceptionCode()`
- Understand `@expectedException` annotations
- Validate exception types and messages

### 4. Advanced PHPUnit Features

**Annotation Support:**
- `@covers` and `@coversDefaultClass` validation
- `@uses`, `@group`, `@testWith` annotations
- Data provider method validation and type checking
- `@runInSeparateProcess`, `@preserveGlobalState` handling

**Constraint System:**
- Support PHPUnit constraint objects in assertions
- Type checking for constraint parameters
- Custom constraint class recognition

## 5. Stub Files Integration

Add PHPUnit stub files similar to phpstan-phpunit:
- TestCase method signatures with proper generics
- MockObject and Stub interface definitions
- Constraint class definitions
- Assert static method signatures

## Scenario 1: Mock Object Type Inference

```php
<?php

use PHPUnit\Framework\TestCase;
use PHPUnit\Framework\MockObject\MockObject;

class UserServiceTest extends TestCase
{
    public function testCreateUser(): void
    {
        // Mago should understand $userRepo is UserRepository&MockObject
        $userRepo = $this->createMock(UserRepository::class);

        // Both MockObject methods and UserRepository methods should be available
        $userRepo->expects($this->once())  // MockObject method
                 ->method('save')          // UserRepository method
                 ->with($this->isInstanceOf(User::class))
                 ->willReturn(true);

        $service = new UserService($userRepo);
        $result = $service->createUser('John Doe');

        $this->assertTrue($result);
    }

    /**
     * @return UserRepository&MockObject
     */
    private function createUserRepositoryMock(): UserRepository
    {
        // Mago should understand the PHPDoc intersection type syntax
        return $this->createMock(UserRepository::class);
    }
}
```

## Scenario 2: Assertion Type Narrowing

```php
<?php

class ValidationTest extends TestCase
{
    public function testTypeNarrowing(): void
    {
        $value = $this->getUnknownValue(); // mixed type

        // After this assertion, Mago should know $value is string
        $this->assertIsString($value);
        $length = strlen($value); // No type error here

        $user = $this->getUser(); // User|null

        // After this assertion, Mago should exclude null
        $this->assertNotNull($user);
        $name = $user->getName(); // No "call on null" error

        $data = $this->getData(); // mixed

        // After this assertion, Mago should know $data is array
        $this->assertIsArray($data);
        $this->assertArrayHasKey('id', $data);
        $id = $data['id']; // Array access should be valid
    }

    public function testInstanceOfNarrowing($object): void
    {
        // $object parameter has mixed type

        $this->assertInstanceOf(User::class, $object);

        // After assertion, Mago should know $object is User
        $name = $object->getName(); // Should not error
        $object->setEmail('test@example.com'); // User method should be available
    }
}
```

## Scenario 3: Complex Mock Builder Usage

```php
<?php

class PaymentProcessorTest extends TestCase
{
    public function testProcessPaymentWithBuilder(): void
    {
        // Mago should understand the MockBuilder chain
        $gateway = $this->getMockBuilder(PaymentGateway::class)
            ->setMethods(['processPayment', 'validateCard'])
            ->disableOriginalConstructor()
            ->getMock();

        // Should understand $gateway is PaymentGateway&MockObject
        $gateway->expects($this->once())
                ->method('validateCard')
                ->with($this->isInstanceOf(CreditCard::class))
                ->willReturn(true);

        $gateway->expects($this->once())
                ->method('processPayment')
                ->willReturn(new PaymentResult(true, 'tx123'));

        $processor = new PaymentProcessor($gateway);
        $result = $processor->process($card, $amount);

        $this->assertInstanceOf(PaymentResult::class, $result);
        $this->assertTrue($result->isSuccessful());
    }
}
```

## Scenario 4: Data Provider Type Checking

```php
<?php

class CalculatorTest extends TestCase
{
    /**
     * @dataProvider additionProvider
     */
    public function testAddition(int $a, int $b, int $expected): void
    {
        // Mago should validate that data provider returns correct types
        $result = Calculator::add($a, $b);
        $this->assertSame($expected, $result);
    }

    /**
     * @return array<array{int, int, int}>
     */
    public function additionProvider(): array
    {
        // Mago should validate this matches the test method signature
        return [
            [1, 2, 3],
            [0, 0, 0],
            [-1, 1, 0],
        ];
    }
}
```

## Benefits for Developers

1. **Accurate Mock Object Types**: No more false positives on mock method calls
2. **Smart Assertion Analysis**: Type narrowing after successful assertions
3. **Better Test Coverage**: Catch actual type errors in test code
4. **IDE Integration**: Proper autocompletion and navigation in test files
5. **Framework Migration**: Easier transition from PHPStan to Mago
6. **Test Quality**: Improve test code quality through static analysis

### 💡 Proposed Solution

_No response_

### 📌 Priority

Must have

### 📝 Additional Context

PHPUnit is the most widely used testing framework in the PHP ecosystem. Comprehensive PHPUnit support is essential for Mago to be considered a viable alternative to PHPStan. Many PHP projects rely heavily on PHPUnit tests, and proper static analysis of test code is crucial for maintaining code quality and catching bugs early.

## Reference Implementation

The official PHPStan PHPUnit extension provides the exact functionality needed:
- **Repository**: https://github.com/phpstan/phpstan-phpunit
- **Key Components**:
  - Mock object intersection type handling
  - Assertion type specifying extensions
  - PHPDoc type node resolver
  - Comprehensive stub files
  - Framework-specific rules

## Current Mago PHPUnit Support

Mago already has basic PHPUnit support:
- **Linter Rules**: `crates/linter/src/rule/correctness/strict_assertions.rs`
- **PHPUnit Utils**: `crates/linter/src/rule/utils/phpunit.rs` with 173+ assertion methods
- **Method Recognition**: Can identify PHPUnit assertion and testing methods

**What's Missing:**
- No type inference for mock objects
- No assertion type narrowing
- No intersection type support
- No MockBuilder chain analysis
- No PHPDoc intersection type conversion

## Community Impact

**Market Position:**
- PHPUnit support is table stakes for PHP static analyzers
- This would remove a major barrier to Mago adoption
- Essential for enterprise PHP projects that rely on comprehensive test suites

**Migration Benefits:**
- Teams using PHPStan + phpstan-phpunit could switch to Mago seamlessly
- Provides path from PHPStan to Mago for testing-heavy projects
- Demonstrates Mago's commitment to full PHP ecosystem support

### LINKED ISSUE #848: Define custom rule
### Feature

Would it be possible for Mago to support the definition of custom rules in PHP code the same way PHPStan does?

### Example

```php

```

### LINKED ISSUE #887: Laravel trans('path.to.translation') function returns string or array making whole project messed
### Feature

Laravel's trans('path.to.translation') function returns string or array and this causes analyzer to report hundreds of warnings of type `array-to-string-conversion`. 

The only fix for this is wrapping all trans('...') into `if (!is_array(trans(...))) { ... }` which is actually a bad idea and it is not realistic to fix all of these warnings. Or I need to add `@mago-ignore` for each trans() call and do more mess than it is acceptable. 
So in result I will disable `array-to-string-conversion` and focus on more real problems, but this way I might loose some real problems with `array-to-string-conversion`.

We need some way to ignore specific functions or a way to set function return type as primary, so that analyzer will treat string as default type returned, while array access will get a warning. 

### Example

```php
function getTranslation(): string {
  // No warning.
  return trans('user.role.admin');
}

function getRoleLables(): array {
  // Warning: string returned as array.
  return trans('user.role');
}
```

### LINKED ISSUE #1385: PHPUnit data provider argument count and type validation
### Feature

Mago should validate that PHPUnit data providers return data that matches the test method signature, both in argument count and types.

Currently, mismatches between a data provider's yielded/returned arrays and the test method's parameters are only caught at runtime when PHPUnit executes the test. 

PHPStan's [phpstan-phpunit](https://github.com/phpstan/phpstan-phpunit) extension implements this via `DataProviderDataRule`

### Example

```php
class MyTest extends TestCase
{
    #[DataProvider('provideData')]
    public function testSomething(int $a, string $b): void
    {
        // ...
    }

    public static function provideData(): iterable
    {
        yield [1];              // too few arguments (1 instead of 2)
        yield [1, 'foo', true]; // too many arguments (3 instead of 2)
        yield [1, 2];           // wrong type for $b (int instead of string)
        yield [1, 'bar'];       // correct
    }
}
```

### LINKED ISSUE #1883: Request for Mago static analysis extendability
### Feature

Hi, I would like Mago to provide a way to extend static type analysis.
All in all analysis works great in Mago, but there are some things that are impossible to do in Laravel:
* Check if Collection->pluck('property.sub_property') points to an existing property, and to statically check it's type
* Similarily as above, Collection->value
* When using Laravel Spatie Data, I cannot check if some DTO::from(['property'=>'value']) is correct

In PHPStan, I'm using my own extensions for checking that, it would be nice to have that possibility also in Mago.

### Example

```php

```

### LINKED ISSUE #2164: Five small additions that would let a plugin implement PHPStan-style rules
Following the "discuss the changes first" step in CONTRIBUTING, before proposing anything.

Someone asked whether PHPStan's custom rules could be compiled to run on Mago. To find out I built a transpiler that reads a PHPStan rule's PHP source and emits a Mago analyzer plugin, and a harness that runs the same corpus through both tools and compares findings.

Twenty six rules and a collector now translate, and for those rules **the two tools produce the same output**: same findings, same lines, same message text, same identifiers. Verified two ways, on annotated fixtures (145 assertions) and on 921 files of real vendor code where 25 of the 26 agree exactly. The one that does not is a difference between the two engines' loop scoping rather than a translation error.

I am not asking you to take the transpiler. It is research and it stays out of tree. What I would like to ask about is the five small things the analyzer would not let a plugin do. Each one stopped a real rule, and I have them as five commits I can open as a PR if you want them.

## The five

1. **Expose the analysed file.** A plugin can reach the codebase, the scope and every expression's type, but not the file it is looking at, because `source_file` is `pub(crate)`. Anything of the form "only in tests", "skip generated code" or "only under this directory" is unwritable outside the crate. Two accessors, no behaviour change.

2. **Hooks for class-like member declarations.** The declaration hooks cover the class-likes but nothing fires for the members inside them, so a rule about a method, property, constant or enum case *declaration* has nothing to attach to. One `ClassLikeMemberHook` firing from the member loop in `analyze_class_like` covers all four class-like kinds.

3. **A hook that runs once after the whole project.** Every other hook fires per file, which makes collector style rules impossible. Accumulating across files already works and cross-file annotations are already attributed correctly, so the only missing piece was the callback.

4. **Resolve a written class name.** A plugin sees `extends Foo` as written and cannot tell which `Foo` the file imported, so "this class must not extend that one" only works when the source spells the name out. The obvious fix would add a third lifetime parameter to every hook signature in the crate, so instead a small `NameLookup` trait erases it, since the lookup is by byte offset and carries no borrow from the arena.

5. **Let a plugin choose its own issue code.** With `IssueCode` closed, every rule a plugin ships reports as the same variant, so they cannot be told apart by anything except message text, which is what a baseline or an editor integration keys off.

No enum change turned out to be needed for that last one. `Issue` already carries `code: Option<String>` and the collector's `report_with_code` already takes `impl Into<String>`, so a code the plugin set is simply kept and the enum argument becomes the fallback. Nothing sets a code today, so no existing caller changes.

## Where this overlaps with your own work

I know about #1663 and #39, and that you said custom rules are not a priority right now. Nothing here is an argument with that. These five are useful to any plugin, not only to transpiled ones, and they are deliberately not a rule language. If the direction is wrong or the timing is wrong, saying so costs you nothing and I will not open the PR.

One thing I found that may be worth knowing regardless: **the analyzer skips the body of any class whose parent or interface it cannot resolve**, members and method bodies both, so no hook fires anywhere inside it. On a corpus where two inherited symbols were unreadable, that silently cost 10 of 169 findings with no warning. Whether that is intended I cannot tell from outside.

## A second, unrelated thing

While comparing output I found `--reporting-format=phpstan` worth having: a formatter that emits the shape PHPStan's `--error-format=json` produces. CI annotators, review bots, editor integrations and baseline generators already read that shape, so a project moving analysis here keeps its tooling instead of replacing it. That is one commit, independent of the five, and it sorts its own output so it does not depend on the ordering fix I opened separately.

## Branches

Both are on top of `1ec88688e`, each standalone, and `cargo fmt --all -- --check`, `cargo clippy --workspace --all-features -- -D warnings` and `cargo test --workspace --locked --all-targets` are clean at every commit:

- the five API commits: https://github.com/SanderMuller/mago/compare/main...SanderMuller:mago:plugin-api-additions
- the formatter: https://github.com/SanderMuller/mago/compare/main...SanderMuller:mago:phpstan-output-format

Happy to reshape any of it, or to drop it. I would rather ask first than send five commits of new public API at you unannounced.

Related: the ordering fix from the same work is #2163, opened separately because it is a plain bug fix and independent of everything above.

azjezz:
Sponsored by [check24](https://check24.de) ❤️ 

More details soon :)

closes #177 
closes #1883 
closes #848
closes #2164
closes #1663 
closes #39 
closes #887 ( should be an external `laravel` extension )
closes #1261 ( should be an external `laravel` extension )
closes #443 ( should be an external `phpunit` extension )
closes #1385 ( should be an external `phpunit` extension )


check24-opensource:
Looking forward to this  🙌

Ocramius (commented):
Just scanned through the patch: a high level interaction diagram would be nice here 😅

SanderMuller:
I've been trying to run PHPStan rules on Mago (background in #2164), so I ported a real rule set to this branch to see how far the SDK gets. Twelve rules, some from `symplify/phpstan-rules` and the rest ours, all written against `Mago\Sdk` as it stands here.

## It works

Eleven of the twelve produce exactly what PHPStan produces, same messages and same positions, on two corpora: fixtures written per rule, and 921 files of real vendor code (doctrine, symfony, phpunit).

The twelfth differs, and it's PHPStan being odd rather than the SDK. For a rule firing inside a trait, PHPStan reports once per using class, so 17 findings across 10 sites. The SDK version reports once per site. Fold the duplicates and the sets are equal.

Findings are stable across pool sizes: the same 31 findings and the same sha256 of sorted output at 1, 4, 8 and adaptive workers. That's where I expected trouble if a rule accidentally held per-worker state.

`LintContext::getResolvedName` is what convinced me the API is sufficient rather than just usable. A rule matching `extends Doctrine\ORM\EntityRepository` has to see through the file's imports, and a version comparing written text passes the fully qualified fixture while failing the imported one. The analyzer plugin API couldn't do that at all.

## Cost

On the 921 files, against PHPStan configured with the same twelve rules and nothing else:

- `mago lint`, no extension: 0.172s CPU
- with the twelve PHP rules: 0.239s CPU at one worker, 0.277s adaptive, 0.491s at eight
- PHPStan, same twelve rules, cold: 30.5s CPU, 4.8s wall

So the rules add roughly 40% to 190% on top of a bare lint run depending on pool size, and land 62x to 128x cheaper than PHPStan on CPU. Wall clock is 0.359s at one worker and then flat around 0.08s from two workers up, while CPU keeps climbing, which reads like per-worker startup and per-file protocol cost rather than the rules themselves. Two to four workers was the sweet spot. This is one corpus on one machine.

## Three things that cost me time

A wrong `NodeKind` in `targets` fails silently. I wrote `NodeKind::Class`, which is a valid case, but it's the `::class` constant expression rather than a class declaration, so the rule never fired and nothing in the output suggested the target was nonsense. The case I wanted was `Class_`. A warning for a rule that subscribes to a kind it never matches would have saved me a while.

Output written before `run()` corrupts the protocol. `Worker::run()` installs an output buffer guard, which is right, but extension construction happens before that. A notice or a stray `var_dump` in a rule constructor goes into the frame stream, and the error that comes back points nowhere near the cause. Moving the guard earlier, or wrapping construction, would make that diagnosable.

There's no `composer.json` for the SDK on this branch, so I hand rolled an autoloader for `Mago\Sdk\` and pulled in `revolt/event-loop` separately. Fine for an experiment, but a package would turn writing an extension into a `composer require`.

## What can't port yet

Twelve of the twenty six rules I care about work. The other fourteen need one of two things.

Type inference, for anything asking what a variable actually is. Expected at this tier, and I'm not asking for it.

Cross file aggregation is the one I'm hoping for. PHPStan calls these collectors: every file contributes facts during the parallel pass, then a rule runs once over everything collected. It's how you find an unused public method, or a class nobody instantiates, without holding the whole project in one process. Nothing in the SDK lets a rule contribute a fact for a later pass to read, and with a worker pool there's a real question about where the aggregate lives and when the final pass runs.

If that's on your roadmap I'd be glad to help, either writing the PHP side against a shape you pick, or porting the fourteen rules as a test case for whatever you land on. The harness that produced the agreement numbers above is reusable for that.

Thanks for building this. Being able to write a rule in PHP and have it agree with PHPStan character for character on real code is further than I expected to get.

azjezz:
@SanderMuller Linter and Analyzer are different parts, we are in the progress of adding analyzer support to workers. for now, its linter only - which operates on CST alone.

SanderMuller:
Good to hear on analyzer workers, that covers a chunk of what I listed.

I probably didn't separate the two gaps clearly enough. Analyzer workers cover the types half. The other half is cross file aggregation, which the analyzer tier wouldn't cover on its own.

A rule that reports an unused public method needs every file to contribute "there is a call to X here" during the parallel pass, then one pass at the end over everything collected. The split is visible in my own set: one of the blocked rules calls no navigation helpers at all, only the three that read collected data, and the collector feeding it reads property type hints and nothing else. Neither touches inference, so types wouldn't unblock either of them. What's missing is somewhere to put a fact and a hook that runs once all files are done.

If the worker protocol is growing a second phase for the analyzer anyway, that might be the moment to consider letting a rule emit something other than an issue. Happy to prototype against whatever shape you want.

azjezz:
> A rule that reports an unused public method needs every file to contribute "there is a call to X here" 

Mago already calculates this information on its own and will send the final analysis results to the worker, which can inspect references and other things.

SanderMuller:
That's better than what I was asking for, so I built it instead of guessing.

I rewrote two rules from `tomasvotruba/unused-public`, the unused public method and property ones, against your shapes: a PHP mirror of `SymbolReferences` plus a stand-in for the declarations in `codex`. The rules see nothing else. There's no analyzer worker yet, so PHPStan's collectors fill those two structures and the rules get called the way a worker would call them.

On 965 files they match the originals exactly. 617 method findings, 22 property, nothing on either side alone. They also shrink, 407 tokens to 130, since the originals do case-insensitive string matching on `"Class::method"` text and a keyed graph turns that into a lookup.

This isn't a reimplementation of `unused_members.rs`, which returns early on public members and only covers protected ones in final classes. The public case is the one that needs the reference graph.

Two things I couldn't see from reading.

The rules need declarations as well as references, so `AnalysisResult` alone isn't enough. The declaration half is `codex`, and it looks complete for this, including keeping which class declares a method separate from where it appears. `LintRequest` is per file today, so what I'd like to know is whether the analyzer equivalent carries the codebase index too. That decides whether the port works at all.

The remaining gap is small. One input has no counterpart anywhere: the method rule skips every public method of a class named as a Symfony form's `data_class`, and that class is never called, it's named in an array literal. No reference graph can contain it. It's a fact a rule spots while walking one file and wants back at the end, one boolean per class.

So my earlier ask was too broad. With the codebase index alongside the result, the reference-shaped rules port as they are. Somewhere to stash a per-file fact and read it back at the end covers the rest.

Happy to port these against the analyzer worker API when it lands.

azjezz:
@SanderMuller I would advice against testing against this branch yet. while the model is fixed ( i.e, we will use workers/IPC ), the protocol is not stable, the API is not stable, and we are still designing everything.

AJenbo:
this is interesting to me 😄

Issue #374, opened 2025-09-03
Warning for 'missing type hint' that cannot be typed due to inheritance


JeftaAtSiip:
### 🐞 Describe the Bug

In Laravel projects, `mago lint` gives a warning when overriding untyped properties:

```
warning[property-type]: Property `$guarded` is missing a type hint.
   ┌─ app/Models/Company.php:18:5
   │
18 │     protected $guarded = false;
   │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Property `$guarded` is declared here.
```

In this case, the `$guarded` property is defined in Laravel's `GuardsAttributes` trait which is included through the `Model` class.

```php
trait GuardsAttributes
{
    /**
     * The attributes that aren't mass assignable.
     *
     * @var array<string>|bool
     */
    protected $guarded = ['*'];
```

If you do add the typehint, php will complain at runtime: `Type of App\Models\Company::$guarded must not be defined (as in class Illuminate\Database\Eloquent\Model)`

### 🔄 Steps to Reproduce

1. Create a Laravel model with `protected $guarded = false;`
2. Run `mago lint`
3. You'll get a warning `warning[property-type]: Property `$guarded` is missing a type hint`, which is unfixable since the property is untyped in the `HasGuarded` trait

### ⚙️ Configuration (mago.toml)

```toml
php-version = "8.4.0"

[source]
paths = ["app/", "database/factories/", "database/seeders/", "tests/"]
includes = ["vendor"]
excludes = []

[linter]
integrations = ["laravel", "phpunit"]
```

### 📜 Command Output

```shell
warning[property-type]: Property `$guarded` is missing a type hint.
   ┌─ app/Models/Company.php:11:5
   │
18 │     protected $guarded = false;
   │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Property `$guarded` is declared here.
```

### 📂 PHP Code Sample (If Applicable)

```php
<?php

declare(strict_types=1);

namespace App\Models;

use Illuminate\Database\Eloquent\Model;

final class Company extends Model
{
    protected $guarded = false;
}
```

### 🖥️ Operating System

Linux

### 📦 How did you install Mago?

Prebuilt Binary (download from GitHub)

azjezz:
I think we should not trigger this if we have a @var docblock 🤔

[...]

PR #2201, opened 2026-08-14
fix(analyzer): report missing property types on items without a default

chrisvasz:
## 📌 What Does This PR Do?

Hello! I was attempting to migrate from phpstan to mago on our large closed-source codebase and I ran into a small issue with the analyzer. It seems `missing-property-type` fires only on properties that carry a default value.  `public $foo;` is never reported; `public $foo = 1;` is.

Here's a reproduction:

```php
<?php

class PropProbe
{
    public $pubNoDefault;           // not reported
    public $pubDefault = 1;         // reported
    protected $protNoDefault;       // not reported
    protected $protDefault = 1;     // reported
    private $privNoDefault;         // not reported
    private $privDefault = 1;       // reported
    public static $statNoDefault;   // not reported
    public static $statDefault = 1; // reported
    public $multiA, $multiB = 2;    // only $multiB reported
    var $legacyVar;                 // not reported

    public array $impreciseNoDefault;      // no imprecise-type
    public array $impreciseDefault = [];   // imprecise-type
}
```

This PR reports `missing-property-type` on properties without a default value.

## 🔍 Context & Motivation

I was doing some research, and it seems mago used to have a linter rule to catch this, but the linter rule was "promoted" to an analyzer rule in https://github.com/carthage-software/mago/commit/4caaa2d5a due to https://github.com/carthage-software/mago/issues/374. Makes sense -- the analyzer can apply semantic analysis to catch things like parent-class typing.

It seems though that the "promotion" from a linter rule to an analyzer rule created the possibility of false negatives on properties without defaults. This PR changes the code to use the existing `Property::variables()` helper, which already returns the variable of every item regardless of whether it has a default value.

## 🛠️ Summary of Changes

- **Bug Fix:** correctly reports `missing-property-type` on properties without a default

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): analyzer

## 📝 Notes for Reviewers

This will add warnings on lines that previously passed without being flagged.

### DIFF
--- a/crates/analyzer/src/utils/missing_type_hints.rs
+++ b/crates/analyzer/src/utils/missing_type_hints.rs
@@ -14,7 +14,6 @@ use mago_syntax::cst::FunctionLikeParameter;
 use mago_syntax::cst::FunctionLikeReturnTypeHint;
 use mago_syntax::cst::Hint;
 use mago_syntax::cst::Property;
-use mago_syntax::cst::PropertyItem;
 
 use crate::code::IssueCode;
 use crate::context::Context;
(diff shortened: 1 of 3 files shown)


azjezz:
Thank you @chrisvasz !

Issue #2136, opened 2026-07-25
Lint Fix: no-fully-qualified-global-class-like shadows implicit import from same namespace


umulmrum:
### Description

I have a class and interface setup as follows:

```php
namespace A;

interface SomeInterface
{
    public function foo(): void;
}

class X implements SomeInterface
{
    public function __construct(private \B\SomeInterface $some) {}
    
    public function foo(): void
    {
        echo 'Hello';
    }
}

namespace B;

interface SomeInterface
{
    public function foo(): void;
}
```

So there is namespace A with an interface and an implementation of this interface, and there is namespace B with another interface with the same name. Class `X` both implements `SomeInterface` from the same namespace `A` and receives an object in the constructor that implements the interface from namespace `B`.  In my system these are all organized in separate files. None of the interfaces are imported, because the interface from `A` doesn't need an import and the interface from `B` cannot be imported as the name is blocked by `A`.

When I run `mago lint --fix --only no-fully-qualified-global-class-like`, `B\SomeInterface` gets imported, shadowing the one from `A` and leading to errors.

I expect that the name conflict is recognized and no import is performed (not sure if the linter has this information though).

Thanks for helping!

### Environment

- Mago 1.44.0, 1.45.0

PR #2225, opened 2026-08-20
fix(linter): avoid unsafe class-like import fixes

marcovmun:
## 📌 What Does This PR Do?

Track name kinds and exact import scopes so auto-fixes do not change existing class-like resolutions. When a direct import conflicts, keep reporting the lint and suggest a collision-free alias.

## 🔍 Context & Motivation

This breaks our application when running `mago lint --fix` on it

## 🛠️ Summary of Changes

- **Bug Fix:** Fixed issue  #2136 
- **Feature:** Suggest alias for namespaces that cannot be automatically fixed

## 📂 Affected Areas

- [x] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 📝 Notes for Reviewers

Not my native programming language, could contain obvious mistakes.
Maybe better to reuse the tests with edge cases and do the fix yourself


crates/linter/tests/issue_2136.rs
```diff
@@ -0,0 +1,276 @@
...
+
+/// Models `A/X.php` while `SomeInterface` is declared in a separate `A/SomeInterface.php`.
+const INPUT_INTERFACE_IN_OTHER_FILE: &str = r#"<?php
+
+namespace A;
+
+class X implements SomeInterface
+{
+    public function __construct(private \B\SomeInterface $some) {}
+
```
azjezz: We already have a testing mechanism we use under the rules using macros. please do not introduce a new testing method.

azjezz (changes requested):

crates/linter/src/rule/consistency/no_fully_qualified_global_class_like.rs
```diff
@@ -885,4 +885,241 @@ mod tests {
             function at(DateTimeImmutable $when): void {}
         "#}
     }
+
+    // Regression tests for https://github.com/carthage-software/mago/issues/2136
```
azjezz: Please remove comment

azjezz (changes requested):

crates/linter/tests/issue_2136.rs
marcovmun: Sorry about that, I moved it

crates/linter/src/rule/consistency/no_fully_qualified_global_class_like.rs
marcovmun: Done

crates/names/src/lib.rs
```diff
@@ -27,6 +75,9 @@ pub struct ResolvedNames<'arena> {
     /// can return `&(&'arena [u8], bool)` references — preserving the original signature
     /// for backward compatibility.
     names: HashMap<u32, (u32, (&'arena [u8], bool))>,
+    /// Resolution metadata keyed by the same source start offset as `names`.
+    #[cfg_attr(feature = "serde", serde(skip))]
+    metadata: HashMap<u32, NameResolutionMetadata<'arena>>,
```
azjezz, line 80: Can this metadata live alongside the existing resolved-name entry instead of using a second `HashMap`? It adds another allocation and hash insertion for every resolved name globally, even when this lint rule is disabled.

crates/linter/src/context.rs
```diff
@@ -79,9 +80,121 @@ where
...
+
+            let root = metadata.root();
+            if !root.eq_ignore_ascii_case(short_name) {
+                return false;
+            }
+
+            let suffix = &original[root.len()..];
+            !resolution_matches_import(resolved, fqn, suffix)
+        })
+    }
```
azjezz, line 124: This performs a full scan plus a metadata hash lookup for every fully qualified reference, and conflict handling calls it again. So performance would be terrible. we can build a per-import-scope collision index once instead

crates/linter/src/context.rs
```diff
@@ -79,9 +80,121 @@ where
...
+    fn has_implicit_name_conflict(&self, fqn: &[u8]) -> bool {
+        let fqn = fqn.strip_prefix(b"\\").unwrap_or(fqn);
+        let short_name = fqn.rsplit(|byte| *byte == b'\\').next().unwrap_or(fqn);
+        let import_scope = self.current_import_scope();
+
+        self.resolved_names.iter().any(|(start, _, resolved, imported)| {
+            let Some(metadata) = self.resolved_names.metadata_at_offset(start) else {
+                return false;
+            };
+
```
azjezz, line 106: This only considers CST-resolved names, so PHPDoc types are not checked. Adding `use B\Foo;` can still change an existing `@param Foo` from `A\Foo` to `B\Foo`.

marcovmun: true, good catch

marcovmun:
this is little to big for me for language where I don;t have experience with.
Other solution is to mark this lint rules as potential unsafe fix? Until feature is improved without any potential behavior change? @azjezz

PR #2243, opened 2026-08-23
feat(formatter): add parentheses-in-return-statement option

### LINKED ISSUE #843: Configurable `return` parentheses
### Feature

Hello, latest Mago adds parentheses around the `return` statement depending on some internal factors. These factors do not seem to be controllable by any formatter settings. I would like to ask for a formatter setting to control this behavior. Also, this issue is similar to the last case from https://github.com/carthage-software/mago/issues/242.

### Example

Consider the last two [examples from the Playground](https://mago.carthage.software/playground#019b673e-f472-5eb1-6c70-1c88eead0b9d), which differ just by the added one more array key, here called `another`:

```php
function complex(array $array)
{
    return empty($array) ? Data::fromRow(['userId' => 0, 'download' => true, 'force' => false]) : Data::fromRow($array);
}

function complex_paren(array $array)
{
    return (
        empty($array)
            ? Data::fromRow(['userId' => 0, 'download' => true, 'force' => false, 'another' => 15.6])
            : Data::fromRow($array)
    );
}
```

Mago formats the first one as one line, and the second one is nicely expanded as multiline ternary, but with added parentheses. It would be great if Mago would allow skipping the parentheses.

bastnic:
## 📌 What Does This PR Do?

Adds a new formatter option, `parentheses-in-return-statement` (default `true`), that controls whether multi-line `return` values get wrapped in parentheses.

With `parentheses-in-return-statement = false`:

```diff
-    return (
-        empty($array)
-            ? Data::fromRow(['userId' => 0, 'download' => true, 'force' => false, 'another' => 15.6])
-            : Data::fromRow($array)
-    );
+    return empty($array)
+        ? Data::fromRow(['userId' => 0, 'download' => true, 'force' => false, 'another' => 15.6])
+        : Data::fromRow($array);
```

The default (`true`) preserves current behavior.

## 🔍 Context & Motivation

Fixes #843. When a `return` value breaks across multiple lines (typically a conditional expression), the formatter adds wrapping parentheses via `IfBreak` markers in `format_return_value()`. These factors are not controllable by any formatter setting, and users who prefer the unparenthesized style have no way to opt out.

The parentheses required for comment placement (a leading own-line comment on the return value) remain unconditional, since they are needed for correct comment rendering regardless of user preference — covered by the new `parentheses_in_return_statement_comment_disabled` fixture.

## 🛠️ Summary of Changes

- **Setting:** added `parentheses_in_return_statement: bool` to `FormatSettings` (`crates/formatter/src/settings.rs`), defaulting to `true`, exposed as `parentheses-in-return-statement` in `mago.toml`. The setting only applies to `return` statements; arrow function bodies keep their existing wrapping behavior.
- **Formatter:** `format_return_value()` now takes an `allow_wrap: bool` parameter gating the conditional parenthesization; `Return::format()` passes the setting value while arrow functions pass `true` (`crates/formatter/src/internal/format/return_value.rs`, `mod.rs`, `expression.rs`).
- **Presets:** `parentheses_in_return_statement: true` added to `DEFAULT_PRESET`.
- **Tests:** three fixtures — `parentheses_in_return_statement` (default behavior), `parentheses_in_return_statement_disabled` (issue example, unparenthesized output), and `parentheses_in_return_statement_comment_disabled` (comment still forces parentheses with option off).
- **Docs:** configuration reference updated for en/fr/zh.

## 📂 Affected Areas

- [ ] Linter
- [x] Formatter
- [ ] CLI
- [ ] Dependencies
- [x] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Closes #843.

## 📝 Notes for Reviewers

- Default settings are unaffected; this is purely an opt-out.
- Full formatter suite green and whole-workspace test suite green (8529 tests); `cargo fmt --check` and `cargo clippy -p mago-formatter --all-targets` clean.


azjezz:
This is an opinionated approach mago uses, not a bug. after all, it is an opinionated formatter. I want to avoid adding new options for this stuff, i prefer we keep it as is.

bastnic:
No worry, thanks for your feedback

PR #2284, opened 2026-08-28
fix(prelude): model how class probes treat string subjects

dotdash:
## 📌 What Does This PR Do?

Corrects the `is_a()`, `is_subclass_of()` and `method_exists()` stubs so
that probing an untrusted string for a class is no longer reported as an
error, and so that a probe which can only ever fail is recognized as
such.

## 🔍 Context & Motivation

These functions declare their subject as `object|class-string`, which
inverts the idiom: they are how you establish that a string names a
class, so demanding a `class-string` going in requires the very fact the
call is meant to determine. Reflection code that checks a string before
using it as a class name is flagged for passing a too-general argument.

The `$allow_string` flag is the part that does constrain the subject,
and it is unmodelled. With the flag off, a string subject makes the call
return false unconditionally, but we infer a plain `bool` and the dead
branch that follows goes unnoticed.

## 🛠️ Summary of Changes

- **Bug Fix:** Widened the probe subject to `object|string` and encoded
  `$allow_string` in a conditional `@return`.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [x] Other (please specify): prelude

### DIFF
--- a/crates/prelude/assets/extensions/core.php
+++ b/crates/prelude/assets/extensions/core.php
@@ -819,7 +819,7 @@ function get_called_class(): string {}
 function get_parent_class(object|string $object_or_class): string|false {}
 
 /**
- * @param object|class-string $object_or_class
+ * @param object|string $object_or_class
  *
  * @pure
  */
@@ -900,16 +900,20 @@ function get_included_files(): array {}
 function get_required_files(): array {}
 
 /**
- * @param object|class-string $object_or_class
+ * @param object|string $object_or_class
  * @param class-string $class
+ *
+ * @return ($allow_string is false ? ($object_or_class is object ? bool : false) : bool)
  *
  * @pure
  */
 function is_subclass_of(mixed $object_or_class, string $class, bool $allow_string = true): bool {}
 
 /**
- * @param object|class-string $object_or_class
+ * @param object|string $object_or_class
  * @param class-string $class
+ *
+ * @return ($allow_string is false ? ($object_or_class is object ? bool : false) : bool)
  *
  * @pure
  */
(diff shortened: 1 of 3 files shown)


crates/prelude/assets/extensions/core.php
```diff
@@ -900,17 +900,21 @@ function get_included_files(): array {}
...
 /**
- * @param object|class-string $object_or_class
+ * @param object|string $object_or_class
  * @param class-string $class
  *
+ * @return ($allow_string is false ? ($object_or_class is object ? bool : false) : bool)
+ *
  * @pure
  */
 function is_a(mixed $object_or_class, string $class, bool $allow_string = false): bool {}
```
azjezz, line 920: @dotdash shouldn't we add `@assert-if-true object|class-string $object_or_class`? so that after this call, a string is narrowed to a class-string?

dotdash: ah, good point. I had a brain fart earlier where I wanted to restrict `$object_or_class` based on `$allow_string`, and then promptly forgot about all further refinements

azjezz: we can detect that using a plugin ( an internal one, not external ).

dotdash: Actually, the narrowing is [here](https://github.com/carthage-software/mago/blob/HEAD/crates/analyzer/src/assertion.rs#L445-L445) and it's more specific than what the annotation can express (I think?)

dotdash: https://github.com/carthage-software/mago/pull/2286 adds tests for the existing narrowing

azjezz: nvm lol

Issue #2007, opened 2026-06-23
Loose variant baseline is not applied on closures


BafS:
### Description

Closures issues still have the precise line information set.

For instance:

```toml
[[issues]]
file = "apps/Foo/src/ConnectionTokenService.php"
code = "mixed-return-statement"
message = "Could not infer a precise return type for function `{closure:apps/Foo/src/ConnectionTokenService.php:113:13}`. Saw type `mixed`."
count = 1
```

It should probably be `{closure:apps/Foo/src/ConnectionTokenService.php}`.

### Environment

Mago 1.30.0

azjezz:
This is sadly not possible to fix. `{closure:apps/Foo/src/ConnectionTokenService.php}` is not enough. as multiple closures could exist in the file, even just line number is not enough, because again, multiple closures or arrow functions could live in the same line. So this is as "loose" as we can go for closures :/

PR #2296, opened 2026-09-01
fix(reporting): drop closure positions from loose baseline keys

sgoo:
## 📌 What Does This PR Do?

Normalizes messages in loose baselines where they reference a specific closure on a specific line number, to strip that line number/source position.

## 🔍 Context & Motivation

In our large legacy codebase when we tried to update mago this week (1.25 -> 1.47) it added a lot of new findings for `unhandled-thrown-type` in Closures.
These findings all pin to specific line numbers, this breaks the standard workflows our teams are use to where baselines don't care for line numbers.

## 🛠️ Summary of Changes

- **Bug Fix:** Fixed line number references to closures in loose baselines.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [x] Analyzer
- [ ] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

#2007 

## 📝 Notes for Reviewers

I sort feel like this isn't the most graceful solution to this problem, and would understand if you would prefer to close this PR in favor of a better change that passes this data in a more structured way, thats probably beyond my understanding of the mago codebase.

### DIFF
--- a/crates/reporting/src/baseline.rs
+++ b/crates/reporting/src/baseline.rs
@@ -158,6 +158,83 @@ fn normalize_path(path: &[u8]) -> String {
     String::from_utf8_lossy(path).replace('\\', "/")
 }
 
+/// The synthetic-name prefixes emitted for anonymous constructs, including the
+/// trailing `:` that separates the prefix from the file path.
+///
+/// See `mago_codex::build_synthetic_name`, which renders these as
+/// `{<prefix>:<path>:<line>:<column>}`.
+const SYNTHETIC_NAME_PREFIXES: [&str; 2] = ["{closure:", "{anonymous-class:"];
+
+/// Strips the `:<line>:<column>` suffix from every synthetic anonymous-construct
+/// name in an issue message, turning `{closure:src/foo.php:12:5}` into
+/// `{closure:src/foo.php}`.
+///
+/// Loose baselines key issues on `(file, code, message)`, so a message that
+/// embeds a closure's position changes whenever unrelated edits shift that
+/// closure — making the entry impossible to keep baselined. Strict baselines
+/// keep the position, since several closures can share a line and only the
+/// column tells them apart.
+///
+/// Baselines written before this normalization store the unnormalized message and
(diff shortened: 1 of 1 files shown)


azjezz:
This would be considered a BC break, and would invalidate all existing baselines using the loose format for existing users. Not sure what's the solution for this, but this change won't be accepted due to the BC concern :/

sgoo:
Could this change be accepted if it applied the same normalization rules as it read baselines?
Then it would generate baselines new the new format, but still validate against baselines in the old format.

Would need to be carful how it worked since multiple entries from the old format would then coalesce into a single entry in the new format.

azjezz:
maybe.. but there's another issue with no tracking col/line of closures ( as well as anonymous classes ): if the original issue is fixed, and then same type of problem gets re-introduced in a completely new code, mago won't complain about it, as the baseline will filter the new one, no?

sgoo:
Ok I'm going to close this PR for now, we have worked around this by adding `// @mago-expect` comments to all our closures.

FWIW, from our pov managing a large code base with many contributors and a loose baseline, we are interested in preventing new issues from being added to the codebase, we periodically re-baseline to capture issues that have been fixed, but we are not interested in strongly tracking things.

We have two major areas of our single codebase, one is new, getting active development and new features, this area we are aiming for a a 0 issue baseline. The other is our legacy codebase, it is complicated and full of baselined issues. We have no intention of solving these issues.

I would be fine if my loose mago baseline just knew, "There are 5 `unhandled-thrown-type` in file Foo.php", and it had no idea what line number or class/function contained those issues.
We need the baseline to be as loose as possible, we have had multiple production incidents where someone was modifying legacy code, and ended up trying to resolve mago findings that were unrelated to their change, but did so in a way that caused a production bug.

We would prefer our mago baseline to be as lenient as possible, maybe one day we will be more serious about getting our baseline down, but for now we really want to be as loose as possible, and be happy that all our new code is has no known issues.

Issue #2287, opened 2026-08-28
@psalm-assert-if-true $arg with argument being $var?->something should make $var not null


mathroc:
### Description

now that I learned how to use @psalm-assert a bit better after https://github.com/carthage-software/mago/issues/2276, I think I found a different issue. assert made on the prop of a nullable variable should make Mago understand that the variable is not null when the assert says `$var?->something` is not null

### Playground Link

https://mago.carthage.software/latest/en/playground/#01a04a0e-f4b3-4e12-7ea6-d82dbf69488b

### Environment

- Mago 1.47.4

Issue #2278, opened 2026-08-28
@psalm-assert-if-true also narrows when function returns false


mathroc:
### Description

I'm not sure what's happening, but the playground shows that Mago thinks everything is fine, while it's a type error : https://3v4l.org/ZsZbq#v

### Playground Link

https://mago.carthage.software/latest/en/playground/#01a047dc-282a-3769-5f27-9a7afd1e0a7b

### Environment

- Mago 1.47.4

PR #2321, opened 2026-09-04
fix(analyzer): preserve exact assertions across disjunctions

### LINKED ISSUE #2320: class-string inferance fails to narrow inside class_exists || interface_exists
### Description

Mago reports "possibly received `string`" on values that have already been tightened using `class_exists()` or `interface_exists()`.

### Playground Link

https://mago.carthage.software/latest/en/playground/#01a06e1e-df34-637b-35b8-a093ebafaf44

### Reproduction

_No response_

### Environment

- Mago v1.47.6
- MacOS
- Composer

shuvroroy:
## 📌 What Does This PR Do?

Preserves `class-string` narrowing after `class_exists($class) || interface_exists($class)`, so passing the checked value to `ReflectionClass` no longer reports `possibly-invalid-argument`. This also handles the equivalent early-return guard using `!class_exists($class) && !interface_exists($class)`.

## 🔍 Context & Motivation

Both existence functions supply exact conditional assertions. The analyzer collects these separately to avoid inferring their inverse when a check fails, but the collector discarded them for a true OR condition or a false AND condition.

The regression test includes the issue's [playground example](https://mago.carthage.software/latest/en/playground/#01a06e1e-df34-637b-35b8-a093ebafaf44), including its `namespace\` calls and PHP 8.4 syntax.

## 🛠️ Summary of Changes

- Combine assertions for variables constrained by both alternatives, distributing OR over their assertion clauses.
- Preserve one-way assertion behavior and use the existing formula size threshold to bound expansion.
- Add regression coverage for both operand orders, early returns, unchecked alternatives, exact assertions on both truth branches, nested conditions, and the expansion limit.

## 📂 Affected Areas

- [x] Analyzer

## 🔗 Related Issues or PRs

Fixes #2320.

## 📝 Notes for Reviewers

The original reproduction and the OR/early-return regression fixture failed before the fix and pass afterward. Existing regressions for #2278 and #2287 also pass.

Validation:

- `just test` — passed (8,540 tests), including all 2,532 analyzer integration tests.
- `just check` — passed (PHP formatting, linting and analysis; Rust formatting, Clippy, and workspace check).
- `just build` — passed; the release binary also analyzes the playground regression without issues.

### DIFF
--- a/crates/analyzer/tests/cases/strings_class_or_interface_exists_narrows.php
+++ b/crates/analyzer/tests/cases/strings_class_or_interface_exists_narrows.php
@@ -2,8 +2,8 @@
 
 declare(strict_types=1);
 
-/** @param class-string|interface-string $class */
-function takes_class_or_interface_string(string $class): void
+/** @param class-string $class */
+function takes_class_string(string $class): void
 {
     echo $class;
 }
@@ -11,11 +11,11 @@ function takes_class_or_interface_string(string $class): void
 function narrows_disjunction(string $class): void
 {
     if (class_exists($class) || interface_exists($class)) {
-        takes_class_or_interface_string($class);
+        takes_class_string($class);
     }
 
     if (interface_exists($class) || class_exists($class)) {
-        takes_class_or_interface_string($class);
+        takes_class_string($class);
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/analyzer/tests/cases/strings_class_or_interface_exists_narrows.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+/** @param class-string|interface-string $class */
```
shadowhand: `interface-exists` is an alias for `class-string`, so this type is effectively `x|x`.

shuvroroy: Thanks for pointing that out. I've simplified the annotation to `class-string` and renamed the helper accordingly. The regression test still passes.

shadowhand (approved):
I can't say much about the Rust side, but the PHP looks good.

Issue #1364, opened 2026-03-11
Missing class_alias() support


dotdash:
### Feature

For example the Google SDK and ParagonIE/Halite make use of class_alias() which makes mago not recognize the referenced classes.

### Example

```php
<?php

class A {}

class_alias(A::class, 'B'); 

new B();
```

[Playground](https://mago.carthage.software/playground#019cded6-66ce-030a-abfe-e80fa861010e)

PR #2326, opened 2026-09-05
fix(analyzer): allow undefined class-alias destinations

### LINKED ISSUE #2323: analyse: class_alias() argument 2 is checked as an existing class, but must not exist
### Description

Argument 2 of `class_alias()` is checked as if it must name an existing class. It is the alias being *created*, and PHP requires it **not** to exist.

The diagnostic is therefore inverted: the only way to satisfy it is to make the call fail.

### Example

```php
<?php

declare(strict_types=1);

class Real {}

// Correct: `Alias` must NOT already exist, or the call fails.
class_alias(Real::class, Alias::class);
```

```
src/repro.php:8:26: error[non-existent-class-like]: Class, interface, enum, or trait `Alias` not found.
 = Help: Ensure the name is correct, including its namespace, and that it's properly defined and autoloadable.
```

Following the help text and defining `Alias` is what breaks the code. On PHP 8.3:

```php
class Real {}
class Taken {}

var_dump(class_alias(Real::class, 'Fresh'));      // bool(true)  — arg 2 did not exist
var_dump(class_alias(Real::class, Taken::class)); // bool(false) — arg 2 already existed
// Warning: Cannot declare class Taken, because the name is already in use
```

### Expected

No diagnostic for argument 2 of `class_alias()`.

### Possibly the same root cause

The check looks general rather than specific to `class_alias`: a bare `$a = \Missing\Name::class;` produces the identical message, so the rule appears to be "`Foo::class` requires `Foo` to be declared". As typo detection that is reasonable, and I am not asking for it to change in general.

But it also fires on the existence-check family, whose purpose is to accept a name that may be absent:

```php
if (class_exists(\Optional\Dependency::class)) { }  // error[non-existent-class-like]
interface_exists(\Optional\Contract::class);
enum_exists(\Optional\Kind::class);
trait_exists(\Optional\Helper::class);
```

That one is a judgement call rather than a clear defect: a typo there does leave the guard silently false forever, so you may prefer to keep a diagnostic, or lower its level, rather than exempt the position. Your call, and I have kept it separate from the `class_alias` case for that reason.

The guarded form that `api-platform/core` 2.7 ships hits both at once:

```php
if (!interface_exists(\Legacy\C::class)) {
    class_alias(\Modern\Contract::class, \Legacy\C::class);
}
```

### Note

Distinct from #1364 (missing `class_alias()` support), which asks mago to register the aliased name as a usable symbol. This is only the false positive on the argument, and it stands whether or not #1364 is implemented.

### Playground Link

https://mago.carthage.software/latest/en/playground/#01a06e88-e9cc-16b3-1e9e-256d978b980c

### Reproduction

_No response_

### Environment

mago 1.47.6, aarch64-apple-darwin, `php-version = "8.3.0"`, no extensions.

shuvroroy:
## 📌 What Does This PR Do?

Fixes the false `non-existent-class-like` diagnostic for `class_alias(Real::class, Alias::class)` when `Alias` is the name being created.

## 🔍 Context & Motivation

The general `::class` check required the alias destination to exist, even though `class_alias()` requires an unused destination name. This change exempts a direct `Name::class` expression passed to the built-in function’s `$alias` parameter.

## 🛠️ Summary of Changes

- Use the resolved function and mapped parameter to support qualified/imported calls and reordered named arguments.
- Scope the exemption with a spare bit in the existing context flags, with no new allocation or context storage. Restrict it to a literal class-name expression and restore it after argument analysis.
- Add a regression fixture covering the reported example and preserving diagnostics for missing source classes, ordinary class references/constants, nested calls, and namespaced functions named `class_alias`.

## 📂 Affected Areas

- [x] Analyzer

## 🔗 Related Issues or PRs

Fixes #2323.

## 📝 Notes for Reviewers

The initial regression fixture failed on main with four unexpected `non-existent-class-like` diagnostics. The fix leaves the separate existence-check policy and alias-symbol registration (#1364) unchanged.

Validation:

- Regression confirmed failing before the fix and passing afterward.
- `just test`: 8,537 tests passed, zero failures.
- Exact issue reproduction analyzed with PHP 8.3 settings: no issues.
- `just build`: passed.

### DIFF
--- a/crates/analyzer/src/context/block_flags.rs
+++ b/crates/analyzer/src/context/block_flags.rs
@@ -25,6 +25,8 @@ impl BlockContextFlags {
     /// class-like is in scope for name resolution, but the arguments are evaluated outside
     /// its body.
     pub const INSIDE_CLASS_LIKE_ATTRIBUTE: u32 = 1 << 18;
+    /// Set only while analyzing a literal `Name::class` used as a `class_alias()` destination.
+    pub const INSIDE_CLASS_ALIAS_DESTINATION: u32 = 1 << 19;
 
     #[inline]
     pub const fn new() -> Self {
@@ -149,6 +151,11 @@ impl BlockContextFlags {
     }
 
     #[inline(always)]
+    pub const fn inside_class_alias_destination(&self) -> bool {
+        self.contains(Self::INSIDE_CLASS_ALIAS_DESTINATION)
+    }
+
+    #[inline(always)]
     pub fn set_inside_conditional(&mut self, value: bool) {
         self.set(Self::INSIDE_CONDITIONAL, value);
     }
@@ -241,5 +248,10 @@ impl BlockContextFlags {
     #[inline(always)]
     pub fn set_inside_class_like_attribute(&mut self, value: bool) {
         self.set(Self::INSIDE_CLASS_LIKE_ATTRIBUTE, value);
+    }
+
+    #[inline(always)]
+    pub fn set_inside_class_alias_destination(&mut self, value: bool) {
+        self.set(Self::INSIDE_CLASS_ALIAS_DESTINATION, value);
     }
 }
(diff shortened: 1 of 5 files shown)


azjezz:
While this suppresses the issue, it does not solve the issue, in a sense, that mago still does not understand aliases, and you can't use them.

I already have an implementation for full alias support in progress :)

PR #2349, opened 2026-09-08
fix(analyzer): recognize exhaustive enum case unions

### LINKED ISSUE #2343: False `possibly-invalid-argument`: native enum type not considered contained in Enum::*
### Description

Passing an arg `Suit` to one annotated `@param Suit::*` reports `possibly-invalid-argument`, though the two denote the same set: an enum has no inhabitants beyond its cases. Mago compares the atom `enum(Suit)` against the union `enum(Suit::Hearts)|enum(Suit::Spades)` without expanding it, so containment fails.

PHPStan: https://phpstan.org/r/ead431a0-bf13-428a-ac29-65d8d99b9575

### Playground Link

https://mago.carthage.software/main/en/playground/#01a07f97-070c-0b43-48f7-f433b2ac48db

### Reproduction

_No response_

### Environment

Mago main as of today

shuvroroy:
## 📌 What Does This PR Do?

Accepts a native enum where a parameter expects all of that enum’s cases. For example, passing `Suit` to `@param Suit::*` now succeeds when the wildcard expands to `Suit::Hearts|Suit::Spades`.

## 🔍 Context & Motivation

The comparator checked the native enum against each individual case and reported `possibly-invalid-argument`, even though the union covered every possible enum value.

## 🛠️ Summary of Changes

- After ordinary atomic matching fails, check whether the container covers every declared enum case.
- Read existing metadata without constructing an expanded union. Reject containers with too few members early, and check case identity so duplicates or unrelated enums cannot stand in for missing cases.
- Add analyzer regression coverage for wildcard/explicit unions, nullable types, returns, generic bounds, backed enums, singleton enums, and incomplete unions.
- Add comparator tests for duplicate cases, name casing, missing/empty metadata, and coercion flags.

## 📂 Affected Areas

- [x] Analyzer

## 🔗 Related Issues or PRs

Fixes #2343.

## 📝 Notes for Reviewers

The original playground regression failed before the fix with `possibly-invalid-argument` and passes afterward. Existing successful atomic comparisons retain their current behavior.

Validation:

- Regression confirmed failing before the fix and passing afterward.
- `just test`: 8,569 tests passed, zero failures.
- `just build`: release build passed.
- Release binary analyzed the original reproduction with PHP 8.4 settings: no issues.


azjezz:
@shuvroroy can you verify the perf regression here?

shuvroroy:
> @shuvroroy can you verify the perf regression here?

@azjezz Thanks for flagging this. I inspected the CodSpeed profiles for both regressed benchmarks: instruction cost is unchanged, and the increases come entirely from simulated cache/memory penalties. Both benchmarks return through the existing equality fast path, before reaching the new enum handling.

The base ran on an EPYC 9V74 and the PR on an EPYC 7763, which explains the environment warning. That doesn’t rule out a code-layout regression from this change, though.

Could we compare base and head on the same Linux machine with identical build settings? That would help establish whether a code change is needed before attributing this to the runner difference.

azjezz:
>Could we compare base and head on the same Linux machine with identical build settings

That's what i was asking :)

Issue #2205, opened 2026-08-17
Wrong preg_match return type when pattern is valid


BafS:
### Description

`preg_match*` should not return false when the pattern is valid, Mago always assume that it can return `false`.

Example with PHPStan: https://phpstan.org/r/c98bdb41-434f-4ffb-9fa5-d05e4366a65c

### Playground Link

https://mago.carthage.software/main/en/playground/#01a00da2-f612-b618-e8e1-952089abd8cb

### Environment

Mago main as of today

shuvroroy:
I reproduced the playground warning. One detail changes the implementation: a valid pattern can still return `false` on runtime errors, and PHPStan retains `0|1|false` / `non-negative-int|false` in its preg_match tests. Its linked playground reports the malformed pattern separately rather than removing `false`.

I am preparing a fix that preserves those return types, accepts integer-or-false `> 0` checks (and the reversed form), narrows the successful branch correctly, and reports the malformed pattern’s delimiter error as an invalid argument. Comparisons such as `>= 0`, arithmetic, and integer-only arguments will retain their existing failure checks. The delimiter check will not claim to validate full PCRE syntax.

Reference: https://github.com/phpstan/phpstan-src/blob/2.2.x/tests/PHPStan/Analyser/nsrt/preg_match_php8.php

azjezz:
as mentioned in #2350 , a valid pattern does not rule out `false`.

PR #2350, opened 2026-09-08
fix(analyzer): validate preg_match pattern delimiters

shuvroroy:
## 📌 What Does This PR Do?

Reports invalid delimiters in literal `preg_match()` and `preg_match_all()` patterns. For example, `preg_match('^[a-fA-F0-9]*$/', $subject)` now reports `invalid-argument` because its opening `^` delimiter is never closed.

## 🔍 Context & Motivation

The malformed pattern in #2205 was not diagnosed. The provider now identifies definite delimiter errors, while the prelude describes successful return values more precisely and retains `false` for failures.

An unhandled `preg_match(...) > 0` comparison continues to report `possibly-false-operand`. Callers should handle `false` explicitly before comparing the integer result.

## 🛠️ Summary of Changes

- Validate literal pattern delimiters for resolved built-in calls, including imported names and named arguments. The scanner uses linear time and constant space without a new dependency.
- Specify `0|1|false` for `preg_match()` and `non-negative-int|false` for `preg_match_all()`.
- Cover malformed delimiters, dynamic patterns, imports, namespaced functions, runtime failure branches, and precise integer ranges after explicit failure handling.
- Preserve existing comparison diagnostics and reconciliation behavior, including the original `strpos() > 0` regression test.

## 📂 Affected Areas

- [x] Analyzer
- [x] Other: prelude

## 🔗 Related Issues or PRs

Related to #2205. This addresses the malformed-pattern diagnostic; it does not remove `false` from valid-pattern return types or suppress the comparison warning from the first example.

## 📝 Notes for Reviewers

Delimiter validation does not validate full PCRE syntax or modifiers. Matching delimiters do not rule out runtime failures, such as invalid UTF-8 or invalid offsets.

The restored `strpos() > 0` fixture fails against the previous PR implementation because the expected warning is suppressed, and passes after restoring the existing diagnostic behavior.

Validation:

- Restored warning regression confirmed failing before the review fix and passing afterward.
- Focused PCRE and #2205 regression tests passed.
- `just test`: 8,566 tests passed, zero failures.


crates/analyzer/src/expression/binary/comparison.rs
```diff
@@ -69,8 +69,15 @@ where
...
-    check_comparison_operand(context, binary.rhs, rhs_type, "Right", &binary.operator);
+    // Comparing an integer-or-false result with `> 0` explicitly excludes the failure value.
+    let allows_false = match binary.operator {
+        BinaryOperator::GreaterThan(_) => is_positive_integer_check(lhs_type, rhs_type),
+        BinaryOperator::LessThan(_) => is_positive_integer_check(rhs_type, lhs_type),
+        _ => false,
+    };
+
+    check_comparison_operand(context, binary.lhs, lhs_type, "Left", &binary.operator, allows_false);
+    check_comparison_operand(context, binary.rhs, rhs_type, "Right", &binary.operator, allows_false);
```
azjezz: This is wrong, it is suppressing a valid issue,and is unrelated to the preg_ function provider. if an operand could be false, we should always report possibly-false operand, because it is most likely something you want to handle in a different way.

crates/analyzer/tests/cases/strings_strpos_arithmetic_with_false_warns.php
```diff
@@ -2,8 +2,8 @@
 
 declare(strict_types=1);
 
-function probe(string $h): bool
+function probe(string $h): int
 {
     /** @mago-expect analysis:possibly-false-operand */
-    return strpos($h, 'x') > 0;
+    return strpos($h, 'x') + 1;
 }
```
azjezz, line 9: should be reverted. see above.

azjezz (changes requested):

PR #2369, opened 2026-09-11
fix(orchestrator): use 4-byte emoji for guard progress bar

bendavies:
## 📌 What Does This PR Do?

Replaces the shield emoji (🛡️) with a padlock (🔒) in the guard progress bar.

before: 
image

after: 

image

## 🔍 Context & Motivation

Same class of problem as #866, which fixed the analysis progress bar. The guard prefix was the last one still using a variation selector sequence.

## 🛠️ Summary of Changes

- **Bug Fix:** Changed the guard progress prefix from `🛡️  Guarding` (7 bytes, 2 chars) to `🔒 Guarding` (4 bytes, 1 char) in `GuardService::run`.

## 📂 Affected Areas

- [ ] Linter
- [ ] Formatter
- [ ] Analyzer
- [x] CLI
- [ ] Dependencies
- [ ] Documentation
- [ ] Other (please specify):

## 🔗 Related Issues or PRs

Follows #866.

## 📝 Notes for Reviewers

- Verified in the PHPStorm terminal with a release build, which is where the shield failed to render.
- Bar alignment is unchanged. `unicode-width` 0.2.2 already measured the shield plus selector as 2 columns, so the `{prefix:<16}` padding in `progress.rs` was never wrong. The double space went with it because the padlock needs no extra spacing.
- `🔬 Analyzing` and `🧹 Linting` were already single 4-byte characters, so no other prefix needed a change.
- Other 4-byte options if the padlock is not preferred: 🛟 🚧 🚨 🧱.

### DIFF
--- a/crates/orchestrator/src/service/guard.rs
+++ b/crates/orchestrator/src/service/guard.rs
@@ -80,7 +80,7 @@ impl GuardService {
     ///
     /// Returns [`OrchestratorError`] when guard pipelines fail or scanning encounters IO errors.
     pub fn run(self) -> Result<GuardResult, OrchestratorError> {
-        const GUARD_PROGRESS_PREFIX: &str = "🛡️  Guarding";
+        const GUARD_PROGRESS_PREFIX: &str = "🔒 Guarding";
 
         // Determine upfront which guards will be skipped due to missing config
         let skipped_perimeter = matches!(self.settings.should_run_perimeter(), Some(false));


crates/orchestrator/src/service/guard.rs
```diff
@@ -80,7 +80,7 @@ impl GuardService {
     ///
     /// Returns [`OrchestratorError`] when guard pipelines fail or scanning encounters IO errors.
     pub fn run(self) -> Result<GuardResult, OrchestratorError> {
-        const GUARD_PROGRESS_PREFIX: &str = "🛡️  Guarding";
+        const GUARD_PROGRESS_PREFIX: &str = "🔒 Guarding";
```
azjezz, line 83: Mmm.. maybe `🚧` is better? 😅

bendavies: sure

azjezz: nvm actually

